PR #2, opened 2015-01-01
Update prepare.sh

543:
The python-bcrypt package is needed to make the application run it seems.


ColdHeat:
Handled in more depth in a later pull request. Thanks for the help though :)

PR #4, opened 2015-01-02
Fixed permanent "referenced before assignment" condition

takeshixx:
Whenever a user submits account data via POST requests to `/profile`, the current password has to be supplied in the "Old Password" titled field. However, when the user omits a "new" password, the password variable will be referenced regardless if it is assigned or not.

In order to make it clear for users where they have to supply a password, additional changes to the `/profile` route are necessary.


ColdHeat:
`/profile` definitely needs some more work. Thanks for the help :)

PR #6, opened 2015-01-02
Missing dependencies added, some weird imports fixed

takeshixx:
I started to cleanup the code because I'm planning to use it for an upcoming workshop. Great work btw, thanks for putting this on GitHub! :)

During the cleanup I stumbled across some missing imports, syntax errors and uncaught exceptions. This commit fixes some of them. Especially the imports look pretty messed up. Most notably, Flask functions should not be imported via the `CTFd` module but directly from `flask`.

Resolving these and some other import issues might allow you to not put every module's code in a pseudo-function where other modules won't be able to import global instances (like `mail`). :)

### DIFF
--- a/CTFd/__init__.py
+++ b/CTFd/__init__.py
@@ -28,6 +28,7 @@ def create_app(subdomain, username="", password=""):
         app.db = db
         # app.setup = True
 
+        global mail
         mail = Mail(app)
 
         Session(app)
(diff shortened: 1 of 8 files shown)


ColdHeat:
So I originally began importing from `CTFd` when I began working on https://ctfd.io/ which required me to support application factories and make some other questionable changes. 

Specifically importing `session` from Flask will cause sessions to be client-side as opposed to the file system with Flask-Session. I hadn't thought of making it global. I'll test our your changes and eventually merge it in.

Thanks a ton for the pull requests :+1:

PR #11, opened 2015-01-08
Optionally allow unregistered users to view challenges

bburky:
Add a Config entry `view_challenges_unregistered` to indicate whether
unregistered users can view challenges. Add the setting to the admin config
page.

Add can_view_challenges() to utils to test if a user is either authed, or the
configuration allow unauthenticated users to view the challenges.

Return a HTTP 401 Unauthorized error when the /chals/solves API can't provide
results for an unauthenticated user. This is needed because the client side
code in `chalboard.js` doesn't know if it's logged in or not and requests this
anyway. (And AJAX doesn't handle redirects very well.) Alternately the client
could actually know if they're logged in and not make needless API calls.

When an unregistered user attempts to submit a flag, it will also fail. The
user will be redirected to a login page.

### DIFF
--- a/CTFd/admin.py
+++ b/CTFd/admin.py
@@ -52,7 +52,12 @@ def init_admin(app):
                 start = None
                 end = None
 
-            print repr(start), repr(end)
+            try:
+                view_challenges_unregistered = bool(request.form.get('view_challenges_unregistered', None))
+            except (ValueError, TypeError):
+                view_challenges_unregistered = None
+
+            print repr(start), repr(end), repr(view_challenges_unregistered)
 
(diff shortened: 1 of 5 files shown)


bburky:
Storing all the config data in the db as strings is kind of ugly. And it converts types at random times. It could be cleaned up some.

Also, you will need to recreate the database. The new Config entry needs be created during setup.

ColdHeat:
I originally stored data as strings in the db so that I wouldn't have to recreate databases as often. It would just be a try-except for existence and creation. But it is a little janky. The right solution is probably to implement database migration. 

I like the idea, and I can always change stuff later :+1:

PR #17, opened 2015-01-26
make the ban and unban switcher work

xkjcf:
there is a ban/uban switcher in admin/exoreboard.html, but when you query the database, you lost the parameter.


ColdHeat:
Thanks!

Issue #15, opened 2015-01-25
Delete Challenges


ColdHeat:
Kind of silly that this isn't already implemented. 

Delete challenges and delete all solves associated with that challenge.

mwinstead3790:
If you'd like, I have implemented this feature in my fork of your repo. I'll be sending over a pull request to you as well. Let me know if you'd like anything changed. 

Also, for ease of acceptance, I have left in debugging and comments which I can strip if need be.

PR #19, opened 2015-03-14
Delete challenges

mwinstead3790:
I have implemented Enhancement issue #15 . If acceptable, I can remove the debugging lines (which are there so you can see what is going on) or change anything as needed.

### DIFF
--- a/CTFd/admin.py
+++ b/CTFd/admin.py
@@ -512,7 +512,7 @@ def init_admin(app):
                 except OSError, e:
                     pass
                     
-                print "removing file %s: %s %s" % (repr(curr_file), curr_file.location, tree_to_delete)
+                #print "removing file %s: %s %s" % (repr(curr_file), curr_file.location, tree_to_delete)
 
                 db.session.delete(curr_file)
 
--- a/CTFd/admin.py
+++ b/CTFd/admin.py
@@ -7,6 +7,9 @@ from socket import inet_aton, inet_ntoa
 from passlib.hash import bcrypt_sha256
 from flask import current_app as app
 
+#For deleting files
+import shutil
+
 import logging
 import hashlib
 import time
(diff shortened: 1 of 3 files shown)


CTFd/admin.py
```diff
@@ -477,6 +480,47 @@ def admin_create_chal():
         db.session.close()
         return redirect('/admin/chals')
 
+    @app.route('/admin/chal/delete', methods=['POST'])
+    def admin_delete_chal():
+        challenge=Challenges.query.filter_by(id=request.form['id']).first()
```
ColdHeat: Needs spacing around the assignment equals sign.

CTFd/admin.py
```diff
@@ -477,6 +480,47 @@ def admin_create_chal():
...
+        if challenge is not None:
```
ColdHeat: Just needs to be `if challenge:`

CTFd/admin.py
```diff
@@ -477,6 +480,47 @@ def admin_create_chal():
...
+            #print challenge, " id: ", challenge.id
+            #Clean up wrong keys
+            #wrong_keys=WrongKeys.query.filter(chal=challenge.id).all()
+            wrong_keys= db.session.query(WrongKeys, Challenges).filter(Challenges.id==WrongKeys.id, Challenges.id==challenge.id).all()
```
ColdHeat: Pretty sure this should be `challenge.id==WrongKeys.chal`

ColdHeat:
Thanks so much for the PR! I left you some comments in the files changed. But I have a feeling more of those queries are incorrect and I think with a portion of CTFd this sensitive I'd rather implement it on my own. 

I'll be sure to credit you in my commit which should be in a few hours.

mwinstead3790:
Hey, sorry for the first pep8 violation, but the if statement is pep8 programming recommendation 2 [1].

Also, the database operation is selecting from the Challenge.id foreign keys. I tried for a couple of hours just fetching with something like the following:
```python
wrong_keys = db.session.query(WrongKeys).filter(WrongKeys.id==challenge.id).all()
```

but I couldn't get rows returned. Thus, I read more about foreign keys on the object relational tutorial for SQLAlchemy 0.9 and when I tried the joins [2], all was good. Thus, this is just making a join table between the two tables and then setting a condition on the Challenges.id field:

```python
wrong_keys= db.session.query(WrongKeys, Challenges).filter(Challenges.id==WrongKeys.id, Challenges.id==challenge.id).all()
 ```
[1] http://docs.sqlalchemy.org/en/rel_0_9/orm/tutorial.html#querying-with-joins
[2]https://www.python.org/dev/peps/pep-0008/#programming-recommendations

Regardless - thanks for responding to the pull request!
-Mike

mwinstead3790:
Gotcha - thanks for getting to looking at the pull request.

I'll probably also implement user promotion and page deletion as well. I'll go back and test/fix these in my fork if you ever want to check it back out - I see where I missed a couple of things. 

Thanks again,
-Mike

PR #30, opened 2015-03-22
Update installation scripts

bburky:
Install libffi-dev in prepare.sh. Bcrypt seems to depend on this.

Add exact versions for all packages in requrements.txt

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-sudo apt-get install build-essential python-dev python-pip
+sudo apt-get install build-essential python-dev python-pip libffi-dev
 pip install -r requirements.txt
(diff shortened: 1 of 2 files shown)


bburky:
Crap, I just realized I forgot to put a version in bcrypt in `requirements.txt`. You might want to do a `pip freeze` and add the version yourself.

ColdHeat:
I'm not particularly concerned. I'll add it if it becomes a problem.

bburky:
Versions aren't a big deal, but I've had them come in handy when I pick up a project after not touching it for a year or something. So you don't have to guess what versions of modules your project actually works with.

...On the other hand if you have versions, you periodically have to update them all to new versions and test that they work with the project.

PR #31, opened 2015-03-22
Fix Mailgun from address

bburky:
Use the `app.config['ADMINS'][0]` address as the from address in Mailgun messages.

TODO: This email address should probably be configurable in the admin settings.

This is a quick fix to not hardcode the address for emails sent via Mailgun. I also must confess I haven't tested this or configured Mailgun yet.

You might want to just set the admin address to <noreply@ctfd.io> on ctfd.io to get the old behavior back.


ColdHeat:
Haven't configured mailgun personally either. Having trouble verifying my domain.

bburky:
I've actually verified my domain, just not hooked it up to CTFd. So I'll try to test this later.

PR #35, opened 2015-05-09
Updated UI to be a bit nicer than before

breadchris:
Here are some screenshots of the new UI:

Homescreen: homescreen
Challenge board: chalboard
Challenge: challenge
Login: login
Login text entered: text
Profile: profile


slinkymanbyday:
I prefer the current interface.

MarvelousBreadchris:
what do you prefer about the current interface?

slinkymanbyday:
I prefer the black header. Also I'm not a big fan of the h1 heading being included in the coloured section, makes the navbar seem a bit fat and doesn't seperate ou the nav section from the title.

I do however like that you've changed the top right hand menu, i don't like it as a dropdown

MarvelousBreadchris:
Since I used sass the site colors can be controlled with a single variable so getting a black header would be pretty simple and the header can be smaller

ColdHeat:
Let's remove some of these files like `nohup.out`, change the color to something a little more neutral (black, dark gray, etc.), and apply the theme to admin panel before we merge. 

But I like the changes overall.

breadchris:
Admin panel is done

screen shot 2016-01-30 at 4 01 41 pm
screen shot 2016-01-30 at 4 02 01 pm
screen shot 2016-01-30 at 4 01 34 pm
screen shot 2016-01-30 at 4 01 27 pm
screen shot 2016-01-30 at 4 01 18 pm
screen shot 2016-01-30 at 4 01 06 pm

breadchris:
Updated to reflect desired changes

breadchris:
This is the money commit

ColdHeat:
Closed due to request.

PR #39, opened 2015-05-18
enable max-attempts per challenge setting

slinkymanbyday:
This allows admin to set a maximum number of attempts per challenge (0 to disable, which is default).

If max attempts is reached, challenge is marked red and displays appropriate error message when more solutions are submitted.

### DIFF
--- a/CTFd/admin.py
+++ b/CTFd/admin.py
@@ -71,6 +71,7 @@ def init_admin(app):
             ctf_name = set_config("ctf_name", request.form.get('ctf_name', None))
             mg_api_key = set_config("mg_api_key", request.form.get('mg_api_key', None))
             do_api_key = set_config("do_api_key", request.form.get('do_api_key', None))
+            max_tries = set_config("max_tries", request.form.get('max_tries', None))
 
             db_start = Config.query.filter_by(key='start').first()
             db_start.value = start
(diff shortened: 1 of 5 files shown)


ColdHeat:
Hmm... 

CSAW, what CTFd was written for, does not restrict key submissions and no jeopardy based CTF I've ever seen does either. 

I have recently used CTFd in an graded class exercise which potentially could have benefited from a max attempts setting. But I'm not sure how beneficial it is to start converting CTFd into a homework grading site like WebAssign. 

I think a feature like this would be best kept in a plugin system which is what I hope to begin writing soon. 

What did you need a max attempts per challenge for?

slinkymanbyday:
The CTF that I'm helping develop needs a max attempts per challenge. 

It's a closed CTF (invite only).

A CTF that i participated in (CySCA, also not public) also had max attempts per challenge.

It's fine if you don't want it in pulled here, just thought others might want it too.

ColdHeat:
It's definitely useful which is why I'm sort of on the fence. Given the variety of CTF structures, a plugin system would be pretty good to have overall. To me something like this belongs more in a plugin.

ColdHeat:
I'll move this into a plugin once a plugin system is setup but for now this is fine :)

Issue #29, opened 2015-03-19
Config flag for allowing challenges after CTF end


ColdHeat:
Mildly annoying to keep being redirected to the homepage.

PR #41, opened 2015-05-18
allow to view challenges after ctf end

slinkymanbyday:
fixes issue #29 

hopefully this is a satisfactory fix. It works, but not sure if this is the most elegant way....


ColdHeat:
Seems a little complicated. Why not in `scoreboard.py` something like: 
`if not is_admin() and (not ctftime() or view_after_ctf()):` with `view_after_ctf()` checking the config and if time is past CTF end?

Issue #34, opened 2015-05-01
Reduce restrictions in auth.py


ColdHeat:
No need to:
- [x] Login after register
- [x] Require only email for password reset
- [ ]  ~~Not notify user of password reset failure~~
- [x] Login should take team name or email address

https://github.com/CTFd/CTFd/tree/34_reduce_auth_restrictions

PR #44, opened 2015-07-02
 Reduce restrictions in auth.py 1/3

nocternology:
Hi ! I tried to treat issue #34. I figured a fix for the user to be automatically logged after his registration. I'm trying to figure out the rest of the issue. 

Also, wouldn't it be better to move the session setting to a separate function to avoid code replication ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,6 @@ target/
 *.log
 .idea/
 static/uploads
+
+# Virtual env
+ctfd_venv/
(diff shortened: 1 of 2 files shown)


ColdHeat:
Hey I like your commits, but can you remove your change to .gitignore? ctfd_venv is a little specific to you. 

Sorry for the slow response.

nocternology:
No problems :-)

ColdHeat:
Thanks for submitting this PR and I'm sorry I really let it fall to the side. I'm taking over development of this in #474.

PR #48, opened 2015-09-14
Block registration while authed, remove user input from prepare.sh

gusnaughton:
Prevent users from registering teams whilst registered. Additionally the prepare.sh script is fully automated now.


ColdHeat:
For starters, this doesn't work. The if statement for checking the request method is done before redirection so direct POST requests will go through. 

Secondly, I'm not sure if this is really necessary. Since the register page link isn't shown on the navbar after you register, I don't really see a lot of users wanting to register again when they've already registered.

ColdHeat:
But I do like your change to prepare.sh :+1: which I'll accept if you want to put that in another PR

PR #56, opened 2015-10-02
fix setup bug w/ variable name typo

remasis:


ColdHeat:
I don't believe this endpoint is actually meant to be hit as /login covers authentication to /admin now.

remasis:
Hmmm....I went through the setup process and was taken to the endpoint itself starting from scratch: i.e. clone the repo into a fresh server. I tested in multiple machines/archs but the issue could be with myself using a weird workflow.

ColdHeat:
Upon first authentication you are supposed to hit `/setup` which is defined here: https://github.com/isislab/CTFd/blob/master/CTFd/views.py#L45

Maybe you can show me a stack trace of how you actually reached this endpoint? It's definitely a bug but it's dead code with no real impact. I'd rather just kill off that part of the function.

remasis:
Edited previous comment to include traceback, sorry.

ColdHeat:
You're being redirected to /setup because the application isn't well setup :smile:!

It's definitely a bug and even though it's in dead code, I'll accept the PR anyway since I probably wouldn't have seen it without you.

remasis:
Ha, ok sweet. Yeah this little fix got me up and running so whatevs.

Thanks.  :)

PR #58, opened 2015-10-04
Fixing Dockerfile using old apt cache and start command

JTWorks:
Fixing Dockerfile using old apt cache and start command. 

I had an issue with the apt trying to pull different versions and then the container would not start due to the "()" in the start command.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 FROM ubuntu:trusty
-RUN apt-get update && apt-get upgrade -y
+RUN apt-get clean && apt-get update && apt-get upgrade -y
 
 RUN apt-get install git gunicorn -y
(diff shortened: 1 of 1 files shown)


ColdHeat:
The `()` is definitely necessary in this scenario. As for clean, I'm not sure it really matters since I imagine Dockerfiles should pull a clean base image?

JTWorks:
Yea I realized that afterwards. I am currently rewriting the whole thing to be more production ready and will create a pull request for that then. My plan is to use uwsgi as the web server and having it run as an unprivileged user within the container. The container will share a volume with the host OS to keep persistent storage. 

I may also create a script to setup a systemd-nspawn container down the road, not sure yet though.

JTWorks:
Also I was having an issue with just having update and upgrade. It was looking for an older version of something... can't remember exactly what but adding the clean before update and upgrade fixed it.

PR #62, opened 2015-10-11
Add netloc check to validate_url

dmuhs:
I have added another check to the url validation util so that urls like `http://` or `https://foo` are not considered valid anymore. My editor also removed some trailing whitespaces. :)

### DIFF
--- a/CTFd/utils.py
+++ b/CTFd/utils.py
@@ -1,7 +1,7 @@
 from CTFd.models import db, WrongKeys, Pages, Config
 from CTFd import mail
 
-from six.moves.urllib.parse import urlparse, urljoin 
+from six.moves.urllib.parse import urlparse, urljoin
 from functools import wraps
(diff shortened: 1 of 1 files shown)


ColdHeat:
Not sure I agree with this check. http:// isn't a valid url but http://foo is valid.

dmuhs:
I don't know how realistic it is to include bare TLDs in url validation as most of them don't resolve anyways (_trivia_: http://to being one of the few). The dot check was meant to make sure we're at least on a domain level without having to use a regex.

ColdHeat:
They're often used to access internal only infrastructure. I don't see much harm in having them for those that will use it. And if they don't resolve, then when used on the public internet, some people's websites just don't load.

dmuhs:
You're right. Thanks for pointing that out! I guess it's not that big of a deal then.

ColdHeat:
:smile:

PR #64, opened 2015-10-16
fix Gunicorn Multi-Processing Problem

tjjh89017:
When using Dockerfile to deploy CTFd by gunicorn, I found there is a problem those gunicorn workers can't shared Flask session. That causes users will logout when requesting another page.

To fix this problem, I change the way to get `SECRET_KEY` by `ENVIRONMENT`. If not setting `ENV`, it will use origin way to generate `SECRET_KEY`

### DIFF
--- a/CTFd/config.py
+++ b/CTFd/config.py
@@ -1,6 +1,6 @@
 import os
 ##### SERVER SETTINGS #####
-SECRET_KEY = os.urandom(64)
+SECRET_KEY = os.environ.get('SECRET_KEY') or os.urandom(64)
 SQLALCHEMY_DATABASE_URI = 'sqlite:///ctfd.db'
 SESSION_TYPE = "filesystem"
 SESSION_FILE_DIR = "/tmp/flask_session"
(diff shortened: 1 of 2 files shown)


tjjh89017:
I think I have to find a better way to implement this feature.

ColdHeat:
Can you explain what issue you're having more specifically? 

You're not supposed to run multiple CTFd instances from a single Docker instance. Also perhaps you'ave there is some other issue related to having a random SECRET_KEY? Have you considered setting it a randomly generated but static value? 

Also I'm toying with removing Flask-Session although I don't trust the built-in Flask sessions.

tjjh89017:
I used isislab's gunicorn Dockerfile to build my service. But I found `-w 4` gunicorn will create 4 process and each process will call `os.urandom(64)` in `config.py` (total call 4 times). If I request another page after login, it will use another process to response my request. But Flask session didn't share data between processes, so I would be asked to login again.
When I replace `os.urandom(64)` into `some_secreet_key`, it will work fine.

ColdHeat:
I have not seen this in practice before but it would be resolved with using a static key. I'll take a look into it but setting a key this way is not how I would approach the issue.

tjjh89017:
Ok, Got it.

ColdHeat:
It looks like this is an issue for others as well. I'll create an issue and look into it later today.

tjjh89017:
Thanks a lot

PR #78, opened 2016-02-01
Updated CTFd UI to look nicer

breadchris:
screen shot 2016-02-01 at 1 02 52 am


ColdHeat:
* Some of your resources are hardcoded for HTTP or HTTPS. Can you set them to load relative to the way the user requested the page (i.e. //website).
* The index.html file in /static/uploads is intentionally left there for those who enable Indexes on their web server. It should be left in the repository. 
*  The hash code function seems to have been removed from chalboard.js. Hashcode is used to normalize user inserted data so that it can be used in selectors and such. 
* Link `boostrap-modal.css` to a cdn hosted file please. 
* Same with bootstrap which is in `style.css`.
*  The colorhash function seems to have been removed from the main graph on the scoreboard? 
* While I have no problems with SASS and bootstrap-sass and compass, it's going to be a tough sell to get them into a primarily Python codebase right now. We really don't need those to deploy. Can you do without those in the `prepare.sh` file? And you really only need to provide the SASS files in my opinon. Can you remove the extra fonts and such?

breadchris:
- No more unneeded files
- No more SASS
- Colorhash and hashcode are back
- All http and https resources are prefixed with //

Issue #79, opened 2016-02-02
Flags/keys with uppercase letters can't be submitted


frohoff:
Code that handles submission appears to only call `.strip().lower()` on the flag/key from the submitting user and __NOT__ on the flag/key from the database when doing the comparison. This causes any flag/key submissions with uppercase letters to spuriously fail.

https://github.com/isislab/CTFd/blob/master/CTFd/challenges.py#L141:
```python
...
            key = str(request.form['key'].strip().lower())
            keys = json.loads(chal.flags)
            for x in keys:
                if x['type'] == 0: #static key
                    print(x['flag'], key.strip().lower())
                    if x['flag'] == key.strip().lower():
                        solve = Solves(chalid=chalid, teamid=session['id'], ip=request.remote_addr, flag=key)
...
```

PR #80, opened 2016-02-02
fix for lowercase comparison bug

frohoff:
Fix for #79


ColdHeat:
Interesting I think this might have gotten lost when I went from storing individual keys to storing them as a JSON blob. 

Good catch, thanks!

PR #90, opened 2016-02-16
Add nonce element in admin/login.html

Bird101:
Solve the admin login page csrf problem caused by the lack of nonce


ColdHeat:
I didn't realize this route was still used but thanks!

PR #105, opened 2016-03-22
Add comments and custom config

kevinhinterlong:
The previous dockerfile did not successfully build. I believe there were errors in copying the current directory (CTFd) into the docker and gunicorn would exit immediately after running the docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,29 @@
+# Pull base image
 FROM alpine:3.2
+
+# ------------------------------------------------------------------------------
+# Install Base
 RUN apk update && apk upgrade
 RUN apk add git gcc musl-dev libffi-dev python python-dev py-pip
 
+# ------------------------------------------------------------------------------
+# Set up CTFd
 RUN mkdir /opt
-COPY . /opt/CTFd
+WORKDIR /opt
+RUN git clone https://github.com/isislab/CTFd.git
 WORKDIR /opt/CTFd
-
 RUN pip install -r requirements.txt
 RUN pip install pymysql
 
-CMD ["gunicorn", "--bind", "0.0.0.0:8000", "-w", "4", "CTFd:create_app()"]
+# ------------------------------------------------------------------------------
+# Configure hostname
+
+#RUN sed -i -e "s/ctfd.io/example.com/" /opt/CTFd/CTFd/config.py
+
+# ------------------------------------------------------------------------------
+# Expose ports.
 EXPOSE 8000
+
+# ------------------------------------------------------------------------------
+# Start gunicorn (defines startup command)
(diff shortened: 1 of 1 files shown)


ColdHeat:
So the difference here is how you acquire the Dockerfile. 

In the current version, if you clone the repo and then build the Dockerfile, you can run properly. 

In your version (which I believe is actually how it was done a few commits ago), the Dockerfile clones the repo on each build. 

I am not sure which is better so if you can justify it and I agree I will accept the PR. But keep in mind that ctfd.io does in some situations rely on customizations which deviate from repo. 

Thoughts?

kevinhinterlong:
I did notice that in the previous commits. Could you verify that if you clone the repo and build a docker from it that it does work?
I was able to build it using `sudo docker build https://github.com/isislab/CTFd.git` but it didn't seem to copy the current directory into the docker if I ran `sudo docker build - < Dockerfile` after `git clone https://github.com/isislab/CTFd.git`

Regarding forked repositories, using git to clone only the master one (this) is definitely an issue, but I can't figure out why it won't copy the files in the situation I described.
The most important thing from this commit was the `CMD` line with the appended `&` to keep gunicorn running.

ColdHeat:
I can check on it later but I've usually just built with 
`git clone https://github.com/isislab/CTFd.git`
`docker build`

Are you sure that your issue is that the files aren't copying? That seems super unlikely to me unless for some reason you didn't have read permissions for the files. 

I'm not sure what you mean by the forked repositories comment... There should definitely be more branches, but I don't think it's that big an issue. 

`&` to the best of my knowledge just puts gunicorn in the background. If it still errors out I don't think gunicorn will continue to run... Unless that's a quirk that I'm unfamiliar with.

kevinhinterlong:
When I cloned and then ran, it threw an error on the line where it installed everything from requirements.txt because it couldn't find the file and the only thing I could think of was that it wasn't actually copying the whole directory.

For the forked repositories, I mean that if anyone forks from this they will also have to fix the Dockerfile to clone their repository instead of yours which isn't that much to change, but it's still not very elegant.

When I ran `docker build https://github.com/isislab/CTFd.git` (which did work) I would always see that the container had exited immediately... If you don't find this to be the case with yours maybe I've got some weird problem I'm not seeing

kevinhinterlong:
So I guess I didn't read up on documentation enough https://docs.docker.com/engine/reference/commandline/build/#build-with building from `STDIN` was my problem. Sorry about that.

I'm still not sure what the issue is with gunicorn though.

ColdHeat:
Are you still unable to deploy the Dockerfile?

kevinhinterlong:
Everything seems to be working now, the issue was with how I was building. Thanks for the help though. And for the very nice framework

ColdHeat:
Thanks! 

Let me know if you need any help or have any ideas.

PR #106, opened 2016-04-01
added static height to chal buttons so they render evenly and added a…

guitarmanj:
Added static height so the challenge buttons render evenly and added a radius to the chal buttons so that they're easier on the eyes and don't show that whole 'black dot' effect as much.


ColdHeat:
This PR isn't needed as much because other designs are being worked on and the current new design gets rid of some of the issues you fixed as well while also creating more space to work with.

PR #111, opened 2016-04-22
Fixed button outlining so that custom buttons do not appear invisible

breadchris:
Buttons would be invisible if they did not have .btn-theme. Now they appear normally :3


ColdHeat:
I merged this without testing it b/c it's minor but it does not work...

PR #114, opened 2016-04-28
Optionally hide admin accounts from the teams page

momopranto:
Controlled via HIDE_ADMIN_TEAMS in config.py


ColdHeat:
Primarily I'm not sure I agree with the idea of hiding admins. But I will allow people to have the choice if they so choose versus me being opinionated on it. 

But you will have to clean up the quality of this PR before it'll be accepted. 

First you want to avoid config.py because that's not very easily edittable from the browser. Secondly theres no need to iterate over the teams list again because you can simply add more filters to the query that gathers teams. 

In addition you should all the necessary logic changes to admin.py and also the templates that are associated.

PR #125, opened 2016-06-17
Make CTFd more Flexible

rbmj:
These patches increase compatibility so CTFd can be deployed in more environments.

The first patch adds the capability for CTFd to be deployed with a SCRIPT_ROOT other than '/', allowing a CTFd instance to be placed as a server block/vhost under e.g. '/ctf'.

The second patch changes many of the SQL queries to be compatible with more strict SQL environments such as PostgreSQL, as currently many of the queries (specifically GROUP_BY) needed some redesigning in order to comply with the SQL standard.  Everything should be fine, with the only thing I'm slightly concerned about with compatibility being ``admin_team`` changing from 1/0 to True/False (needed by psycopg2).

The third patch is simple, and just adds an example uwsgi configuration file to help explain deployment under uwsgi+nginx, as the wiki could be more clear on how to accomplish this and it took a bit of trial and error.


ColdHeat:
This is an impressive commit and I thank you for taking the time to go through it. 

Thank you very much. I'm going through it right now.

ColdHeat:
All of these changes with 

```javascript
var server_root = $(".server-root-uri")[0].id
```
don't make a lot of sense b/c you can just add something like

```html
<script>var server_root = "{{ request.script_root }}"</script>
```
into base.html

CTFd/templates/admin/teams.html
```diff
@@ -123,10 +123,10 @@ <h2 class="text-center">Edit User</h2>
             {% for team in teams %}
             <tr name="{{ team.id }}">
                 <td class="team-id">{{ team.id }}</td>
-                <td class="team-name"><a href="/admin/team/{{ team.id }}">{{ team.name | truncate(32) }}</a>
+                <td class="team-name"><a href="{{ request.script_root }}/admin/team/{{ team.id }}">{{ team.name | truncate(32) }}</a>
                 </td>
                 <td class="team-email">{{ team.email | truncate(32) }}</td>
-                <td class="team-website">{% if team.website and team.website.startswith('http') %}<a href="{{ team.website }}">{{ team.website | truncate(32) }}</a>{% endif %}
+                <td class="team-website">{% if team.website and team.website.startswith('http') %}<a href="{{ request.script_root }}{{ team.website }}">{{ team.website | truncate(32) }}</a>{% endif %}
```
ColdHeat: This should not have a `{{ request.script_root }}` as it's a user provided website.

CTFd/templates/teams.html
```diff
@@ -22,8 +22,8 @@
         <tbody>
         {% for team in teams %}
             <tr>
-                <td><a href="/team/{{ team.id }}">{{ team.name }}</a></td>
-                <td>{% if team.website and team.website.startswith('http://') %}<a href="{{ team.website }}">{{ team.website }}</a>{% endif %}</td>
+                <td><a href="{{ request.script_root }}/team/{{ team.id }}">{{ team.name }}</a></td>
+                <td>{% if team.website and team.website.startswith('http://') %}<a href="{{ request.script_root }}{{ team.website }}">{{ team.website }}</a>{% endif %}</td>
```
ColdHeat: This should not have a `{{ request.script_root }}` as it's a user provided website.

CTFd/challenges.py
```diff
@@ -36,7 +36,7 @@ def chals():
             if view_after_ctf():
                 pass
             else:
-                return redirect('/')
+                return redirect(url_for('/'))
```
ColdHeat: Not sure this is valid, but if you specify the root in `config.py` you should be able to pull that value and redirect to it.

CTFd/templates/admin/editor.html
```diff
@@ -14,7 +14,7 @@
         <div class="large-8 large-centered columns">
             <div data-alert class="alert-box alert radius centered text-center">
                 <span>{{ error }}</span>
-                <a href="#" class="close">×</a>
+                <a href="{{ request.script_root }}#" class="close">×</a>
```
ColdHeat: Shouldn't have a `{{ request.script_root }}`

CTFd/templates/admin/chals.html
```diff
@@ -146,10 +146,10 @@ <h3 class="chal-title text-center"></h3>
...
 
                         <div class="form-group">
-                            <a href="#" data-toggle="modal" data-target="#update-tags" class="btn btn-primary">Tags</a>
-                            <a href="#" data-toggle="modal" data-target="#update-files" class="btn btn-primary">Files</a>
-                            <a href="#" data-toggle="modal" data-target="#update-keys" class="btn btn-primary">Keys</a>
-                            <a href="#" data-toggle="modal" data-target="#delete-chal" class="btn btn-danger">Delete</a>
+                            <a href="{{ request.script_root }}#" data-toggle="modal" data-target="#update-tags" class="btn btn-primary">Tags</a>
+                            <a href="{{ request.script_root }}#" data-toggle="modal" data-target="#update-files" class="btn btn-primary">Files</a>
+                            <a href="{{ request.script_root }}#" data-toggle="modal" data-target="#update-keys" class="btn btn-primary">Keys</a>
+                            <a href="{{ request.script_root }}#" data-toggle="modal" data-target="#delete-chal" class="btn btn-danger">Delete</a>
```
ColdHeat: None of these should have a `{{ request.script_root }}`

CTFd/templates/admin/chals.html
```diff
@@ -190,8 +190,8 @@ <h3 class="chal-title text-center"></h3>
                     <h3>Keys</h3>
                 </div>
                 <div class="modal-body">
-                    <form method="POST" action="/admin/keys" style="text-align:center">
-                        <a href="#" id="create-key" class="btn btn-primary" style="margin-bottom:15px;">New Key</a>
+                    <form method="POST" action="{{ request.script_root }}/admin/keys" style="text-align:center">
+                        <a href="{{ request.script_root }}#" id="create-key" class="btn btn-primary" style="margin-bottom:15px;">New Key</a>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/admin/chals.html
```diff
@@ -279,7 +279,7 @@ <h3 class="chal-title text-center"></h3>
                     </div>
                     <br/>
 
-                    <a href="#" id="submit-tags" class="button">Update</a>
+                    <a href="{{ request.script_root }}#" id="submit-tags" class="button">Update</a>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/scoreboard.py
```diff
@@ -5,57 +5,45 @@
 
 scoreboard = Blueprint('scoreboard', __name__)
 
+def get_standings(admin=False, count=-1):
```
ColdHeat: `-1` doesn't make much sense here. Better to simply go with and check for `None`

Hypersonic: You should default count to `None` here, not -1 (and also change the `if count == -1` check later to an `if count is not None`)

CTFd/admin.py
```diff
@@ -711,8 +697,14 @@ def admin_stats():
     wrong_count = db.session.query(db.func.count(WrongKeys.id)).first()[0]
     solve_count = db.session.query(db.func.count(Solves.id)).first()[0]
     challenge_count = db.session.query(db.func.count(Challenges.id)).first()[0]
-    most_solved_chal = Solves.query.add_columns(db.func.count(Solves.chalid).label('solves')).group_by(Solves.chalid).order_by('solves DESC').first()
-    least_solved_chal = Challenges.query.add_columns(db.func.count(Solves.chalid).label('solves')).outerjoin(Solves).group_by(Challenges.id).order_by('solves ASC').first()
+    most_solved_chal_sub = db.session.query(Solves.chalid, db.func.count(Solves.chalid).label('solves')).group_by(Solves.chalid).subquery()
+    most_solved_chal = Challenges.query.add_columns(coalesce(most_solved_chal_sub.columns.solves, 0).label('solves')) \
+        .join(most_solved_chal_sub, most_solved_chal_sub.columns.chalid == Challenges.id) \
+        .order_by(most_solved_chal_sub.columns.solves.desc().nullslast()).first()
```
ColdHeat: `.nullslast()` seems to break SQLite...

image

CTFd/admin.py
```diff
@@ -711,8 +697,14 @@ def admin_stats():
...
+    least_solved_chal_sub = db.session.query(Solves.chalid, db.func.count(Solves.chalid).label('solves')).group_by(Solves.chalid).subquery()
+    least_solved_chal = Challenges.query.add_columns(coalesce(most_solved_chal_sub.columns.solves, 0).label('solves')) \
+        .outerjoin(most_solved_chal_sub, most_solved_chal_sub.columns.chalid == Challenges.id) \
+        .order_by(most_solved_chal_sub.columns.solves.nullsfirst()).first()
```
ColdHeat: `.nullslast()` seems to break SQLite...

image

ctfd.ini
```diff
@@ -0,0 +1,35 @@
...
+master = true
+processes = 1
+threads = 1
+
+vacuum = true
+
+# Where you've put CTFD
+chdir = /var/www/ctfd/
+# Change /ctf to whatever SCRIPT_ROOT you want
+mount = /ctf=wsgi.py
```
ColdHeat: Can you make this not default to hosting in `/ctf`

CTFd/admin.py
ColdHeat: Also I'm not sure what these subqueries are for. Can you elaborate as I can't see the change visually since sqlite is broken...

CTFd/admin.py
```diff
@@ -452,10 +455,11 @@ def admin_team(teamid):
     elif request.method == 'POST':
         admin_user = request.form.get('admin', None)
         if admin_user:
-            admin_user = 1 if admin_user == "true" else 0
+            admin_user = True if admin_user == 'true' else False
             user.admin = admin_user
-            user.banned = admin_user
+            #user.banned = admin_user
```
ColdHeat: Can you uncomment this? Hiding admins from the scoreboard is something CTFd does.

CTFd/templates/admin/config.html
```diff
@@ -29,10 +29,10 @@
 
             <ul class="nav nav-tabs" role="tablist">
                 <li role="presentation" class="active">
-                    <a href="#mailserver" aria-controls="mailserver" role="tab" data-toggle="tab">Mail Server</a>
+                    <a href="{{ request.script_root }}#mailserver" aria-controls="mailserver" role="tab" data-toggle="tab">Mail Server</a>
                 </li>
                 <li role="presentation">
-                    <a href="#mailgun" aria-controls="mailgun" role="tab" data-toggle="tab">Mailgun</a>
+                    <a href="{{ request.script_root }}#mailgun" aria-controls="mailgun" role="tab" data-toggle="tab">Mailgun</a>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/admin/config.html
```diff
@@ -29,10 +29,10 @@
 
             <ul class="nav nav-tabs" role="tablist">
                 <li role="presentation" class="active">
-                    <a href="#mailserver" aria-controls="mailserver" role="tab" data-toggle="tab">Mail Server</a>
+                    <a href="{{ request.script_root }}#mailserver" aria-controls="mailserver" role="tab" data-toggle="tab">Mail Server</a>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/admin/config.html
```diff
@@ -106,10 +106,10 @@
             <div>
             <ul class="nav nav-tabs" role="tablist">
                 <li role="presentation" class="active">
-                    <a href="#start-date" aria-controls="start-date" role="tab" data-toggle="tab">Start Time</a>
+                    <a href="{{ request.script_root }}#start-date" aria-controls="start-date" role="tab" data-toggle="tab">Start Time</a>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/admin/config.html
```diff
@@ -106,10 +106,10 @@
...
                 </li>
                 <li role="presentation">
-                    <a href="#end-date" aria-controls="end-date" role="tab" data-toggle="tab">End Time</a>
+                    <a href="{{ request.script_root }}#end-date" aria-controls="end-date" role="tab" data-toggle="tab">End Time</a>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/chals.html
```diff
@@ -90,8 +90,8 @@
         <div class="modal-content content">
             <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
             <ul class="nav nav-tabs" role="tablist">
-                <li role="presentation" class="active"><a href="#challenge" aria-controls="challenge" role="tab">Challenge</a></li>
-                <li role="presentation"><a href="#solves" aria-controls="solves" class="chal-solves" role="tab"></a></li>
+                <li role="presentation" class="active"><a href="{{ request.script_root }}#challenge" aria-controls="challenge" role="tab">Challenge</a></li>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/chals.html
```diff
@@ -90,8 +90,8 @@
...
+                <li role="presentation"><a href="{{ request.script_root }}#solves" aria-controls="solves" class="chal-solves" role="tab"></a></li>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/admin/chals.html
```diff
@@ -42,8 +42,8 @@
                         </div>
 
                         <ul class="nav nav-tabs" role="tablist" id="new-desc-edit">
-                            <li role="presentation" class="active"><a href="#new-desc-write" aria-controls="home" role="tab" data-toggle="tab">Write</a></li>
-                            <li role="presentation"><a href="#new-desc-preview" aria-controls="home" role="tab" data-toggle="tab">Preview</a></li>
+                            <li role="presentation" class="active"><a href="{{ request.script_root }}#new-desc-write" aria-controls="home" role="tab" data-toggle="tab">Write</a></li>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/admin/chals.html
```diff
@@ -42,8 +42,8 @@
...
+                            <li role="presentation"><a href="{{ request.script_root }}#new-desc-preview" aria-controls="home" role="tab" data-toggle="tab">Preview</a></li>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/admin/chals.html
```diff
@@ -118,8 +118,8 @@ <h3 class="chal-title text-center"></h3>
                         </div>
 
                         <ul class="nav nav-tabs" role="tablist" id="desc-edit">
-                            <li role="presentation" class="active"><a href="#desc-write" id="desc-write-link" aria-controls="home" role="tab" data-toggle="tab">Write</a></li>
-                            <li role="presentation"><a href="#desc-preview" aria-controls="home" role="tab" data-toggle="tab">Preview</a></li>
+                            <li role="presentation" class="active"><a href="{{ request.script_root }}#desc-write" id="desc-write-link" aria-controls="home" role="tab" data-toggle="tab">Write</a></li>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

CTFd/templates/admin/chals.html
```diff
@@ -118,8 +118,8 @@ <h3 class="chal-title text-center"></h3>
...
+                            <li role="presentation"><a href="{{ request.script_root }}#desc-preview" aria-controls="home" role="tab" data-toggle="tab">Preview</a></li>
```
ColdHeat: This shouldn't have a `{{ request.script_root }}`

rbmj:
Ugh, that's what happens when you get lazy and start changing files using sed.

I'll try to fix these, thanks for the QA

ctfd.ini
rbmj: It's mostly meant as an example (not just to be blindly copied/pasted), but yes I can change

CTFd/admin.py
rbmj: Subqueries are needed under PostgreSQL (and I imagine others - if anyone ever decided to host under MSSQL haha) because in standard SQL when you do a SELECT ... GROUP BY all fields selected need to be in the GROUP BY clause or part of an aggregate function.  Thus, a SELECT * approach won't work, and you need to run a subquery and then join with the other table.  Under a reasonable SQL server, the query plan should be the same.

CTFd/admin.py
rbmj: Ok, it just looked funny.  Perhaps there should be a comment, as it wasn't clear?

I'll add.

ColdHeat:
Also have some general comments about some of the functions. Overall, thanks for the help :)

CTFd/scoreboard.py
rbmj: You're right, fixing.

CTFd/challenges.py
rbmj: You actually don't want to specify the root in ``config.py`` - you want that to be handled by uwsgi.  If you specify ``APPLICATION_ROOT`` it causes issues with the session cookie

rbmj:
Alright, I've fixed most of the issues where server_root is referenced when it shouldn't.  I ran a grep on ``}}#`` and it came up empty.

Still to be resolved:
 - unnecessary DOM querying
 - SQLite compatibility (``nullslast()`` crashing)
 - redirects on ``'/'``

rbmj:
I've run into issues inside of certain script blocks (specifically graphs.html) with ``{{ request.server_root }}`` not getting replaced properly (it just appears to become the empty string).  I'm not well versed enough with flask/jinja2 to understand why this is happening, but I'll work on transitioning javscript all over to using a global variable.

I've read that jinja2 treats things differently inside script blocks, so I don't know if this is part of that behavior or not.  I'm not experienced enough with webdev or flask to know what's going on here (I'm mostly a backend guy).

CTFd/challenges.py
ColdHeat: But that seems too uwsgi specific, when you could also deploy with gunicorn or Tornado or just the Flask server

ColdHeat:
http://flask.pocoo.org/docs/0.10/patterns/jquery/#where-is-my-site

This discussion might help you, seems to need some filters/options for Jinja2.

CTFd/admin.py
rbmj: I don't know if this is how sqlalchemy works, but it seems like if you use coalesce then you should be able to order by the result column ``solves`` since those will already be replaced with zeros.

I'd test it, but my connection to our server just died - looks like a random power failure or an issue with our modem - so it'll have to wait until I can get in and debug what's going on with our server.

CTFd/challenges.py
rbmj: I don't have experience deploying under gunicorn, so I don't know how that's supposed to be handled.

Based on my limited experience, it appears that you want ``SCRIPT_NAME`` and ``PATH_INFO`` to be managed at the wsgi level (whether that be uwsgi, gunicorn, or tornado) - I'm just not sure how that works for each of the different setups.

I agree that it smells a bit, but until I get a chance to try under gunicorn I'm inclined to go with behavior that is known good with uwsgi (which seems to be a somewhat standard/supported deployment option at the distro level)

rbmj:
The last two commits should fix the concerns you raised

rbmj:
Apologize for the couple of extraneous commits - the changes seemed trivial at the time but apparently I can't do HTML properly.

ColdHeat:
I can't review this right now but you can do a git squash to combine all of your commits into one.

rbmj:
Rebased and squashed

ColdHeat:
Still have the issue with `url_for`

image

rbmj:
Fixed/squashed/updated

I just manually prepended request.script_root to the path.

Alternatively, you can use ``url_for('views.static_html')`` and ``url_for('admin.list_container')``.  If you'd rather do that, let me know

ColdHeat:
Hmm would `url_for('views.static_html')` work? 

I don't have a proper setup for testing hosting out of a folder so can you test if that would work properly? If so then definitely replace it with that. 

I think this is looking pretty good. I'll test some stuff out when I can and merge eventually.

rbmj:
I tested both on my instance - I'll go ahead and change it over

rbmj:
Fixed/tested/squashed

Also made one more change to the team view in admin (another SQL group_by issue).

I've grepped all files for group_by and came up blank, so all queries should be fixed.

ColdHeat:
I have some other niggling gripes but they are overall less important than the general ideas. I think this is good to merge.

rbmj:
Awesome - just let me know what your gripes are and I'll address.

ColdHeat:
Already merged :)

rbmj:
Thanks - what I meant was that if you have some cleanup or changes you'd like me to make for a future pull request, just let me know.  Happy to help if I can.

PR #126, opened 2016-06-23
Various small fixes

rbmj:
1.  Fix the deprecation warnings by changing from `flask.ext.$name` to `flask_$name`

 The most recent version of flask has deprecated the redirect namespace, so this is just cleanup.  This should only cause compatibility issues if someone is using an *ancient* version of a flask extension (which they shouldn't be, as you recommend using pip to install things).

2.  Be consistent with import style in admin.py

 One liner change from #125 so that it's consistent with the rest of the file.  Purely cosmetic.

3.  Fix a residual script_root fix

 These keep popping up as my friends do more usability testing.  Again, one liner.

4.  Store IPs in 32 bits again

 This is the only potentially breaking change.  Switching it to a 64 bit integer fixed non-SQLite databases (since the SQLAlchemy integer datatype is the standard SQL integer, which is signed), but is space inefficient and feels quite wrong.  It also was compatible with SQLite since data types and constraints don't actually exist in SQLite.  This change moves it back to 32 bits (which breaks non SQLite, but I believe I'm the only person running this, and I've transitioned my DB).  The difference is the IP address is now stored as a signed 32 bit integer, so it'll fit into the standard SQL integer.

 I've maintained backwards compatibility with old SQLite databases by wrapping the `struct.pack` call in a try-except that will fall back to the old unsigned data type if the result from the database is out of range.  IP addresses that fall within the signed range will be unaffected as well.

I tested everything that could forseeably be affected on my system, and everything works.

### DIFF
--- a/CTFd/__init__.py
+++ b/CTFd/__init__.py
@@ -1,7 +1,7 @@
 from flask import Flask, render_template, request, redirect, abort, session, jsonify, json as json_mod, url_for
-from flask.ext.sqlalchemy import SQLAlchemy
+from flask_sqlalchemy import SQLAlchemy
 from logging.handlers import RotatingFileHandler
-from flask.ext.session import Session
+from flask_session import Session
 from sqlalchemy_utils import database_exists, create_database
 import os
 import sqlalchemy
--- a/CTFd/admin.py
+++ b/CTFd/admin.py
@@ -1,6 +1,7 @@
 from flask import render_template, request, redirect, abort, jsonify, url_for, session, Blueprint
 from CTFd.utils import sha512, is_safe_url, authed, admins_only, is_admin, unix_time, unix_time_millis, get_config, set_config, sendmail, rmdir, create_image, delete_image, run_image, container_status, container_ports, container_stop, container_start
 from CTFd.models import db, Teams, Solves, Awards, Containers, Challenges, WrongKeys, Keys, Tags, Files, Tracking, Pages, Config, DatabaseError
+from CTFd.scoreboard import get_standings
 from itsdangerous import TimedSerializer, BadTimeSignature
 from sqlalchemy.sql import and_, or_, not_
 from sqlalchemy.sql.expression import union_all
@@ -18,8 +19,6 @@ import os
 import json
 import datetime
 import calendar
-
-from scoreboard import get_standings
 
 admin = Blueprint('admin', __name__)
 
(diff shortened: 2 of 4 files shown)


ColdHeat:
You're definitely not the only person using an actual database but I did not run into problems using MySQL...

Perhaps the answer is to not use an integer and instead use a string? http://sqlalchemy-utils.readthedocs.io/en/latest/_modules/sqlalchemy_utils/types/ip_address.html

I am almost definitely sure that `!I` is the format we want (https://github.com/google/ipaddr-py/blob/master/ipaddr.py#L122-L137) but I do see where you're getting at with the `!i`.

Let me test some things and get back to you.

rbmj:
MySQL is a strange beast... Integers in SQL are signed.  MySQL offers unsigned integers as an extension, but these aren't accessible from SQLAlchemy without some dialect plugin wizardry.  It would appear that either the MySQL plugin or the database engine itself is being "smart" (depending on your definition of smart) and doing some type conversions for you, even though the column isn't declared as unsigned.

Since signedness or integer width don't appear to be strictly enforced by SQLAlchemy, this doesn't appear to be a breaking change for MySQL.  YMMV, but if you don't need your tracking information you're best off just renaming the table to e.g. tracking_bak so it'll recreate with the new schema. SQLite shouldn't break because data types are a lie there.

Using a signed integer can still represent all IP addresses (just some will be negative), so I don't see any need to change to storing as a string.  PostgreSQL has a native IP address type as well, but I'm sure you don't want to depend on that.  If you're running CSAW off this your tracking database might get rather large, so I'd imagine that 16 bytes (string) per record vs. 4 could cause some decent storage savings on disk, at the cost of making generating statistics almost unnoticeably more expensive.

I initially changed to bigint because I was getting constraint failures on PostgreSQL, and then changed back to int and just modified the signedness when I realized that it would still work and not waste 4 bytes unnecessarily.

rbmj:
Also, a couple of things just to let you know:

I'm currently working on an autobahn-based extension to allow programs to be served over websockets.  This is important for my team specifically because our school has very strict egress restrictions (HTTPS only, no high numbered ports), and we only have one room on campus that has unfiltered internet that we can compete from.  My use case is running an internal CTF that people can play from their rooms.

It'll be autobahn based (so an additional dependency, plus python3), but Flask/WSGI's support of websockets wasn't deployment option portable or (AFAICT) sufficiently asynchronous.  So, it would run as a separate process, and be placed on e.g. `/socket` via nginx's `proxy_pass`.  It'll take a while, and I'm not great with UX, but it should be at least be an interesting toy project.

Also, I wrote a couple of pages in the wiki on deploying onto a non-root `SCRIPT_ROOT`, and these might serve as more comprehensive documentation on deploying using uWSGI.  I'm thinking of changing the suggested config to not include a `threads` setting (to disable GIL) and increase the number of worker processes (as everything *should* just work, though I don't have the means to stress test this).  This is based on guidance from [here](http://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html).

I'm interested to know what sort of scaling you all do in production, as I've never had the experience with deploying non-trivial web applications at scale.

ColdHeat:
I am also sort of thinking about how to do Websockets with CTFd. I think Flask-Websockets is better suited but if you have an implementation you like I'd love to see it. 

In general there's not a lot of tricks at scale for Python that don't apply in other languages. Load balance between web servers and have some replicated databases. While that might not work for really large deployments it works for me. CSAW doesn't even really load balance too much but we're looking to solve some of those problems. 

I think your changes are fine so I'm going to merge them :)

ColdHeat:
Also perhaps this would be useful:

https://github.com/joewalnes/websocketd

Issue #136, opened 2016-08-09
Exception when displaying IP on Team Detail page in Admin Mode


snurilov:
When attempting to display a detailed view of a Team in the Admin Mode, if the first octet of the IP address is bigger than 127, an exception is generated. (e.g. 192.168.0.1)

This happens because the IP address is interpreted as an signed long and the call to inet_ntoa fails in the long2ip function in util.py.

ColdHeat:
I think I know what the issue+fix is but can I get a stack trace just so I can be sure

ColdHeat:
Not sure that's the best fix in this case as the issue is due to the conflict with the long2ip function in models.py

ColdHeat:
Can you take a look at a4e51fcc761afea4a28ebe6968dfdab4dd04556c and make sure it fixes your issue?

snurilov:
Looks like it fixed the issue.

PR #137, opened 2016-08-09
Fix issue reported in #136

snurilov:
Prevent IP addresses from being treated as a negative number.

Cast to an unsigned INT.


ColdHeat:
Fix already implemented in a different way.

PR #150, opened 2016-09-03
Fix MySQL availability tries counter

nashe:
The variable `i` was not declared and never incremented.

The behaviour was the right one (waiting for MySQL to be listening before starting the application) but was displaying errors.


ColdHeat:
Did a cursory review here but this looks fine. Thanks!

PR #151, opened 2016-09-08
Fix deleting wrong keys from admin panel

itsZN:
- Fixed wrong_keys.html template using WrongKeys.team instead of
  WrongKeys.teamid
- Added the id of the wrong key to the wrong_keys.html template entries
- Changed /admin/wrong_keys to only take the key id instead of team and
  challenge id, to allow deleting the intended key when a team has more
  than one wrong key for a challenge. (Previously it would delete only
  the first wrong key entry for that challenge)

### DIFF
--- a/CTFd/admin.py
+++ b/CTFd/admin.py
@@ -705,14 +705,17 @@ def delete_solve(teamid, chalid):
     return '1'
 
 
-@admin.route('/admin/wrong_keys/<teamid>/<chalid>/delete', methods=['POST'])
+@admin.route('/admin/wrong_keys/<keyid>/delete', methods=['POST'])
 @admins_only
-def delete_wrong_key(teamid, chalid):
-    wrong_key = WrongKeys.query.filter_by(teamid=teamid, chalid=chalid).first()
+def delete_wrong_key(keyid):
+    wrong_key = WrongKeys.query.filter_by(id=keyid).first()
+    if (wrong_key == None):
+        return '0'
     db.session.delete(wrong_key)
     db.session.commit()
     db.session.close()
     return '1'
+
 
 
 @admin.route('/admin/statistics', methods=['GET'])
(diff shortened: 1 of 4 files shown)


CTFd/admin.py
```diff
@@ -705,16 +705,19 @@ def delete_solve(teamid, chalid):
     return '1'
 
 
-@admin.route('/admin/wrong_keys/<teamid>/<chalid>/delete', methods=['POST'])
+@admin.route('/admin/wrong_keys/<keyid>/delete', methods=['POST'])
 @admins_only
-def delete_wrong_key(teamid, chalid):
-    wrong_key = WrongKeys.query.filter_by(teamid=teamid, chalid=chalid).first()
+def delete_wrong_key(keyid):
+    wrong_key = WrongKeys.query.filter_by(id=keyid).first()
```
ColdHeat: This should probably be `first_or_404()`

CTFd/templates/original/admin/wrong_keys.html
```diff
@@ -98,11 +98,11 @@
             })
         });
 
-        function load_confirm_modal(team, chal, team_name, chal_name) {
+        function load_confirm_modal(keyId, team_name, chal_name) {
```
ColdHeat: Can we use underscores instead of camelcase?

CTFd/templates/original/admin/wrong_keys.html
```diff
@@ -98,11 +98,11 @@
...
             var modal = $('#confirm')
             modal.find('#confirm-team-name').text(team_name);
             modal.find('#confirm-chal-name').text(chal_name);
-            $('#confirm form').attr('action', '{{ request.script_root }}/admin/wrong_keys/' + team + '/' + chal + '/delete');
+            $('#confirm form').attr('action', '{{ request.script_root }}/admin/wrong_keys/' + keyId + '/delete');
```
ColdHeat: Can we use underscores instead of camelcase?

CTFd/templates/original/admin/wrong_keys.html
```diff
@@ -112,7 +112,8 @@
             var chal_name = elem.find('.chal').text().trim();
             var team = elem.find('.team').attr('id');
             var team_name = elem.find('.team').text().trim();
-            load_confirm_modal(team, chal, team_name, chal_name);
+            var keyId = elem.find('.flag').attr('id');
```
ColdHeat: Can we use underscores instead of camelcase?

ColdHeat:
Hey sorry itsZN, I forgot to review your PR. I went ahead and commented on a few lines. Nothing too nitpicky, just some suggestions.

ColdHeat:
This breaks deleting wrong keys from the `/admin/team/<teamid>` area and also causes some differences in how the backend looks for deleting wrong keys vs deleting solves.

ColdHeat:
Thanks for bringing this to my attention and sorry about the delay!

PR #162, opened 2016-10-27
Added documentation for mysql in docker + formatting

ps1337:
I added documentation on how to run CTFd along with a dockerized mysql and how to set it up. This allows users to quickly check out CTFd with a running mysql.


ColdHeat:
No longer applicable since we have `docker-compose`

PR #166, opened 2016-11-12
Gender-neutral alternative

ldionmarcil:
Gender-neutralism is good practice. Diversity etc.

### DIFF
--- a/CTFd/auth.py
+++ b/CTFd/auth.py
@@ -37,7 +37,7 @@ def confirm_user(data=None):
         if authed():
             return redirect(url_for('challenges.challenges_view'))
         return redirect(url_for('auth.login'))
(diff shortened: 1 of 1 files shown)


tjgerot:
As seemingly trivial as it may be, it deserves a merge.

Owlz:
For reference, this should be "his or her". "Their" is a plural while the rest of the sentence is not.

rkkautsar:
@Owlz see [Singular they](https://en.wikipedia.org/wiki/Singular_they)

ColdHeat:
It was a [CSAW CTF](https://csaw.engineering.nyu.edu/ctf) challenge at some point to submit a PR to CTFd. This solved the challenge but the comment it referred to no longer exists so I'm going to close this.

PR #170, opened 2016-11-30
Changed to dyniamic argument of mail function

seadog007:


ColdHeat:
Thanks @seadog007! 

SMTP has sort of been forgotten so thanks for submitting this!

PR #171, opened 2016-12-05
Added side nav pills to config page

nategraf:
Closes: 155


ColdHeat:
Looks great! 

Some niggling things but I think I can handle integrating those.

Issue #169, opened 2016-11-23
updatesolves can conflict with names


ColdHeat:
The JS function updatesolves in the original theme can mark solves incorrectly if two challenges have the same name.

PR #174, opened 2016-12-07
Reference solves by id (#169)

nategraf:
Changed the /chals/colves endpoint and associated javascript to
index solves by challenge id rather than by challenge name

Closes (#169)


ColdHeat:
Thanks @nategraf, looks good! 

I might want to change the endpoint to also report the name but not a huge concern 👍

PR #188, opened 2017-01-08
Display country flag in the scoreboard and Improve country selection

arthaud:
List of changes:

* The scoreboard now shows the flag picture of the country of each team
* Each team can change his country via a `<select>` with the list of all countries, rather than an `<input>`
* Add the fields website, affiliation, country to the registration form


tjgerot:
This would be a wonderful optional feature. Why was it closed?

ColdHeat:
I had discussed this with arthaud about the implementation. For one thing it's not necessary to have every flag be a PNG. The flags can be an SVG with some CSS. Also the inclusion of the GPL is a bit of a turn off. 

Also I would prefer to not have the country be required on registration. 

I will implement this feature soon personally so that it lines up with what I'd like to see.

PR #191, opened 2017-01-15
Add a populating script for awards.

palkeo:


ColdHeat:
Seems like this generates negative and positive award values? 

I dont think that's an issue but I want to think if that's the right idea here. 

In addition this seems to knock loose some issues with the Team template so I'll patch that up in master.

palkeo:
Yes, the reasoning was that it could make sense to attribute negative award to people cheating, or whatever… But we can change it to be always positive, if you want :)

ColdHeat:
I think this is good to merge into testing. 

If anything I need to make the populate script do a little more so that it can be used for more than just generate testing data.

PR #192, opened 2017-01-16
Use <int:xxx> in routes to prevent some errors 500

arthaud:
This is basically to prevent errors 500 (for instance, when visiting /teams/1xxxx, /team/1xxx or /admin/team/1xxx)


ColdHeat:
I used this Flask feature in the past and it would make debugging a little difficult but I can't think of a valid reason to not merge this so I will merge into testing for now. 

Thanks again for the PRs!

PR #193, opened 2017-01-16
Use first_or_404() to prevent some errors 500

arthaud:
This fixes some other errors that can be triggered by using invalid IDs in URLs.


ColdHeat:
I once again do not have a concrete reason to be against this although I have been against first_or_404() as it abstracts away some logic that I feel is useful for debugging. 

But since I have not myself written the logic I could definitely benefit from its usage. Merging into testing.

PR #199, opened 2017-02-02
debian + archlinux install

noraj1337:
No more need for prepare.sh or need to make a smart menu prompting which os the user is running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,27 @@
 
 CTFd is a CTF in a can. Easily modifiable and has everything you need to run a jeopardy style CTF.
 
-Install: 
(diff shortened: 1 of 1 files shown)


ColdHeat:
Changes to the setup process need to be very streamlined. I would much rather see `prepare.sh` be improved to understand what distro it's on and install requirements that way

noraj1337:
@ColdHeat : Ok but you can add 
1. Clone the CTFd repository: `git clone https://github.com/isislab/CTFd.git ctfd`
2. Then `cd ctfd`

Because that lacks

PR #200, opened 2017-02-02
apt-get + pacman automatic install

noraj:


ColdHeat:
A step in the right direction but once again I would rathere see that prepare.sh automatically understood what package manager to use or detected the distro in some way. More user interaction is not something that I would like to see in the setup process.

noraj1337:
@ColdHeat : I agree with you. But for now it's beter than nothing. If you want something more smart that detect OS or package manager without user interaction, write t in a TODO list and wait for someone to do it :) I have no time right now.

ColdHeat:
If you have the time to write something that auto detected which package manager was installed I'll merge it in but as it is right now, I would rather the user not have to specify something that a script can automatically do.

noraj1337:
@ColdHeat : now supporting automatic detection of package manager

ColdHeat:
I'm not sure I agree with the idea of trying to run the program to detect if it's there. Can't you use a shell built in like `type` or `command`?

Also please add the commands for your unsupported package managers or remove them. It's somewhat annoying to just have an unsupported package manager. 

Perhaps add `apk` for alpine?

noraj1337:
@ColdHeat 

> I'm not sure I agree with the idea of trying to run the program to detect if it's there. Can't you use a shell built in like type or command?

Can you tell me why trying to run the program is bad? The `--version` just display the version and exit the program. I don't see any of using `type`. Furthermore `type` will works for a broken program but trying to run the program will tell us if it is broken or not. So I think what I do is better.

> Also please add the commands for your unsupported package managers or remove them. It's somewhat annoying to just have an unsupported package manager.

No. I only have archlinux, I won't install all others distribution just to test if the package manager commands. And I won't remove them either. This is a skeleton for people who want to contribute.

> Perhaps add apk for alpine?

I can but there is a lot of others package managers and I can't add them all. People who need an installation for their specific distribution are welcome to contribute and add it in the script.

I think my script/PR is ready to be accepted and is better than the previous debian only version.

ColdHeat:
Well for one thing it's simpler to keep a function which checks if a program exists while using `type` or `command -v` and checking the status code. 

It makes sense to not add in every package managers but if I was to commit this to master I would aim to support common package managers. `apk` as evidenced by some of the docker PRs is a commonly used package manager. 

You can use [pacapt](https://github.com/icy/pacapt/blob/master/pacapt) for inspiration.

noraj1337:
Sorry I won't. I explained why. So close the PR and don't merge it. I'm sadden having spending time on this project.

PR #202, opened 2017-02-06
Fix exception on cofirmation screen

snurilov:
When a user attempts to confirm an e-mail address, an exception is thrown because the db session is closed prior to logging.

The line db.session.close() has to move after the logging, otherwise the team parameters from the orm object are discarded and an exception is thrown.

Closing the session after logging, fixes the issue.


ColdHeat:
Thanks for the PR!

PR #204, opened 2017-02-13
Docker enhancements

erasche:
- Use python:2.7-alpine as the base image to reduce size
- Include a script to completely configure the CTF from a mounted yaml file (admin, settings, challenges, etc)
- Add readme section mentioning docker configuration

### DIFF
--- a/.sample.yml
+++ b/.sample.yml
@@ -1,0 +1,48 @@
+meta:
+    ctf_name: My CTF
+    ctf_theme: original
+    #max attempts per challenge
+    max_tries: 0
+    ## Challenges cannot be viewed by unregistered users
+    view_challenges_unregistered: True
+    ## Allow/Disallow registration
+    prevent_registration: False
+    ## Verify emails
+    verify_emails: False
+    prevent_name_change: True
+    # Mail Settings
+    mail_server: None
+    mail_port: None
+    mail_tls: None
+    mail_ssl: None
+    mail_username: None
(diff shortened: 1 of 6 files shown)


ColdHeat:
Thanks for the PR!

Ultimately I think exporting and populating a CTF from a yaml file is a good idea and I'm pretty sure there's an issue for it but I think it's simpler to support JSON. I would rather see JSON and YAML supported and I would prefer that the ability to export as well as import would be there. 

If you could seperate your docker changes into a separate PR (minus the import checking) I'll merge that in more quickly as it does solve a pain point (https://github.com/isislab/CTFd/issues/187).

ColdHeat:
Also I'd prefer if you left Docker as Ubuntu for now. The docker image might be really big but it makes things a little cleaner in terms of the setup scripts. 

I'll weigh the pros and cons about alpine later but for now I think Ubuntu is fine for most users.

erasche:
Hi @ColdHeat,

I'm afraid I don't have the time to make these changes to this PR any time in the next 2-3 months. If you want these changes, please feel free to cherry pick the subset you like and close the PR.

I chose YAML because it supports comments (nice for documenting examples that an end user will be modifying).

cynicXer:
Ignore that.  I can't close that PR for some reason.

ColdHeat:
Very old PR, many ideas have been implemented already

PR #205, opened 2017-02-18
Fixed hard coded upload directory

nategraf:
With this commit, the upload folder configured in config.py will be used as one might expect

Hard coded upload directory name was preventing work in importing/exporting challenge specs in a smooth manner.


ColdHeat:
Hmm while this definitely makes sense with the file handler route in place, it raises some concerns for me about upgrading users. 

This overall does make sense though. 

I've just broken down admin.py into seperate smaller files, can you merge your changes into those new files?

nategraf:
What do you mean with your first sentence?

Ya, I'll go ahead and merge this change into the new structure

nategraf:
Closing PR in favor of PR #206

PR #206, opened 2017-02-19
Fixed hard coded upload directory

nategraf:
Updates of PR #205


ColdHeat:
So my original concern with updates would be that older versions of CTFd use static/uploads so to get those into the new folder they'd need to be moved. Just a small thing that I wondered about implementing.

nategraf:
Should something be changed before this PR is merged, or is it simply something you do not wish to change?

ColdHeat:
Sorry was just working on other things and this ultimately didn't meet the goals of what I wanted to accomplish a few days ago. 

I will be taking a closer look at this soon.

PR #208, opened 2017-02-23
add scoreboard freeze

rkkautsar:
Sometimes I need to freeze the scoreboard for local contests, so I thought I should make a pull request.


ColdHeat:
Freezing the scoreboard makes the scoreboard only provide values up until the freeze time?

rkkautsar:
Yes, exactly. The admin can still see the whole score, though. I came from competitive programming background, and it is a common practice in contests to hide the winner before the announcement.

ColdHeat:
I agree there's value in it. I'm working on a config value to hide scores. I'm not sure if freezing is the right way to go... Is that how you've seen it done in programming competitions? I don't really play those.

rkkautsar:
Yes, for instance, here's a [standard for an ACM ICPC scoreboard](https://clics.ecs.baylor.edu/index.php/Contest_Control_System#Freezing_the_Scoreboard). But yeah, I know for CTF it's not a known practice, but it's nice to have the option for local contests.

ColdHeat:
Thought about this again and couldn't a team simply go through the individual scores of each user in this? 

How does ACM deal with situations like that?

rkkautsar:
AFAIK, we can't see individual scores of other teams in ICPC. I honestly missed that in CTFd we can see individual scores. Maybe we should just hide the submissions after freezing the scoreboard from individual scores? I'll update the PR for that

ColdHeat:
Thanks for being so proactive! 

Ultimately I think I should push up the code to hide scores and then you can base freeze time off of that since the logic is ultimately similar. Just trying to save you some brain cycles :)

rkkautsar:
Okay, thanks!
I'll wait for the commit

ColdHeat:
This is unfortunately a harder problem than it seems because of all the different areas that solves are displayed. 

I've implemented hiding scores in a branch:
https://github.com/isislab/CTFd/tree/hide_scores

But implementing a freeze is a little more complicated... If you're still interested in implementing it let me know if you need any assistance.

rkkautsar:
Sorry, haven't been able to work on this, yes there are many areas that display solves. Maybe we should just hide it from another team like you do with the hide score config? My concern is the score and place displayed in the team page, I think we should just hide it after the freeze (from both another team, or logged in team), because those are computed in the model and are hard to decouple.

rkkautsar:
I think the functionality is pretty much done, should I add some kind of indicator in the views that the scoreboard is frozen?

CTFd/models.py
```diff
@@ -159,20 +159,34 @@ def __init__(self, name, email, password):
...
-    def score(self):
+    def score(self, date_filter=None):
         score = db.func.sum(Challenges.value).label('score')
-        team = db.session.query(Solves.teamid, score).join(Teams).join(Challenges).filter(Teams.banned == False, Teams.id == self.id).group_by(Solves.teamid).first()
+        team = db.session.query(Solves.teamid, score).join(Teams).join(Challenges).filter(Teams.banned == False, Teams.id == self.id)
         award_score = db.func.sum(Awards.value).label('award_score')
-        award = db.session.query(award_score).filter_by(teamid=self.id).first()
+        award = db.session.query(award_score).filter_by(teamid=self.id)
+
+        if date_filter:
```
ColdHeat: Perhaps it's better to automatically grab the freeze time from the db instead of having it passed in?

CTFd/views.py
```diff
@@ -139,11 +139,29 @@ def team(teamid):
...
     solves = Solves.query.filter_by(teamid=teamid)
-    awards = Awards.query.filter_by(teamid=teamid).all()
-    score = user.score()
-    place = user.place()
+    awards = Awards.query.filter_by(teamid=teamid)
+
+    place = None
+    score = None
+
+    if freeze:
```
ColdHeat: I think it's a better idea to have user.place() always understand where the user's place is regardless of settings. It makes it a little easier to have config related changes propagate.

rkkautsar: Hmm I can't think of any other way to get user place before the freeze, should I just overload user.place?

rkkautsar: Ah just read the review, so I should just take account of the freeze config automatically without relying on passed parameter?

rkkautsar: Ah, what if I have an admin=False parameter like in scoreboard?

ColdHeat:
Thank you again for implementing this, I've got it from here!

PR #229, opened 2017-03-10
PEP8 lint project and add linting checks to build process

mschwager:
Linter can be manually run with `pep8 --ignore E501 CTFd/ tests/`.

Personally, I don't find E501 (lines > 80 characters) THAT useful, so I normally ignore it.


mschwager:
One thing to note, `not variable` is not the same as `variable == False`. Having `not variable` will return `True` for things like `[]` or `{}`, but `variable == False` will return `False` for those cases. If you expect there to be multiple types for those checks then we may have to rethink those changes. However, it's typically bad practice to have the possibility of different types in a conditional, so if that's the case, that may be a code smell in and of itself.

ColdHeat:
Most of those equality compares were for sqlalchemy. I'm not sure their best practice but I've seen `==` used there frequently.

Issue #235, opened 2017-04-04
Need a way to customize the challenge edit modal through plugins


nategraf:
Although the  newest changes allow for customizing the challenge create modal, the edit modal is still add within the chals.html file and cannot be overridden in a similar.

ColdHeat:
I'm pretty sure this is closed because the update.hbs file contain the modals now.

PR #236, opened 2017-04-04
challenge update modal is now replaceable

nategraf:
By defining
* [type]-challenge-update.hbs
* [type]-challenge-modals.hbs
* [type]-challenge-update.js

in the /static/admin/js/templates/challenges/[type] folder the
challenge updat modal will be defined for any challneges of
the given type. This allows for essentailly full customizability
of how you will edit custom challenge types in the admin UI.

The reason for having two files, *update.hbs and *modals.hbs, is
that *update.hbs defines the body for the main challenge update
modal, while *modals.hbs deines any addtional modals which will be
used within the main modal

There is one function which is required in *update.js is
`openchal(id)` which will be passed the id of the challenge to be
edited and should open the modal as well as load any needed data

#235


ColdHeat:
Initially it seemed like the buttons for Tags, Keys, Hints, etc, didn't work but they appear to work now...

1st thing, the secondary modals seem to fade in now which I think makes it less obvious that they're there. 2nd thing is the modals don't seem to stack now. The secondary modals close both modals. 

A little modal crazy right now.

nategraf:
Took quite a bit of time to figure out, but I eventually got to the heart of it and solved a couple issues. The really nasty one was a double counting issue caused by redundant calls to the mult-modal.js plugin. The issues mentioned above should be fixed.

ColdHeat:
Thanks for fixing that multi-modal issue it was bugging me as well elsewhere. 

This overall looks good and solves the problem of customizing the challenge UI in the admin panel but I wonder if perhaps we are too modal heavy...

I don't for see the UI changing anytime soon so overall I think this is the right way forward for now.

PR #240, opened 2017-04-08
improving get_config

ColdHeat:
Extends get_config to first query config.py for the value. Unsure if the database should come first or not...


ColdHeat:
This could raise some issues with plugins leaking internal `config.py` values if users could upload plugins to a CTFd instance. But a plugin could sort of always do that since it's provided the app instance on load...

PR #274, opened 2017-06-09
support windows

yanir3:
There is no inet_pton/inet_ntop on Windows so using win_inet_pton package to make these functions available

### DIFF
--- a/CTFd/models.py
+++ b/CTFd/models.py
@@ -1,12 +1,18 @@
 import datetime
 import hashlib
 import json
-from socket import inet_pton, inet_ntop, AF_INET, AF_INET6
+from socket import AF_INET, AF_INET6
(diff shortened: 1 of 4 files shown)


ColdHeat:
I'm sorry if it comes to requiring some external code I would rather use `netaddr` or `ipaddress` than `win_inet_pton` and checking if we're Windows or not. 

Also in the future it would be useful to separate your changes :) 

I appreciate the PR though.

yanir3:
@ColdHeat Depends, on Python 3.4 they added pton and ntop for Windows. But I suppose you'd want this to run on Python 2 as well?

ColdHeat:
Yeah absolutely. I already added `netaddr` to CTFd though.

PR #277, opened 2017-06-10
Fixing theme switching

ColdHeat:
Something in Flask 0.12 changed the behavior of Jinja templates. There
are a couple open issues about it:

https://github.com/pallets/flask/issues/1907

### DIFF
--- a/CTFd/models.py
+++ b/CTFd/models.py
@@ -37,7 +37,7 @@ class Pages(db.Model):
         self.html = html
 
     def __repr__(self):
-        return "<Pages {0} for challenge {1}>".format(self.tag, self.chal)
+        return "<Pages route {0}>".format(self.route)
 
 
 class Containers(db.Model):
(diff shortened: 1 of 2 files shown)


frohoff:
FYI, pallets/flask#1907 seems to have been fixed so it might be worth looking at re-enabling template caching

ColdHeat:
@frohoff does removing the fix that was merged in here work for you by chance?

frohoff:
It wasn't clear to me from the description in this PR what the exact problem was but I assumed it had something to do with stale jinja templates after a template change? I've never done anything with CTFd themes so wasn't sure how to test that, though presumably if there was still an issue one could probably programmatically flush the jinja cache when the theme was changed.

Random note: commenting out the cache disabling code seems to yield a 50%+ drop in response latency for me (250-300ms to 100-130ms) for the pages I was testing.

ColdHeat:
This was put in because switching themes would result in weird template errors. I think perhaps the simpler solution is to just clear the cache on configuration changes.

ColdHeat:
@frohoff https://github.com/CTFd/CTFd/pull/630 should be useful

Issue #284, opened 2017-06-17
Add vagrant support


royragsdale:
[Vagrant][] provides easily reproducible development environments by making it really easy to spin up a pre-configured virtual machine.  This could be a nice addition, particularly if the vm was configured with docker support, to make getting started and working with the container features even easier.

The most simple workflow could be as easy as:
```
vagrant up
```
Behind the scenes this could pull down a base operating system, install the dependencies, and start a development server. So that after a few minutes (first time downloading the vm and dependencies), a developer could just visit `http://localhost:4000` and have a CTFd instance running with their source code synced into the vm.

I've implemented this if it is of interest, PR to follow shortly.

[Vagrant]:https://www.vagrantup.com/

royragsdale:
I would be happy to make modifications to the PR as well.  Also I would be happy to provide a wiki page on how the vm is configured and some basic `vagrant` workflows.

PR #285, opened 2017-06-17
Add Vagrant with docker support

royragsdale:
Reference #284 

This pull request adds a Vagrant configuration (`Vagrantfile`) and all the necessary configuration to provision the machine to be immediately useful.

Within the vm, on first boot all dependencies are installed.  Python dependencies are installed using `virtualenvwrapper`.  `docker` is installed and configured.

Source code is synchronized into the vm in the `/vagrant` folder.

The ports for running in directly in development mode (4000) and with docker (8000) are exposed from the vm to the host machine so a developer can still just hit localhost.  Also the machine is on a host-only network at `http://10.9.8.7` to facilitate access to network/web based challenge development.

Only minor modifications where made to support this:
- Changed `docker-compose.yml` to mount mysql data as a named volume due to a permissions conflict when mysql is used in a vagrant synced folder
- Changed `serve.py` to bind to `0.0.0.0` rather than `loopback` so the port can be exposed outside the vm


royragsdale:
Tested using `debian stretch` as a host OS, `Vagrant 1.9.5`, and `VirtualBox  5.1.22 r115126`.

ColdHeat:
Hmm interesting... Do you need to modify the `docker-compose.yml` file or the `serve.py` file? 

I think it would probably be better to use gunicorn inside of the VM similar to what the Dockerfiles do.

royragsdale:
I made minor changes the `docker-compose.yml` file to change how the mysql data was mounted. I couldn't think of a usecase where having the db files on the local system would be more useful than in a named volume.  This change was necessary because when `docker-compose up` was run in the VM, the synced folder caused permissions issues.  Even when syncing the source in with `777` permissions the the db container had issues with ownership/permissions on the  `.data` folder 

For the the `serve.py` file I just changed to bind to all interfaces rather than just loopback. This allows the port to be forwarded to the host machine and eases access.

In general both were just an attempt to reuse existing scripts/configurations and matching the workflows described in the wiki rather than creating a whole different setup or adding more complexity.  However if it is a better fit, I'd be happy to create a different work flows such as the mentioned `gunicorn` method.

royragsdale:
Made the requested change to use `gunicorn` inside the VM and removed the modification to `serve.py` (back to the version on master).

ColdHeat:
Hey @royragsdale I'm sorry this fell to the wayside for a bit. 

I asked a friend since I don't actually use Vagrant and I got some feedback that things weren't working: 

```
vagrant up
vargant ssh
---
vagrant@contrib-jessie:~$ tmux ls
failed to connect to server: Connection refused
```

What's the proper way to use this Vagrant VM?

royragsdale:
Interesting, most likely that means the `gunicorn` server didn't start correctly (and thus the default tmux session died). Perhaps due to a failed download/dependency or something.  If they could share the log from `vagrant up` that would be helpful.

Alternative running `vagrant reload` might be a good first troubleshooting step to reboot the machine (and thus restart the gunicorn service).

A second trouble shooting step would be:
```
vagrant ssh
cd /vagrant
workon ctfd
python serve.py
```
This will give more information on if it was an issue provisioning the machine and installing dependencies, or it it's just an issue with launching the tmux session/gunicorn server.

What os are they on (I've only tested on linux)?

I'll post a doc describing more about the VM, but certainly want to make this more robust.

royragsdale:
Hey, @ColdHeat I pushed some documentation on Vagrant and the VM configuration on my fork's wiki:

https://github.com/royragsdale/CTFd/wiki/Vagrant

Hopefully this helps explain a little better what's going on. Once we figure out the bug, if this PR is accepted, perhaps this could be pulled into the main wiki as well.

royragsdale:
Alternatively if the current version is considered to fragile to merge, but the idea is still of interest, I can re-write the provisioning stuff using some configuration management like ansible. That way things are more robust and won't fail silently/in a big wall of text.

ColdHeat:
I confirmed this on my own that this Vagrantfile does not work...

It doesn't even come up on its own.

ColdHeat:
Seems like the issue for coming up is not updating in `prepare.sh` so will test a little more.

PR #295, opened 2017-06-21
Add Tags as a CSS class on challenge boxes

Hypersonic:
Previously, there was no way to know what tags were on a challenge on the client-side without making a separate query. This was inconvenient for theme authors who might want to add some styling based on Tags, without needing to insert or modify JS in the middle of the chalboard files.

This will, for instance, allow theme authors to give admins an easy way to customize challenges (imagine a "Hot" set of challenges, moderated by Tags, or challenges sorted on the client-side by tag).

I put the tag on the outermost element for each challenge, though it wouldn't be unreasonable to move it to one of the inner ones.

### DIFF
--- a/CTFd/themes/original/static/js/chalboard.js
+++ b/CTFd/themes/original/static/js/chalboard.js
@@ -231,8 +231,8 @@ function loadchals(refresh) {
             var chalheader = $("<h5>{0}</h5>".format(chalinfo.name));
             var chalscore = $("<span>{0}</span>".format(chalinfo.value));
             for (var j = 0; j < chalinfo.tags.length; j++) {
-                chalwrap.addClass(chalinfo.tags[j]);
-                console.log(chalinfo.tags[j]);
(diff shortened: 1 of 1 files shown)


ColdHeat:
This makes sense, the only catch is that CTFd allows tags to contain spaces. The question I guess is should CTFd prevent tags from containing spaces or should plugins/themes deal with the spaces.

Hypersonic:
Good point, didn't think of that. I think it would be most sane to ban spaces in tags -- is there any good reason to allow them?

Hypersonic:
Alternatively, the client could do something like `s/ /-` before adding the class, and perhaps prefix `tag-` for namespacing, which would probably be the least invasive change to make this happen.

ColdHeat:
I think, the best approach here is to replace the spaces with dashes or similar. I think putting spaces in Tags is probably a mistake but since this is restricted to the original theme itself, I don't see a big deal with leaving tags as they are now and just encoding some things when they're drawn.

PR #304, opened 2017-07-08
Sample deployment of CTFd to AWS using terraform - wip

gambtho:
This isn't quite finished, but wanted to get some feedback before I did much more with it.   It seems like a number of CTFs are using this awesome setup on AWS, but aren't quite sure how to get up and running quickly.   

The idea with this PR is to provide a nearly turnkey solution to having things up and running on AWS with no in-depth knowledge of AWS


ColdHeat:
Hey @gambtho thanks A LOT for this PR. I'm not very familiar with Terraform but I do think this will be very useful but I think it might be better suited in its own repo. 

Do you think it would be possible to structure this as its own repo and not exist inside of CTFd itself?

gambtho:
Sure, no problem.   I'll set it up that way

Issue #309, opened 2017-07-12
Render plugin's config.html with variables


farisv:
Currently in `CTFd.admin.admin_plugin_config`, `config.html` for each configurable plugins are rendered with `render_template_string(config)`. It will be nice if we can pass some variables from plugin code to Jinja when rendering config.html. Or, allow plugin interface to render certain own files for `/admin/plugins/<plugin>` route so it can use `render_template(file, var=value)` for rendering.

ColdHeat:
Hmm, I had a feeling `config.html` would evolve into this but can you give me some examples where you feel `config.html` isn't sufficient for your needs? 

Given that templates can call `get_config()` I think `config.html` works for a good amount of plugins but this could be useful.

farisv:
For example, a plugin that create new database and it need to display database content in config page.

ColdHeat:
That makes sense, but I feel like you can also load this data into `config.html` with an AJAX request and another route. 

My concern is that it becomes more difficult to tell which plugins are going to expose a configuration panel.

farisv:
Yes, this is just an alternative for developer.

In my pull request #310 , my approach is to extend `CTFd.admin.admin_plugin_config`. The behavior is still the same like the original implementation, but for any plugin that has `config.html` and `plugin_config_route` method, return `plugin.plugin_config_route(config, request)` instead of `render_template_string(config)`.

If the plugin has `config.html `and has no `plugin_config_route` method, just return `render_template_string(config)` like usual. If it doesn't have `config.html`, return 404 for GET request or set config for any key-value data except for `nonce` for POST request, just like the original implementation.

ColdHeat:
I think https://github.com/CTFd/CTFd/pull/427 closes this as plugins can now control `config.json` which allows devs to control the configuration links.

PR #310, opened 2017-07-12
Give plugins more control to render config page

farisv:
This is my suggestion to give plugins more control to render config page #309 

Developer can add `plugin_config_route()` method in plugin module so it can render config page with passed variables to Jinja, render a file other than config.html, or simply render a string for `/admin/plugins/<plugin>` route.

By default, request to `/admin/plugins/<plugin>` will call plugin's plugin_config_route(config_template, request) (with config is config.html template string and request is processed request within route) if it's exist. If not exist, the behavior is still the same like before.

Example:
```
def plugin_config_route(config_template, request):
    value = 0
    if request.method == 'POST':
        value = request.form.get('key', value)
    return render_template_string(config_template, key=value)
```

### DIFF
--- a/CTFd/admin/__init__.py
+++ b/CTFd/admin/__init__.py
@@ -1,4 +1,5 @@
 import hashlib
+import importlib
 import json
 import os
 import datetime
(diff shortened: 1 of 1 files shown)


farisv:
Actually, I'm thinking to also allowing plugins to process POST request data in `/admin/plugins/<plugin>` route. Currently, any POST request to `/admin/plugins/<plugin>` will call `utils.set_config(k, v)` even if plugin doesn't exist. What do you think?

ColdHeat:
As of https://github.com/CTFd/CTFd/pull/425 this PR is no longer necessary because plugins can now control where their configuration page is via `config.json`.

PR #314, opened 2017-07-15
Environment var for UPLOAD_FOLDER

Owlz:
It would be helpful to have the option to set the UPLOAD_FOLDER via environment variable. This change allows that.


ColdHeat:
Thanks!

ColdHeat:
Although looking at this in more depth, it's not very likely that this PR will work as expected because of how CTFd performs some additional path manipulation to get the final upload_folder. 

Some more changes will need to be done but I can handle those.

PR #323, opened 2017-07-25
Clear cache after set config in admin_plugin_config()

farisv:
The `/admin/plugins/<plugin>` route for POST method does not clear the cache after `utils.set_config(k, v)` so the results for `utils.get_config(k)` are not updated immediately.


ColdHeat:
Thanks for the PR!

PR #339, opened 2017-08-07
Don't try to log in when no username is given

Fxrh:
It might be that no login is required to send mails (e.g. if the
mail server filters by network). In this case, CTFd should not
try to log in with an empty user/password.


ColdHeat:
Interestingly you and #340 implemented something similar. But you submitted the PR first so you're closer to merging. 

But I'm not sure, is #340 technically the correct fix? Is there ever a situation that you'd provide no username but provide a password?

ColdHeat:
Closing in favor of #341

PR #340, opened 2017-08-08
Added check for SMTP AUTH support

SkeyeLlama:
An error condition exists where the configured smtp server does not support the AUTH extension, as may be the case with a local smtp server configured to provide mail relay for services on the ctf box.  

This modification is on line 445 and will verify whether the smtp server has reported AUTH support before attempting to log in. 

Note, that this change will result in skipping the login quietly when AUTH is not supported, regardless of whether a user/pass has been configured.

### DIFF
--- a/CTFd/utils.py
+++ b/CTFd/utils.py
@@ -442,7 +442,8 @@ def get_smtp(host, port, username=None, password=None, TLS=None, SSL=None):
     if TLS:
         smtp.starttls()
         smtp.ehlo()
-    smtp.login(username, password)
+    if smtp.has_extn('AUTH'):
+        smtp.login(username, password)
     return smtp
 
 


ColdHeat:
Interesting, you and #339 implemented something similar. I think for correctness, they should both be merged but it will cause a conflict that will need to be resolved by one of you. 

I'll likely merge #339 in first because it came in first, but can you clarify one thing for me:

* If there is no username specified is it still possible that the AUTH module is available?

SkeyeLlama:
339 is similar, but these achieve a different result.

340 prevents an error condition from occurring if SMTP AUTH is attempted (default behavior) but not supported, regardless of CTFd mail server config. Based on current exception information, it is not clear to the admin through CTFd whether email is failing due to credentials or lack of AUTH support.

339 changes the default behavior to prevent SMTP AUTH if a username is not defined. There may be an additional exception edge case unintentionally included in 339 if user error or unforeseen configuration requirements should actually require username to be blank. I don't know if that is even possible under RFC. 

I think this really requires a multipart solution that neither 340 nor 339 fully addresses to achieve the most complete/graceful result. 

1) A configuration box to enable/disable smtp auth explicitly (Current function is to login as NONE/NONE). This seems to address the complaint in 339.
2) Better exception handling around credential failure and detection of support for AUTH. Ideally the admin will be informed with distinct messages if 1) SMTP AUTH is enabled in CTFd mail server config but not supported by the server or 2) the configured credentials are wrong. This would more formally address the situation I am attempting to address in 340.

Fxrh:
The specific error I had in my installation and which I tried to fix in my pull request would be fixed by this PR, too.

I agree with @SkeyeLlama  that it might be a good idea to have a separate checkbox for enabling authentication above the username and password field (not only to remove possible errors, but also to make it more explicit for the admin how to disable authentication).

Slightly offtopic: It seems like the SSL checkbox is not used (at least the parameter given to ```get_smtp``` is not used). Can it be removed?

ColdHeat:
Sure, I can remove the checkbox. 

I like #341 the most given that it integrates more. I'm going to close #340 and #339 in favor of that.

PR #341, opened 2017-08-08
Fixing smtp combine fix 339 and 340 more complete

SkeyeLlama:
Here is a more elegant (complex) solution to 339 & 340. Still missing here is error handling with feedback for the exception cases. 

Added a toggle to the config page and updated all the obviously relevant areas to reflect the new config option.


ColdHeat:
Great! Thanks for the PR. Only things I would add are to hide the username and password input boxes unless the toggle is checked. 

But I'll gladly add that in after merging. 

I'll review this in about 30 minutes hopefully and then merge.

ColdHeat:
@SkeyeLlama can you make a Github issue about the error handling and feedback?

PR #347, opened 2017-08-11
Allow importing of longer-running docker builds

wyattearp:
The current docker builds can time out, even for simple things. This pull request allows the user to do their own "docker build" in the background and then import the image into the database so that it can be launched as needed. By default the image is not launched.

### DIFF
--- a/CTFd/utils.py
+++ b/CTFd/utils.py
@@ -577,10 +577,9 @@ def is_port_free(port):
 
 def import_image(name):
     try:
-        #TODO: probably super bad command line injection (only as root, so not super bad)
         info = json.loads(subprocess.check_output(['docker', 'inspect', '--type=image', name]))
         print(info)
-        container = Containers(name,"<none>")
+        container = Containers(name, "<none>")
         db.session.add(container)
         db.session.commit()
         db.session.close()
--- a/CTFd/utils.py
+++ b/CTFd/utils.py
@@ -537,6 +537,20 @@ def is_port_free(port):
     return True
 
 
+def import_image(name):
+    try:
+        #TODO: probably super bad command line injection (only as root, so not super bad)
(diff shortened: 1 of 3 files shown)


CTFd/utils.py
```diff
@@ -575,6 +575,19 @@ def is_port_free(port):
     return True
 
 
+def import_image(name):
+    try:
+        info = json.loads(subprocess.check_output(['docker', 'inspect', '--type=image', name]))
+        print(info)
+        container = Containers(name, "<none>")
```
ColdHeat: You should be able to simply set the `buildfile` to None here. 

Also can you explicitly specify the keywords here (i.e. `name=name, buildfile=None`)

ColdHeat:
I'm not particularly against merging this but you should know that the Containers functionality is going to be taken out of CTFd and put into a plugin. 

May I ask how you have been using the Container functionality? It's not very commonly used afaik.

wyattearp:
Yea, I figured it wasn't a very common thing that people used. The way I have been using it is to spin up the challenges while the system is active, and turn them off when it's not (so if there's no solves). I've taken the previous approach to the CTF problems where we just move to a stdio server to keep from having accidental bugs in the networking code (unless a problem specifically requires it). Here's the docker build I was using in case you see something of value you'd like to move into the plugin:

```docker
FROM python:3
RUN apt-get update && apt-get install -y --no-install-recommends \
        ucspi-tcp
RUN rm -rf /var/lib/apt/lists/*
WORKDIR /opt/chroot/test_problem1
ADD . /opt/chroot/test_problem1
EXPOSE 12345
CMD ["tcpserver", "-RHl0", "0.0.0.0", "12345", "python3", "test_problem1.py"]
```

PR #348, opened 2017-08-12
Refactor Containers into a plugin

ColdHeat:
This mostly just removes existing code to get it into a plugin. 

The new plugin is https://github.com/CTFd/CTFd-Docker. It's just a port of all the code over. Assuming this isn't broken there should be nothing to see here. 

The plugin itself should be updated to not use `subprocess` though. Perhaps explore the idea of managing Docker clusters from within CTFd as well.


ColdHeat:
I'm starting to wonder more about this. 

If Containers are refactored into a plugin, CTFd loses the ability to manage service based challenges by default. The benefit here is that a CTFd export might contain all the information needed to entirely recreate a CTF if they were to make full use of CTFd..

Perhaps this needs to be redesigned instead of making it a plugin? Or perhaps make it a plugin and then design a new system considering the Container system really isn't so good at what it was intended to do.

ColdHeat:
Decided that the current implementation of CTFd containers is not sustainable long term. It should be refactored into Services and made a first class CTFd feature.

PR #349, opened 2017-08-12
Add sorting of labels in the plotly scoreboard chart

kara71:
## Description
Currently, labels in the scoreboard chart are sorted in lexicographic order.
This PR adds sorting in a way that labels appear in the same order as the top 10 (1st team on top, 10th at the bottom).

## Example:
The current scoreboard is:
scoreboard

The old way of displaying labels is:
old

The new way of displaying labels is:
new


ColdHeat:
Thanks for the PR, it's a much needed one!

PR #357, opened 2017-08-22
Configure mail_server and mail_port from config for easier deployments

tophertimzen:
Add MAIL_SERVER and MAIL_PORT options to config.py to allow for scripted deployments of CTFd.

### DIFF
--- a/CTFd/config.py
+++ b/CTFd/config.py
@@ -73,7 +73,7 @@ class Config(object):
     MAILFROM_ADDR = "noreply@ctfd.io"
 
     '''
-    MAIL_SERVER is the name of the mail server emails are sent from if not overriden in the configuration panel. 
+    MAIL_SERVER is the name of the mail server emails are sent from if not overriden in the configuration panel.
     '''
     MAIL_SERVER = "localhost"
 
@@ -81,7 +81,6 @@ class Config(object):
     MAIL_PORT is the port of the mail server emails are sent from if not overridden in the configuration panel.
     '''
     MAIL_PORT = ""
-
 
     '''
     UPLOAD_FOLDER is the location where files are uploaded.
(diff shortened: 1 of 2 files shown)


tophertimzen:
the 'pep8' test should be added to the Development wiki as it was not included in `nosetests`. This resulted in an additional commit to satisfy the requirements when I could have located the issue locally.

ColdHeat:
Thanks for the PR, I'll review it but it looks pretty straight forward. I added a line to the Wiki.

ColdHeat:
So it looks like `get_config` will already do the behavior that you want, it is just not case insensitive so if you set MAIL_SERVER it won't find it. I think that's the real issue not so much that the config values aren't there.

tophertimzen:
Ah, I was going off of the naming convention of the config file such as `MAILFROM_ADDR` for `app.config.get` throughout CTFd for the mail server settings. It did not seem there are many configurations that `app.config.get` actually grabs, so I figured that it was more easily understandable by adding all configuration options to `config.py`.

ColdHeat:
Yeah that makes sense. 

I definitely approve of the idea you're going for so I will probably create a new PR to make load config values better and then revisit this so we can have all the configuration options in config.py.

tophertimzen:
That sounds great to me, @ColdHeat!

ColdHeat:
Hey @tophertimzen, I just merged https://github.com/CTFd/CTFd/pull/381 this in giving get_config the ability to see those values in app.config more easily. Do you mind updating your PR to reflect that?

ColdHeat:
I ended up having to revert the get_config change because it changed more behavior than I had expected. 

In the meantime you can set the lower case values inside of `config.py` so that CTFd will prioritize those over values set in the admin panel.

tophertimzen:
I haven't had time to change my PR, but that is good to know. Thanks!  Keep me posted and I'll change as needed to merge :)

ColdHeat:
I just created https://github.com/CTFd/CTFd/issues/640 which is related to this.

ColdHeat:
I believe this behavior is now supported as of 2.0.0.

Issue #367, opened 2017-08-26
500 after clearing Teams table


nia2202:
Hi,

I am getting Internal Server Error. Below is my stack trace:

gunicorn[1537]:     conn = self.session.connection(**kw)
gunicorn[1537]:   File "/home/ctf/ctfd/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 998, in connection
gunicorn[1537]:     execution_options=execution_options)
gunicorn[1537]:   File "/home/ctf/ctfd/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 1003, in _connection_for_bind
gunicorn[1537]:     engine, execution_options)
gunicorn[1537]:   File "/home/ctf/ctfd/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 382, in _connection_for_bind
gunicorn[1537]:     self._assert_active()
gunicorn[1537]:   File "/home/ctf/ctfd/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 276, in _assert_active
gunicorn[1537]:     % self._rollback_exception
gunicorn[1537]: InvalidRequestError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (pymysql.err.IntegrityError) (1452, u'Cannot add or update a child row: a foreign key constraint fails (`ctfdb`.`tracking`, CONSTRAINT `tracking_ibfk_1` FOREIGN KEY (`team`) REFERENCES `teams` (`id`))') [SQL: u'INSERT INTO tracking (ip, team, date) VALUES (%(ip)s, %(team)s,%(date)s)'] [parameters: {'date': datetime.datetime(2017, 8, 26, 16, 15, 18, 115186), 'ip':'127.0.0.1', 'team': 5}]

ColdHeat:
Clear the cookies in your browser.

nia2202:
It worked. Thank you

PR #368, opened 2017-08-26
Handle when database is cleared

ColdHeat:
I'm not even really sure this is a bug to be honest. If you clear out the database, you should also clear your cookies. 

Fundamentally the flaw in #367 is that when you clear the database, the cookie in your browser will still point to a Team id that does not exist. Then attempting to track on a user that doesn't exist will cause SQLAlchemy to complain. 

I think overall the code doesn't get any worse so it should be fine but the tracker code is called very often and I don't want to add any complexity or lag there.


ColdHeat:
Merged because this could happen if a team is deleted from the database say if the team is deleted.

PR #433, opened 2017-10-29
fix uploads folder detection

uqam-fob:
When started from outside the tree, ctfd fails to properly detect the
`uploads/` directory path and, subsequently, fails to serve challenge files.

This commit adds `abspath()` to properly detect the installation directory.

### DIFF
--- a/CTFd/config.py
+++ b/CTFd/config.py
@@ -77,7 +77,7 @@ class Config(object):
     The default destination is the CTFd/uploads folder. If you need Amazon S3 files
     you can use the CTFd S3 plugin: https://github.com/ColdHeat/CTFd-S3-plugin
     '''
-    UPLOAD_FOLDER = os.environ.get('UPLOAD_FOLDER') or os.path.join(os.path.dirname(__file__), 'uploads')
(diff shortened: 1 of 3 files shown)


ColdHeat:
Thanks for the PR @uqam-fob, do you think you could write a test for this as well? 

Perhaps simply checking that the `uploads` folder is pointed to by the `UPLOAD_FOLDER` variable?

uqam-fob:
I see three (3) aspects worth writing tests for:

- check that `UPLOAD_FOLDER` is properly used;
- check that an out-of-tree run properly uses the `uploads` directory;
- check that an in-tree run properly uses the `uploads` directory.

Would that be okay?

ColdHeat:
That would be perfect. I am just not sure how you are going to accomplish running "out of tree".

uqam-fob:
Added 2 tests.

Regarding in/out-of tree runs, I think it would be cleaner to use a shell script to start instances rather than running them from unit tests. To be honest, though, I am pretty certain that using `abspath` eliminates the issue and, therefore, believe that testing with these configurations is unnecessary.

ColdHeat:
Hey @uqam-fob, thanks for taking the time to write those tests. I think they kind of look a little odd so I reverted them. But I do agree that `abspath` should be all that's necessary so I'm just going to go ahead and merge and come up with a proper test later.

uqam-fob:
No problem. Thanks for your work.

Issue #437, opened 2017-11-07
Hints can be unlocked after the end of a CTF for no cost


iAppropriate:
There is a case where if the CTF is frozen/temporarily ended a user can not only access the hints but also "buy" hints for free (regardless of the cost).

Steps to reproduce:
1. Have a user login and view the challenges page
2. Use an administrative account to end/freeze the CTF (in another window)
3. Have the user "buy" hints and log out of the CTFd
4. Have the admin resume the CTF
5. When the user logs back in, the hint will "relock" but points will not be reduced

My suspicion is hints are already being served by the web server but since the scoreboard is frozen the team points are not deducted.

I came up with a fix which seems to do the trick. By replacing **Line 21** in **challenges.py**, this should prevent hints from being leaked:

Original:
`if not utils.ctf_started():`

Suggested:
`if not utils.ctftime() or is_scoreboard_frozen():`

I have made a pull request with the fix.

ColdHeat:
In my testing, in a frozen CTF I unlocked a hint and was able to unlock that hint and lost points for it. 

In an ended CTF I unlocked a hint and was able to unlock that hint but did not lose points for it. 

Thus it would appear that the issue is that hints can be unlocked after the CTF **ends** for free. 

Your PR likely will not fix this issue because the affected code is actually [here](https://github.com/CTFd/CTFd/blob/master/CTFd/challenges.py#L55). This code should not allow hints to be seen after the end of a CTF unless "Viewing challenges after CTF end" is enabled. 

if you update your PR please include a test for this behavior if possible. 

Thanks again for making the issue and the PR!

PR #438, opened 2017-11-08
Prevent Hints from Leaking

iAppropriate:
Patch for Case when CTFd is Frozen or Temporarily Ended: CTFd#437


ColdHeat:
Hey @iAppropriate, thanks for the PR and the issue. It looks like your PR doesn't pass the existing tests. You should access the `is_scoreboard_frozen` function as `utils.is_scoreboard_frozen()`. 

Also I'm not sure I agree with this. I think Hints should still be unlocked when the scoreboard is frozen but the affect to points should not be reflected in the scoreboard.

ColdHeat:
I believe #439 should supercede this and fix #437 so I'm closing this.

PR #463, opened 2017-11-15
Added trailing '/' in path of CTFd logo on main page

reenberg:
When you are viewing a preview of the index page in the admin panel, the CTFd
logo is not shown correctly as the path for the logo i relative and thus becomes
'/admin/pages/themes/original/static/img/logo.png' instead of
'/themes/original/static/img/logo.png'.

### DIFF
--- a/CTFd/views.py
+++ b/CTFd/views.py
@@ -38,7 +38,7 @@ def setup():
 
             # Index page
             page = Pages('index', """<div class="container main-container">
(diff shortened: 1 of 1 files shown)


ColdHeat:
This is so that if a user chooses to deploy under a directory (e.g `/ctf`) the CTFd logo still shows up. 

The flaw here is really the preview, not the content of the page itself. The preview itself should probably be an entirely separate window.

reenberg:
Point taken.

PR #464, opened 2017-11-15
Feature/challenge plugin event

reenberg:
So it ended up being a few commits, but I have tried to break the changes up into logical steps, so that something may be taken out, and most of the other stuff should still work.  Each commit should have a good long description of what is done.

I know that the whole introduction of `NameSpace` in e01c8c6bac2d7a0bed5d68049cda1831150b69c0 make that commit look huge, but actually it is quite simple.  It just add `ns.` to almost all lines :)
Actually I ended up no needing it to implement CTFd/DynamicValueChallenge, once I molded the `CTFdStandardChallenge` to be generic enough.  
Thus we could roll without this, but that would prohibit anyone from creating a plugin (not nessesarily a challenge plugin) that directly changes the actual variable in the local scope, as opposed to anything that it references internally.  In other words, with this we offer great possibilities to anyone creating a plugin and wanting to add side effects to other challenge plugins.

Btw, the change in 55a4288f684face00df288791c9daf845a9b19a8 really should be done to all relevant models, if I read the documentation correctly?  It is currently a No-Op on MySQL, but if someone would use another database, it would actually make it better.

Please comment.


reenberg:
Hmm the nosetests runs flawlessly on my setup using python 2.7.  I admit, I totally forgot to run it using python 3 though.

reenberg:
I have written/ported the following plugins that use this feature, which all have been tested this week on a live event.  

- https://github.com/reenberg/CTFd_FirstBlood
- https://github.com/reenberg/DynamicValueChallenge/tree/feature/event-based
    - [Comparison with master](https://github.com/reenberg/DynamicValueChallenge/compare/master...reenberg:feature/event-based?diff=split&name=feature%2Fevent-based#diff-743b1adbf37db12e16e01ef62a6f7d72)
    - Main changes are in "*\_\_init\_\_.py*" which really benefits from the use of class methods in `CTFdStandardChallenge` and the fact that most things are now handled in a more generic way.  Allowing for less code duplication.

zevlag:
@ColdHeat Are you opposed to implementing the EventEmitter as defined in the PR?  I'd be willing to port this code to the current master if it's something that would be accepted.

ColdHeat:
We had discussed this some time ago and it was decided that the current OOP/inheritance approach was deemed to be simpler than having events and I still agree. 

Events themselves I think are another layer of the plugin system that I don't think I want to take on maintaining just yet.

PR #478, opened 2017-11-23
Rename Keys.key_type to Keys.type (#459)

### LINKED ISSUE #459: Rename Keys.key_type to Keys.type
Seems cleaner to do this instead of saying key again.

DreamN:
Closes #459 
@ColdHeat


ColdHeat:
Mmmm not quite... You can't just rename all instances of key_type because if there are existing CTFd instances that update they will be running outdated schemas. 

Instead you need to rename all instances only within the `CTFd` & `tests` folders and then create a migration for [Flask-Migrate](https://flask-migrate.readthedocs.io/en/latest/)

DreamN:
@ColdHeat Fixed :D

DreamN:
Fixed @ColdHeat

ColdHeat:
Yes I can see that. Please be patient. This issue is slated for v1.1 so it will get merged into that when the time is right.

DreamN:
Awesome, Thank you 😄 
That's my very first pull request in github. 🍻

PR #485, opened 2017-11-26
sort categories alphabetically

maltek:
Right now, categories are listed on the page in database order of their challenges, so basically random. With this commit, they are sorted alphabetically.

What I really want, is to sort the categories in the same order as the lessons for a workshop I'm running, but this allows a close enough approximation for my purposes (by prefixing the category names with numbers).


ColdHeat:
An official paid theme implements a natural sort on challenge categories allowing you more kinds of control regarding how to display challenges: https://ctfd.io/store/

image

PR #517, opened 2017-12-10
Add APPLICATION_ROOT in config.py

bianyifan:
If someone want to deploy CTFd in a subdirectory (by setting reverse-proxy) with Docker, he will need this way to change that without rebuilding the image.

### DIFF
--- a/CTFd/config.py
+++ b/CTFd/config.py
@@ -63,6 +63,11 @@ class Config(object):
     PERMANENT_SESSION_LIFETIME = 604800  # 7 days in seconds
 
     '''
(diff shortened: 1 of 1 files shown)


bianyifan:
Still have problems

ColdHeat:
This is very deceptive because by setting this value you do not actually deploy anything on the subdirectory. 

If you want to deploy on a subdirectory you should read the [Deployment page in the wiki](https://github.com/CTFd/CTFd/wiki/Deployment).

Also bear in mind that subdirectory deployment is not very well supported.

PR #520, opened 2017-12-11
To allow previous version imports of keys to work - fill in key `type` from `key_type`.

dbaxa:
Signed-off-by: David Black <dblack@atlassian.com>

### DIFF
--- a/CTFd/utils.py
+++ b/CTFd/utils.py
@@ -883,6 +883,11 @@ def import_ctf(backup, segments=None, erase=False):
                 saved = json.loads(data)
                 for entry in saved['results']:
                     entry_id = entry.pop('id', None)
+                    if item == 'keys':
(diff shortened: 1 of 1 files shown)


ColdHeat:
codecov is a recent addition, so at the moment you don't need to worry about it. 

I'm busy working on some other code but I will get to your PRs in a bit.

ColdHeat:
What bothers me is that these kind of changes will become impossible to maintain as CTFd gets more migrations. Ultimately I think there needs to be some kind of way to apply migrations onto an export. 

Because some exports will be missing alembic_versions I think going forward another script/project should come out with the ability to run migrations on export files and generate new ones. 

However, I think going from 1.0.5 to 1.1.0 is an important enough change that this should be supported.

PR #557, opened 2018-01-20
Move Fontawesome script load to after body

### LINKED ISSUE #556: Safari - Challenge links on "Challenge" page  do not work
**Environment**:

  - CTFd Version/Commit: 
  - Operating System: Mac OS X - Version 10.12.6
  - Web Browser and Version:  Safari Version 11.0.2

**What happened?**
On Safari - the modals for some challenge links do not work
On Chrome - All modals for challenge links work.

Functional parity between Chrome and Safari

As an admin - 
  Create a challenge - name it X
  Access the "Challenge" page
  on the X challenge, click the icon representing "preview X"
  ---- no modal comes up
  on the X challenge, click the icon representing "edit X"
  ---- modal comes up

No

ColdHeat:
Closes #556 

For some reason Safari specifically is having trouble with the document.ready call. Either they're not being called under certain situations or they're being called out of order or something to that extent. 

This moves the font awesome script to after the body so that we don't "iconify" before we try to add handlers.


ColdHeat:
Actually this should be worked on alongside the Bootstrap v4 upgrade.

Issue #468, opened 2017-11-18
Gunicorn [CRITICAL] WORKER TIMEOUT 


FreakyFreddie:
**Environment**:

  - CTFd Version/Commit: Latest as of 11/11/2017
  - Operating System: Ubuntu 16.04 LTS, deploying platform in Docker containers, using docker-compose up
  - Web Browser and Version: Firefox: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:56.0) Gecko/20100101 Firefox/56.0 and Chrome: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36

**What happened?**
Seeing a lot of Gunicorn [CRITICAL] WORKER TIMEOUT entries in the error.log file. I do not fully understand what this means. I think it happened around the time the application stopped responding. I had to reboot the containers to fix the issue.

Any thoughts?

**Any associated stack traces or error logs**
Error.log
```
[2017-11-13 14:05:10 +0000] [216] [CRITICAL] WORKER TIMEOUT (pid:252)
[2017-11-13 14:05:10 +0000] [252] [INFO] Worker exiting (pid: 252)
[2017-11-13 14:05:10 +0000] [265] [INFO] Booting worker with pid: 265
[2017-11-13 14:05:41 +0000] [216] [CRITICAL] WORKER TIMEOUT (pid:265)
[2017-11-13 14:05:41 +0000] [265] [INFO] Worker exiting (pid: 265)
[2017-11-13 14:05:41 +0000] [278] [INFO] Booting worker with pid: 278
[2017-11-13 14:06:12 +0000] [216] [CRITICAL] WORKER TIMEOUT (pid:278)
[2017-11-13 14:06:12 +0000] [278] [INFO] Worker exiting (pid: 278)
[2017-11-13 14:06:12 +0000] [291] [INFO] Booting worker with pid: 291
[2017-11-13 14:06:43 +0000] [216] [CRITICAL] WORKER TIMEOUT (pid:291)
[2017-11-13 14:06:43 +0000] [291] [INFO] Worker exiting (pid: 291)
[2017-11-13 14:06:43 +0000] [304] [INFO] Booting worker with pid: 304
[2017-11-13 14:07:14 +0000] [216] [CRITICAL] WORKER TIMEOUT (pid:304)
[...]

PR #560, opened 2018-01-28
Improve the flexibility and ease-of-use for docker-compose deployment

nategraf:
`docker-compose` has the potential for being the easiest way to deploy both for development and production.
Here are some changes I'd suggest to the development side of that equation

1. I added `gevent` as the default worker, as because A) it is more performant B) I was experiencing some nasty timeout bugs with `sync` and I imagine I'm not the only one
2. Make an easy way to deploy with SSL by writing a self-signed cert to the ssl folder
3. Mount code read-only and enable reload to allow changes to be reflected in the server without rebuild or restart
4. Put `mysql` credentials in a separate file so most users don't need to look at docker-compose.yml, ensure that the usernames and password are the same on server and db, and to prevent edits to docker-compose.yml in the future from causing merge conflicts if the user does edit docker-compose.yml and then pulls from master
5.  Add plugin requirements download via pip if they included a requirements.txt file
6. Move the db to a network that is not connected to the internet

Additionally this PR includes a `docker-compose-production.yml` file which adds an NGINX server to proxy connections to Gunicorn. This configuration should be capable of serving a CTF with 1000+ registered users (it is similar to how my team deployed TAMUctf with ~800 registered users and the medium AWS node we ran on hardly broke a sweat)

If these changes are merged I'll go ahead and update the wiki


ColdHeat:
+1 on the changes to the Dockerfile
+1 on the changes to generating secret keys
+1 on customizable log directories

-1 on having a mysql_secrets file. I think this is unnecessary. I think it makes more sense in the docker-compose file. It doesn't change very often from my perspective and a merge isn't the worst thing ever. 
-1 on managing SSL for users or promoting self signed. It would be much better to simply point to LetsEncrypt.

Neutral on using `gunicorn_config.py`. I think it adds more files than is really needed for very little gain if there's no SSL management. 

Perhaps we should be exploring a separate repo for deploying CTFd?

nategraf:
On the mysql_secrets file, I see your point. I'll merge those env variable back in to the compose file

SSL is a topic new admins will be unsure on, and would be helpful to provide docs, and make sure there can use it without needing to add extra stuff beyond setting the right configs or environment variables. both. I definitely do not want to encourage people to use self-signed certs beyond being a way to test deployment with SSL for themselves. I'll remove the suggestion that they use a self-signed cert, and make it so the there is no default location, but have it specified by environment variable which Gunicorn and Nginx will look for. This environment variable can be mentioned in the docs

There is other value the gunicorn.py file adds:
* It allows you to set the log locations to match where the env variable directs it
* Allows you to set reload behavior based on an env variable
* Sets the number of workers based on cores rather than a static number
* Assembles the DB access string (although this is not as important)
* Gives the user a base to build on

The last bit in the gunicorn file with secret key management to allow for using the system when it is mounted read-only is something that should probably be rolled into config.py, and I'll flesh out what that would look like.

ColdHeat:
Okay let's get this merged in. 

Please move the gunicorn config, docker-compose-production, and nginx.conf into a separate PR. Those changes belong to a different issue: https://github.com/CTFd/CTFd/issues/565. 

This is fine to add to the `docker-compose.yml` though: 
```
+      - UPLOAD_FOLDER=/var/uploads
+      - LOG_FOLDER=/var/log/CTFd
```

I like the rest of your changes.

ColdHeat:
In terms of gunicorn, lets just add `--worker-class gevent` to hopefully fix https://github.com/CTFd/CTFd/issues/468. 

I'll create a new repository with the deployment related files and you can retarget your deployment files to it.

nategraf:
@ColdHeat I removed the extra config files. Let me know if it looks good.

I set gunicorn to use 4 workers by default as that should work on most any system for small moderate sized CTFs. I had experienced latency increases at around 50 users with 1 worker

ColdHeat:
Only thought is that as I mentioned in #568, if you have multiple workers, you start to need a shared cache between them because it appears to occasionally result in errors.

Perhaps docker-compose should also deploy redis?

nategraf:
Ah, ok. That's definitely something to include in the "production" deployment, but I'll reduce this version to 1 worker

docker-compose.yml
```diff
@@ -7,20 +7,34 @@ services:
...
     environment:
+      - UPLOAD_FOLDER=/var/uploads
+      - LOG_FOLDER=/var/log/CTFd
       - DATABASE_URL=mysql+pymysql://root:ctfd@db/ctfd
     volumes:
-      - .data/CTFd/logs:/opt/CTFd/CTFd/logs
-      - .data/CTFd/uploads:/opt/CTFd/CTFd/uploads
+      - .data/CTFd/logs:/var/log/CTFd
+      - .data/CTFd/uploads:/var/uploads
+      - .:/opt/CTFd:ro
```
ColdHeat: How does this affect the Dockerfile COPY'ing into the image? Perhaps this should be removed?

nategraf: It will overshadow the copied in files. Doesn't effect other deployment strategies (e.g. `docker run ctfd`) but it makes it pick up changes without rebuild when deployed like this

ColdHeat:
I took a little time to benchmark gevent and it does indeed seem faster in this completely unscientific mini one worker scenario. 

```
./wrk -t12 -c400 -d30s http://192.168.1.21:8000/scoreboard
Running 30s test @ http://192.168.1.21:8000/scoreboard
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   631.60ms  685.34ms   1.98s    74.14%
    Req/Sec     6.25      6.49    60.00     93.43%
  393 requests in 30.02s, 4.19MB read
  Socket errors: connect 0, read 0, write 0, timeout 219
Requests/sec:     13.09
Transfer/sec:    143.07KB
```

```
./wrk -t12 -c400 -d30s http://192.168.1.21:8000/scoreboard
Running 30s test @ http://192.168.1.21:8000/scoreboard
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.16s   509.08ms   1.96s    64.71%
    Req/Sec     8.33      4.02    20.00     76.22%
  297 requests in 30.09s, 3.17MB read
  Socket errors: connect 0, read 74, write 0, timeout 280
Requests/sec:      9.87
Transfer/sec:    107.82KB
```

Seems like a good call @nategraf, huge thanks for the PR!

PR #580, opened 2018-02-28
Add unlisted dependencies, remove dependency on sudo

batterystaples:
This commit adds the dependency on libssl-dev and cffi 1.11.5, and removes the implied dependency on sudo. Discussed on Slack:
https://ctfdio.slack.com/archives/C2ZJ9SULF/p1519779490000351


ColdHeat:
This is too old at this point and I think these changes are already in CTFd in some way or another. 

Thanks for making the PR though.

PR #589, opened 2018-03-16
Prevent scrollbar showing on pages that are smaller than the screen size

balthasar-m:
With the footer having distance 0 from the bottom, a scrollbar is shown on each page, even when it's not completely filled (Tested on Firefox/Ubuntu and Chrome/Windows10). Setting the distance to 1px solves this problem.

### DIFF
--- a/CTFd/themes/core/static/css/sticky-footer.css
+++ b/CTFd/themes/core/static/css/sticky-footer.css
@@ -11,7 +11,7 @@ body {
 
 .footer {
     position: absolute;
-    bottom: 0;
+    bottom: 1px;
     width: 100%;
     height: 60px; /* Set the fixed height of the footer here */
     line-height: 60px; /* Vertically center the text there */


CTFd/themes/core/static/css/sticky-footer.css
```diff
@@ -11,7 +11,7 @@ body {
 
 .footer {
     position: absolute;
-    bottom: 0;
+    bottom: 1px;
```
gehaxelt: Needs a comment!

ColdHeat: Yeah I agree. @balthasar-m want to add a simple comment saying that this prevents scrollbars from showing on pages that don't use the full page height?

ColdHeat: @balthasar-m I also think that this change belongs in the admin copy of this file?

balthasar-m: @ColdHeat Both done.

ColdHeat:
Interestingly it seems like this adds a horizontal scrollbar on IE11 😆

PR #590, opened 2018-03-19
Don't display team rank in workshop mode

atti1a:


ColdHeat:
Hey @atti1a thanks for this PR. 

One catch though is I don't think this test is right. I believe that you should check for the full HTML like "1st <small>place</small>". 

Can you double check that?

PR #630, opened 2018-05-17
Only clear jinja_env cache on config changes

ColdHeat:
This simplifies jinja cache clearing so that we only clear when we update config (i.e. switch themes). This should increase page load speeds because templates will now be cached.

### DIFF
--- a/CTFd/admin/__init__.py
+++ b/CTFd/admin/__init__.py
@@ -177,6 +177,7 @@ def admin_config():
         db.session.close()
         with app.app_context():
             cache.clear()
+            app.jinja_env.cache = {}
         return redirect(url_for('admin.admin_config'))
 
     # Clear the cache so that we don't get stale values
--- a/CTFd/utils/__init__.py
+++ b/CTFd/utils/__init__.py
@@ -204,10 +204,6 @@ def init_utils(app):
             if session['nonce'] != request.form.get('nonce'):
                 abort(403)
 
-    @app.before_request
-    def disable_jinja_cache():
-        app.jinja_env.cache = {}
-
 
 @cache.memoize()
 def ctf_name():


frohoff:
Wondering if that will work across multiple workers or if it will only clear it for the current worker.

ColdHeat:
@frohoff good thought. That's probably why I cleared on every request...

frohoff:
Yeah, seems tricky. Some possible options, though many are probably not ideal:
* Operator must manually restart workers when theme is changed
* Automatically trigger a worker restart when theme is changed
* Find a way to namespace cache keys by the theme name so cache lookups after a theme change will not hit cache entries from the old theme
* Use a shared cache between workers (redis, etc)

ColdHeat:
I think the best solution is to use the Flask-Caching cache. But it's not immediately clear how to do that.

ColdHeat:
Probably create a new [bytecode_cache](http://jinja.pocoo.org/docs/2.10/api/#bytecode-cache) using Flask-Caching?

frohoff:
That's what I was thinking. Looks fairly straightforward from some of the examples:

https://gist.github.com/rtt/5029885

Presume it would just take a `flask_caching.Cache` instance in the ctor and then use it for get/set operations.

frohoff:
Looks like it will probably at least read the template from disk and checksum the contents before it can get a cache hit on the bytecode cache, which may or may not help much depending on the relative performance cost of file read, compilation, and redis read.

`jinja2/environment.py`:

```python
    @internalcode
    def _load_template(self, name, globals):
        if self.loader is None:
            raise TypeError('no loader for this environment specified')
        cache_key = (weakref.ref(self.loader), name)
        if self.cache is not None:
            template = self.cache.get(cache_key)
            if template is not None and (not self.auto_reload or
                                         template.is_up_to_date):
                return template
        template = self.loader.load(self, name, globals)
        if self.cache is not None:
            self.cache[cache_key] = template
        return template
```

`jinja2/loaders.py`:

```python
    @internalcode
    def load(self, environment, name, globals=None):
        """Loads a template.  This method looks up the template in the cache
        or loads one by calling :meth:`get_source`.  Subclasses should not
        override this method as loaders working on collections of other
        loaders (such as :class:`PrefixLoader` or :class:`ChoiceLoader`)
        will not call this method but `get_source` directly.
        """
        code = None
        if globals is None:
            globals = {}

        # first we try to get the source for this template together
        # with the filename and the uptodate function.
        source, filename, uptodate = self.get_source(environment, name)

        # try to load the code from the bytecode cache if there is a
        # bytecode cache configured.
        bcc = environment.bytecode_cache
        if bcc is not None:
            bucket = bcc.get_bucket(environment, name, filename, source)
            code = bucket.code

        # if we don't have code so far (not cached, no longer up to
        # date) etc. we compile the template
        if code is None:
            code = environment.compile(source, name, filename)

        # if the bytecode cache is available and the bucket doesn't
        # have a code so far, we give the bucket the new code and put
        # it back to the bytecode cache.
        if bcc is not None and bucket.code is None:
            bucket.code = code
            bcc.set_bucket(bucket)

        return environment.template_class.from_code(environment, code,
                                                    globals, uptodate)
```

`jinja2/bccache.py`:

```python
    def get_bucket(self, environment, name, filename, source):
        """Return a cache bucket for the given template.  All arguments are
        mandatory but filename may be `None`.
        """
        key = self.get_cache_key(name, filename)
        checksum = self.get_source_checksum(source)
        bucket = Bucket(environment, key, checksum)
        self.load_bytecode(bucket)
        return bucket
```

ColdHeat:
This PR has no chance of working. Looking at implementing the BytecodeCache based off of Flask-Caching runs into the issue that it's not easy to invalidate cached templates because the SimpleCache doesn't seem to have a way to search through keys.

PR #651, opened 2018-07-14
Display import_ctf Exceptions via repr

Zanadar:
- Wraps exceptions on `/admin/import` returned to users in a `repr()`, making debugging easier.

### DIFF
--- a/CTFd/admin/__init__.py
+++ b/CTFd/admin/__init__.py
@@ -70,7 +70,7 @@ def admin_import_ctf():
             import_ctf(backup)
     except Exception as e:
         print(e)
-        errors.append(type(e).__name__)
+        errors.append(repr(e))
 
     if errors:
         return errors[0], 500


ColdHeat:
Sorry this actually slipped off my radar. I believe the choice to only show the name is intentional so that we do not leak any information about the internals of the CTFd process. 

Does this happen to do that? Can you provide a difference between the two?

Zanadar:
Nothing should be leaked, except a vaguely more useful error. In any case, this would happen only during the import process (I used it to import the config from Juice Shop and it helped me debug the fact I was trying to import too many options). 
I don't have access to my laptop at the moment to show what it actually looks like but here's a representative example:

```
>>> try:
...     print 10 / 0
... except Exception as e:
...     print repr(e)
...
ZeroDivisionError('integer division or modulo by zero',)
```

ColdHeat:
I think this change makes sense so I will merge it into 2.0.0.

PR #664, opened 2018-08-16
Return getsolves() promise

antifob:
Hi,

For a plugin I'm writing, I need to act on the result of the `getsolves()` call, but its promise isn't returned.

This patch makes `getsolves()` pass the promise returned by jQuery to the caller. Similar patches for other functions may be desired, but this solves my problem and see no benefit in investigating further; I prefer to let these be evaluated on a case-by-case basis.

### DIFF
--- a/CTFd/themes/core/static/js/chalboard.js
+++ b/CTFd/themes/core/static/js/chalboard.js
@@ -222,7 +222,7 @@ function updatesolves(cb){
 }
 
 function getsolves(id){
-  $.get(script_root + '/chal/'+id+'/solves', function (data) {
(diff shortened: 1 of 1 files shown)


ColdHeat:
This isn't a major priority for me since most of my focus is on 2.0.0 and I haven't gotten to rewriting the JS. Is there a reason you need this merged into master and you can't just use your own custom theme?

antifob:
I PRed this because I didn't want to force a theme on users. Your comment made me realize I can simply shadow the original and replace it myself using plugin assets. Thanks a lot!

I'm closing this since it's no longer needed. Thanks again.

PR #665, opened 2018-08-19
Wrong URL in email verification response email

CJHackerz:
URL path is given in verification mail ends up invalid because of following reason,
The current format is http://localhost:8000/ASDJdshifad32.... if user click on this verification links he/she ends up getting error 404 since token is added directly to the path and there will be no route defined for it and according to auth.py verification tokens are handled by /confirm route so correct format will be http://localhost:8000/confirm/ASDJdshifad32... hence {url}/confirm/{token}

### DIFF
--- a/CTFd/utils/__init__.py
+++ b/CTFd/utils/__init__.py
@@ -667,7 +667,7 @@ def sendmail(addr, text):
 def verify_email(addr):
     s = TimedSerializer(app.config['SECRET_KEY'])
     token = s.dumps(addr)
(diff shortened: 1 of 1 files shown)


ColdHeat:
Hi @CJHackerz, looks like you're failing a test. Are you sure that this is broken to begin with?

CJHackerz:
Oh lol, in my implementation I have not used url_for() method, instead I had directly given path to url="www.example.com" so participants in my CTF can receive correct link instead of http://localhost:8000/confirm/Arjfdjd...

ColdHeat:
I'm not sure this is still necessary so closing this PR. Feel free to re-open.

PR #670, opened 2018-09-01
Updated wrong description

amanaryan:
Updated the description shown when admin marks a submission as correct.  The team name and challenge name were exchanged in the original code.


ColdHeat:
I confirm that this is indeed an issue. Changing this to 2.0.0 where development is taking place.

ColdHeat:
My apologies, I ended up not using this fix and this is quite out of date so I'm going to go ahead and close this.

PR #672, opened 2018-09-05
reorganized Dockerfile to improve layer caching

frohoff:
Separately add `requirements.txt`, then run `pip install`, then add remaining files so as to not cause re-installation of python packages whenever any code files change. Should help make for faster docker-based dev cycle.


ColdHeat:
This looks reasonable. The only thing I think looks weird is `COPY . .` but it should be correct.

PR #673, opened 2018-09-05
Optional nginx support

frohoff:
Optional nginx docker-compose configuration to allow for improved performance:
* Caches static file responses from flask/gunicorn in nginx and in browser for 4h (1m for `user.css`) and responds with `304 Not Modified` status as appropriate
* Gzips text content (html, css, js, etc) if supported by client
* Buffers slow clients to prevent DoS impacts (http://docs.gunicorn.org/en/stable/deploy.html: *"we strongly advise that you use Nginx"*)

Run as follows:
```
docker-compose -f docker-compose.yml -f docker-compose.nginx.yml up
```

### DIFF
--- a/docker-compose.nginx.yml
+++ b/docker-compose.nginx.yml
@@ -1,0 +1,13 @@
+version: '2'
+
+services:
+  nginx:
+    image: nginx
+    ports:
+      - "80:80"
+    volumes:
+      - .data/nginx/logs:/var/log/nginx
+      - ./nginx.conf:/etc/nginx/nginx.conf
+    restart: unless-stopped
+    depends_on:
+      - ctfd
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,0 +1,50 @@
+events {
+  worker_connections 1024;
+}
+
+http {
+  proxy_cache_path /tmp/cache levels=1:2 keys_zone=cache:32m max_size=1G;
+#  error_log /dev/stdout info;
+#  access_log /dev/stdout;
+
+  upstream app {
+    server ctfd:8000;
+  }
+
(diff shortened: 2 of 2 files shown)


ColdHeat:
There's no need to have a second Dockerfile for nginx because you can treat it as a service and mount the nginx.conf file. Secondly, gevent is an acceptable alternative to a reverse proxy even according to the gunicorn docs. It might even be preferable to use waitress to avoid this.

I don't believe CTFd as a project should be setting up nginx for users at the moment. Users can use what they choose to setup and there's less to configure and debug. I don't think it's worth the headaches it will introduce.

frohoff:
Fair enough, though it is only optional (`docker-compose up` will not include nginx by default). Latest version volume-mounts the `nginx.conf` in case you change your mind. 

I'd be interested in reading up on the gevent-over-reverse-proxy stuff if you have a link.

ColdHeat:
Hmm I wouldn't say that gevent is preferred over a reverse proxy but it deals with the slow clients issue and is a more acceptable alternative in production to the default sync workers. 

> If you want to be able to handle streaming request/responses or other fancy features like Comet, Long polling, or Web sockets, you need to turn off the proxy buffering. When you do this you must run with one of the async worker classes.

http://docs.gunicorn.org/en/stable/deploy.html and also this gunicorn issue: https://git.io/fAEI9

I put some thought into this and I think it is okay to have nginx in docker-compose for 1.2.0 because it is likely we will have other services there as well. It might make the most sense to keep deployment unified. Granted, there are other considerations so I am going to close this and create an issue targeted for 2.0.0 instead.

Issue #603, opened 2018-04-04
Email whitelisting


ColdHeat:
How do we add the ability to whitelist emails such that we can only allow certain email addresses? 

Perhaps this should be implemented as a regular expression on the email address?

stefan2904:
Related to #441

ColdHeat:
Functionality for this exists in the 2.0.0 branch but is missing a way to configure the whitelisted domains at the moment.

ColdHeat:
This is now implemented in 2.0.0.

Issue #441, opened 2017-11-08
Restrict registration to whitelisted domains


ColdHeat:
We can restrict registration to whitelisted domains to make CTFd better at internal trainings.

ColdHeat:
This should be implemented in 2.0.0.

PR #675, opened 2018-09-11
Extension that allows to restrict the allowed users to configured domains or emails

pedromigueladao:
This extension allows the admin to configure a set of domains and usernames such that only users from these domains (or those addresses explicitly specified) can register.
Invalid e-mails and domains are silently discarded.

### DIFF
--- a/CTFd/admin/__init__.py
+++ b/CTFd/admin/__init__.py
@@ -158,28 +158,29 @@ def admin_config():
         utils.set_config("ctf_theme", request.form.get('ctf_theme', None))
         utils.set_config('css', request.form.get('css', None))
 
-        allowed_users = request.form.get('allowed_users', None)
+        white_listed_users = request.form.get('white_listed_users', None)
 
-        if allowed_users is None:
-            allowed_users = []
+        if white_listed_users is None:
+            white_listed_users = []
         else:
-            allowed_users = allowed_users.split('\n')
-        allowed_domains = []
-        allowed_mails = []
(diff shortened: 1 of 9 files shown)


pedromigueladao:
resolves #441 resolves #603

ColdHeat:
Hi @pedromigueladao. Thanks for putting in the time to write this and get tests working! I think this is better suited as a plugin to CTFd instead of a core feature at the moment. 

Also development is happening in the 2.0.0 branch so any major features should go there.

pedromigueladao:
Thanks @ColdHeat . I will need this for a course that I am going to teach this fall so I was developing over the current stable version.
Is any major difference between 2.0.0 and the current one? (to measure the effort taken to develop for that version)

pedromigueladao:
Also @ColdHeat  I see it available in the ctfd.io. Will you make it available for 2.0.0 (by default or as a plugin)?

ColdHeat:
@pedromigueladao 2.0.0 is a very large rewrite. It's likely that your plugin will break in 2.0.0. I would wait until it is stabilized and then try to develop it. 

However the current version works just fine so I don't think there's much issue with simply targeting 1.2.0 and then updating later. 

As for future versions yes email whitelisting is listed as part of 2.0.0. https://github.com/CTFd/CTFd/issues/603

ColdHeat:
2.0.0 now has whitelisting email domains built in so I'm going to close this.

PR #676, opened 2018-09-13
How to use this themes in detail

votinhthuong:
Hi, 

Can you explain detail to step by step how to install this themes for my CTFd?


ColdHeat:
Modern themes can be dropped into the themes folder inside of CTFd. From there you can choose the theme from the Config page of the admin panel.

PR #679, opened 2018-09-21
Add ability to specify packages to be installed with apk

overallcoma:
This addresses issue with using Google Captcha plugin in CTFd.  To successfully add this in Docker, packages need to be added with apk before adding packages with pip.

This allows plugin creators to add an "apk-requirements.txt" file to their plugins and have them installed at the build phase.

Will require modification to plugin guidelines.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,13 @@ COPY . /opt/CTFd
 
 VOLUME ["/opt/CTFd"]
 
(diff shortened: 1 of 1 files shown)


ColdHeat:
Couldn't you create a new Dockerfile or a child Dockerfile? I don't think it makes sense to tie the repo entirely to alpine...

Issue #682, opened 2018-09-28
Change the default filename of the PNG plot


SkypLabs:
Hello,

I've noticed that the default filename of the "Download plot as a png" feature is `newplot.png`. It would be better to name it following a more appropriate pattern such as `<Name of the CTF>.<date>.<time>.scoreboard.png` or something similar.

I can open a pull request if needed.

Thank you.

ColdHeat:
Thanks for reporting this @SkypLabs. I noticed it a long time ago but I was never able to figure out how to change the name. If you could let me know how to override that I can make the appropriate changes.

kara71:
Apparently the feature exists but badly documented : https://github.com/plotly/plotly.js/issues/2444 - I'll work on a fix.

SkypLabs:
Ok cool! Thanks for that.

Let me know if I can give you a hand.

kara71:
The fix is in #684 :)

PR #684, opened 2018-10-01
Change default filename for PNG export

kara71:
Fixes https://github.com/CTFd/CTFd/issues/682 .


ColdHeat:
Hi @kara71 can you apply this to the other graphs? Also is there a way to do this from the Plotly constructor at all?

kara71:
I will !
See the issue referenced in https://github.com/CTFd/CTFd/issues/682 - Plotly is currently working on a better way to name the outfile, but this remains the only way for now...

ColdHeat:
Thanks for figuring out how to do this!

PR #689, opened 2018-10-06
Fix access to unreleased challenges through /chals/<id> endpoint

lava:
We had to hot-patch this after we saw users exploiting it, submitting correct flags for challenges that weren't even released yet.

### DIFF
--- a/CTFd/challenges.py
+++ b/CTFd/challenges.py
@@ -133,6 +133,8 @@ def chal_view(chal_id):
     teamid = session.get('id')
 
     chal = Challenges.query.filter_by(id=chal_id).first_or_404()
+    if chal.hidden:
+        abort(404)
     chal_class = get_chal_class(chal.type)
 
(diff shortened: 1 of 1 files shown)


ColdHeat:
All development is happening on the 2.0.0 branch. 

Also you're telling me that users were able to guess flags for your challenges without any prior knowledge?

ColdHeat:
Ah actually I misunderstood the point of your PR.

ColdHeat:
Ruminating about this, how are people solving the challenge? The solve
endpoint prevents hidden challenges from being solved.

lava:
They were not getting the challenges marked as solved on the scoreboard, but we observed people submitting flags for challenges that hadn't been released yet. (offline challenges that consisted only of files) 

And once we made the challenge public, if we hadn't changed the flag they could have immediately submitted the flag they already knew.

PR #699, opened 2018-10-24
Code cleanup

jsfan:
* Removed unused imports
* Removed redundant/unreachable code
* Fixed up some convoluted logic

### DIFF
--- a/CTFd/challenges.py
+++ b/CTFd/challenges.py
@@ -1,19 +1,15 @@
-import json
 import logging
-import re
 import time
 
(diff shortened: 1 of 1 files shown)


ColdHeat:
This code is quite out of date. A benefit would be if you could implement a complete solution to unused imports like perhaps #823

jsfan:
This cleanup was a by-product of an unrelated code audit. I didn't realise that the next major version was imminent when I made the PR.

I might try contributing a pure linting fix time permitting.

PR #709, opened 2018-10-31
fix admin_export bugs

yozgif:
python 2.7  flask send_file function need attachment_filename type is unicode, not str


yozgif:
> Hi can you target this to the 2.0.0 branch? Also what were the arguments to the string (ctf name, etc) that caused this?

when ctf_name has some Chinese characters such as  汉字, admin_export can't work. Maybe all non-ascii characters can't work (untested).

ColdHeat:
This has been implemented upstream already. Thanks for the PR and reporting the issue.

PR #721, opened 2018-11-05
Remove @during_ctf_time_only from /challenges

alokmenghrajani:
Method already gracefully handles being requested before the CTF starts.

By removing @during_ctf_time_only, we get a nicer error message ("<ctf
name> has not started yet" vs a 403).


ColdHeat:
You'd need to target this to 2.0.0 and it's unlikely that that change will be accepted b/c of how heavily `during_ctf_time_only` is used there.

PR #732, opened 2018-11-14
setting APPLICATION_ROOT from env var

phretor:
This allows to set `APPLICATION_ROOT` externally by setting an env variable, making hosting under a subfolder easier. The `dockerfile-entrypoint.sh` has been adapted accordingly, so the `APPLICATION_ROOT` env var can be passed when calling `docker` or set in the `docker-compose.yml` file.

### DIFF
--- a/CTFd/config.py
+++ b/CTFd/config.py
@@ -197,6 +197,14 @@ class Config(object):
     OAUTH_CLIENT_ID = None
     OAUTH_CLIENT_SECRET = None
 
+    '''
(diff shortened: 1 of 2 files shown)


ColdHeat:
Is it necessary to set `APPLICAITON_ROOT` inside of docker-entrypoint as well as config.py?

phretor:
@ColdHeat it's not necessary to set it in both. I just provided sane defaults in there.

Issue #726, opened 2018-11-06
Minimum MariaDB database version requirement.


GrahamDumpleton:
**Environment**:

  - CTFd Version/Commit:

2.0.0 branch

  - Operating System:

MacOS X, but using docker so doesn't matter.

  - Web Browser and Version:

N/A

**What happened?**

Setup fails when using MariaDB 10.2.8.

Note that it is assumed that this is a problem with MariaDB 10.2.8 and that newer version will possibly work.

SQLAlchemy does warn:

```
/opt/app-root/lib/python3.6/site-packages/sqlalchemy/dialects/mysql/base.py:1909: SAWarning: MariaDB (10, 2, 8) before 10.2.9 has known issues regarding CHECK constraints, which impact handling of NULL values with SQLAlchemy's boolean datatype (MDEV-13596). An additional issue prevents proper migrations of columns with CHECK constraints (MDEV-11114).  Please upgrade to MariaDB 10.2.9 or greater, or use the MariaDB 10.1 series, to avoid these issues.
```

So possibly hitting one of the problems described.

Creating this issue more as documentation for others who may find this. Close issue if is the MariaDB version.

Note that you also can't use MariaDB 10.1 either. In that case the problem appears to be that ``python manage.py db upgrade`` fails because SQL is output the syntax of which MariaDB 10.1 doesn't understand. Again stating this more for documentation in case others have issue.

Using MySQL 5.7 does work though, so using that instead.

**What did you expect to happen?**

That it worked.

**How to reproduce your issue**

Use MariaDB 10.2.8.

**Any associated stack traces or error logs**

```

[2018-11-06 04:44:20,502] ERROR in app: Exception on /setup [POST]
--
  | Traceback (most recent call last):
  | File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1193, in _execute_context
  | context)
  | File "/opt/app-root/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 509, in do_execute
  | cursor.execute(statement, parameters)
  | File "/opt/app-root/lib/python3.6/site-packages/pymysql/cursors.py", line 170, in execute
  | result = self._query(query)
  | File "/opt/app-root/lib/python3.6/site-packages/pymysql/cursors.py", line 328, in _query
  | conn.query(q)
  | File "/opt/app-root/lib/python3.6/site-packages/pymysql/connections.py", line 516, in query
[...]

PR #762, opened 2018-11-28
Update mariadb to 10.4 based on #726

ColdHeat:
* Update mariadb to 10.4
* Remove `LOG_FOLDER` from docker-compose.yml


ColdHeat:
@ConnorNelson That has to do with `docker-compose` mounting the CTFd source directory inside of the container so changes in the host work in the container without rebuilding. You could remove `- .:/opt/CTFd:ro` as well. 

A recent commit re-enabled this logging which causes this issue.

PR #765, opened 2018-11-29
Make sure update.read() is not str then decode

mayoneko:
In my environment 500 error came when admin tried to make a challenge.
The errors that occurred were as follows.
`AttributeError: 'str' object has no attribute 'decode'`
For this reason, make sure `update.read()` is not `str` then `decode('utf-8')` .


ColdHeat:
This change feels off. I'll handle this one. Thank you for reporting the issue though.

mayoneko:
I am not good at Python and English, so finally I understood your intention just now. Thank you very much for reading unskilled English. I'm looking forward to future updates.

PR #777, opened 2018-12-03
Fix admin cannot modify verified status in Edit User

ramadistra:
The Edit User page has a form field to modify the user's Verified, Hidden and Banned status. Both the Hidden and Banned status can be modified by an admin, however the Verified status cannot be modified. This appears to be a bug.

Edit User
  
There are two potential solutions for this bug:
- Remove the Verified field from the form
- Allow admin to modify `verified` field in `UserSchema`

In this pull request I propose that admins should be allowed access to modify the `verified` field in `UserSchema`. 

This change has two advantages:
- It makes the existing Edit User form work as intended. 
- It allows admins to manually verify users, which enables contests to employ user verification without setting up a mail server.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,9 @@
+coverage:
+  status:
+    project:
+      default:
+        # Fail the status if coverage drops by >= 1%
+        threshold: 1
+    patch:
+      default:
+        threshold: 1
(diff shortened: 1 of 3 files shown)


ramadistra:
Hi @ColdHeat , I've modified the admin patch user test to exercise the new functionality.

Issue #570, opened 2018-02-13
Translations


ColdHeat:
Jinja has translation support.... Is it possible for us to implement translations for the common text? 

My main concern would be the modals...

frankli0324:
consider jquery i18n plugins? needn't play with python code

kaardeco:
There is a possibility that a master file with all the words that people were just translating would make it easier and many non-programmers could help.

If it is not viable, it could at least have a readme.md showing which files to translate.

vient:
Bumping the issue.
Any way to translate that won't brake over non-major updates would be nice.

----

> If it is not viable, it could at least have a readme.md showing which files to translate.

I don't think it is a good idea, using proper localization method shouldn't be that hard. Regarding translating itself, [LibreOffice is using](https://wiki.documentfoundation.org/Translating_LibreOffice) Pootle or Weblate to crowdsource the localization, something like this may be applicable to CTFd too.

socketz:
You can combine crowdtranslation with gettext libraries using .po files and .pot template files and is extended in translators community. Zanata.org it's another good alternative should be considered.

This post explain well the implementation of gettext in python projects:
https://simpleit.rocks/python/how-to-translate-a-python-project-with-gettext-the-easy-way/

ColdHeat:
I did some research and I think the right approach here is to leverage `.po` files because of the standard but also translate them to a JSON file format that can be consumed by the front end. I do not believe that the backend needs to handle much translation so the usage of `.po` files is really just for the standardization. 

I liked https://github.com/wikimedia/jquery.i18n because of the data attribute solution but another library could be used if it worked out well. 

This should be accomplished with:
- A new folder to store translation files. 
- Each supported language should have a .po file and a .json file that's used by the front end
- The core theme and admin panel should support translation
- We should detect user language preferences with the `Accept-Language` header and probably store it in a cookie. 
- Language switching I think could be accomplished with a switcher in the footer but I'm not concrete on this. 

Happy to work with someone on a PR on this if you're willing to tackle it.

faelcorreia:
Hello. I'm translating CTFd again for my native language (pt-BR), using an updated core theme. A lot of labels are rendering on backend now, and full frontend-only translation is becoming quite difficult.
Are you willing to move all text from the core theme from the frontend to backend now?
Maybe I could help with this, but I need to understand what you are thinking about translation on the core theme.

ColdHeat:
[...]

PR #780, opened 2018-12-04
create core_zh-cn theme

whirlwind110:
I will write a theme suitable for Chinese user.

1, I have translated the core theme to Chinese language.(also can use Jinja2 Extensions,but I'm a phper.https://github.com/CTFd/CTFd/issues/570)
2,My company has held too much  CTF,and we have many beautiful & cool themes and useful plugins ,I want to share it.


ColdHeat:
Hi @whirlwind110, thanks so much for PR'ing this! Unfortunately I don't think creating child themes is enough to solve the translation problem and I don't have the bandwidth to support multiple core themes. 

For translations we would need to solve this generically versus child themes. 

But what you could do is separate out your theme into a new repo and then add it over to the themes repo: https://github.com/CTFd/themes !

PR #782, opened 2018-12-05
Fix divison-by-zero when adding requirement to dynamic challenge

ptr0x01:
I noticed that when I wanted to add a pre-requisite challenge to a dynamic challenge, I received a divison-by-zero error.

During a requirement addition, only the new required challenge id is sent. Because the initial, minimum and decay values are not set, the update path in the __init.py set them to a default 0. 
Later on, when a calculation takes place with these values, a divison-by-zero happens before anything gets committed to the database.

I've added a simple check before other calculations to check if data received from client includes a requirements (is not None). If there is data there, it means the request is a requirement addition so we can just commit those changes and return. At the same time, updating values and other data will not include a requirement field so everything else will work as it does currently.

### DIFF
--- a/CTFd/plugins/dynamic_challenges/__init__.py
+++ b/CTFd/plugins/dynamic_challenges/__init__.py
@@ -89,6 +89,9 @@ class DynamicValueChallenge(BaseChallenge):
         data = request.form or request.get_json()
 
         for attr, value in data.items():
+            # We need to set these to floats so that the next operations don't operate on strings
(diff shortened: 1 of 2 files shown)


ColdHeat:
Thanks for reporting this issue. I'll take a look at it when I have the time tomorrow. Feel free to ping me if it slips!

ptr0x01:
Thanks, I went ahead and added the unit test for this check to improve coverage.

Issue #725, opened 2018-11-06
Use sqlalchemy Python script to test availability of database.


GrahamDumpleton:
**Environment**:

  - CTFd Version/Commit:

2.0.0 branch

  - Operating System:

MacOS X, but building/deploying container images.

  - Web Browser and Version:

N/A

**What happened?**

The check using ``mysqladmin`` in ``docker-entrypoint.sh`` will fail if the target system only has MySQL libs package and doesn't have MySQL clients package.

Further, the wiki says:

> My personal suggestion is MySQL or MariaDB, but PostgreSQL will work just fine.

The ``docker-entrypoint.sh`` script is hard wired to MySQL and expects MySQL clients package to be installed. Would be nice that if also want to use PostgreSQL and only provide alternate ``Dockerfile`` and not change anything else, that this would work and didn't have to rewrite it. Similarly, if use alternate build system like S2I for OpenShift, that can reuse existing scripts unchanged.

**What did you expect to happen?**

Would hope database check is database agnostic and work with only database libs available.

**How to reproduce your issue**

In my case hit it as building from source using S2I (Source-to-Image) builder in OpenShift. The Python base images do not include MySQL clients package.

**Any associated stack traces or error logs**

```
Waiting for ctfd-mariadb to be ready
  | /opt/app-root/src/docker-entrypoint.sh: line 33: mysqladmin: command not found
  | /opt/app-root/src/docker-entrypoint.sh: line 33: mysqladmin: command not found
  | /opt/app-root/src/docker-entrypoint.sh: line 33: mysqladmin: command not found
  | /opt/app-root/src/docker-entrypoint.sh: line 33: mysqladmin: command not found
 ...
```

**Solution**

Suggest database agnostic check be used which uses small Python script and ``sqlalchemy``.

```
import os
from sqlalchemy import create_engine
database = create_engine(os.getenv('DATABASE_URL'))
database.execute('select 1')
```

This has added benefit that also checks that required Python packages installed correctly and work before starting up the application.

It should also technically work for SQLite if someone wanted to override ``DATABASE_URL`` and use SQLite with file system path, to directory in a mounted persistent volume, rather than have database created in default location of source code directories where not possible to mounted a persistent volume. Although in SQLite case the database file would technically need to exist first.

ColdHeat:
This is vaguely reasonable but in the wiki (which needs MAJOR rewriting): 

[...]

PR #801, opened 2018-12-12
PostgreSQL support for docker

tmarlok88:
This adds postgreSQL support to the docker configuration.
Based on the fix proposed by @GrahamDumpleton in #725

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:2.7-alpine
 RUN apk update && \
-    apk add python python-dev libffi-dev gcc make musl-dev py-pip mysql-client git openssl-dev
+    apk add python python-dev libffi-dev gcc make musl-dev py-pip mysql-client git postgresql-dev
 
 WORKDIR /opt/CTFd
 RUN mkdir -p /opt/CTFd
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM python:2.7-alpine
 RUN apk update && \
-    apk add python python-dev libffi-dev gcc make musl-dev py-pip mysql-client git openssl-dev
+    apk add python python-dev libffi-dev gcc make musl-dev py-pip mysql-client git postgresql-dev
```
ColdHeat: Why are you replacing `openssl-dev` with `postgresql-dev`?

ColdHeat:
Are you in a position where you're unable to run MySQL or MariaDB? There are no guarantees that Postgres will be supported forever and I'm not sure it makes sense to have every installation installing `psycopg2` if the "official" supported database is MySQL.

ColdHeat:
After some thought I don't think having the idea that Postgres is officially supported is worthwhile long-term. Users could create a child Dockerfile to install the required packages to get Postgres running. 

The `database_ready` function makes a good amount of sense. Can you update your PR to remove everything but that change?

ColdHeat:
Another point I should make is that alpine is actually going to be switching back to openssl-dev instead of using libressl-dev. https://lists.alpinelinux.org/alpine-devel/6073.html

PR #812, opened 2018-12-27
Fix wrong user mode in challenge_solves_box

invom:
I found that even when using `Users` mode, the url of the person who solved the problem is still in `/teams/id` format, not `/users/id`. So I added the result of the current user mode to the `data[]` returned by `/api/v1/challenges/id/solves` to generate the correct url.
Although this would insert a lot of redundant `user_mod` to `data[]`, but this is the solution I can think of with the least change to the code : )


ColdHeat:
Hi @invom, thanks for pointing this out. I'll think about the right fix here. I think you're pretty close but I think it's probably better to change `'user_mode': user_mode` to `account_type`.

invom:
Hi @ColdHeat, I forgot to review the code specification before I submitted the code last time, sorry :laughing:

ColdHeat:
Okay, the way that I think this should work is that the API should return `account_url` which will point to the correct resource instead of having the client build out the URL on their own.

PR #814, opened 2018-12-30
added virtual env

Abemarkar23:
so that users can safely install all python dependencies without messing up other dependencies of their other projects.


ColdHeat:
Closing this. It is up to the individual user to decide if they want to use a virtualenv or not. CTFd isn't a project that needs to push any best practices.

Issue #823, opened 2019-01-06
Lint with flake8


ColdHeat:
We should lint with flake8 so that we can catch some bugs at build time. 

A safe start would be:
```
flake8 --ignore E501 CTFd/
```

jsfan:
I am planning to look into this in the next days. However, you suggest that you would like this configuration to be able to catch bugs at build time. So, is this what you are after?

1. Run pylint/flake8/pycodestyle over the code and address all errors
2. Integrate a call to the same tool in the Travis configuration to make sure violations are caught at buid time

Probably rather independent of the approach, a PR would likely have to be merged pass because it would certainly develop conflicts quickly. How do we coordinate that?

ColdHeat:
Sorry I think build time is the wrong word to use. I think at test time is a better phrase. 

I think I can start a 2.1 branch and then you can PR in the flake8 changes and slowly remove the violations. I imagine most of them are going to be about unused imports.

jsfan:
There definitely are a good number of violations but I have not made any effort yet to determine how many different ones it might be.

I'm happy to do the manual cleanup with a good number of small pull requests which should allow quick merging without too much hassle.

I suspect the end goal still is to enforce these same rules in Travis eventually, right?

ColdHeat:
Yes the build should fail on flake8 errors just like how it fails on pycodestyle errors now.

jsfan:
I've managed to get through all the unused imports in the CTFd package and fully rebased against master. As you have not created a 2.1 branch, yet, I'm unsure if I should send a PR.

What is still outstanding is all other flake8 errors (the Travis config will only check for unused imports for now) and going through the unit tests as well.

ColdHeat:
@jsfan here you go: https://github.com/CTFd/CTFd/tree/2.1.0-dev 😄

jsfan:
That was quick! The PR will need to be squashed. I just did the cleanup in stages to see if anything was breaking.

I've also had a look at a few other violations and noticed that my workarounds for some only seemigly unused imports have resulted in unused variables. AFAIK, flake8 does not allow declaring a local exception with a code comment the way pylint does. So, I'll have to think about how to address these remaining issues.

ColdHeat:
I haven’t tried it myself at the moment but I believe you can use # noqa
and the exceptions you want to ignore?

jsfan:
[...]

PR #842, opened 2019-01-20
Flake8

jsfan:
Removes unused imports from package CTFd.

* flake8 is not integrated into Travis configuration to check for unused imports.
* No checks on unit tests
* Other violations flake8 detects have not been addressed.

Partially addresses #823.


ColdHeat:
I think we should ignore the same lint errors in pycodestyle and flake8 so `E501,E712,E402`. Can you remove the noqa's for those lint errors?

jsfan:
Makes sense.

jsfan:
I've now fixed up the tests as well. That actually has the nice side effect of making some of the imports much more explicit as well.

Issue #867, opened 2019-02-03
Exports may have a different type per db for challenge requirements


ColdHeat:
I've seen a few times where the resulting challenges.json in an import had different types. I've seen it be a string and also a dictionary. The question here is does this affect teh re-importing and which type should we consider the correct one?

ColdHeat:
Note that we also need a backwards-compat hack to import requirements that's JSON as a string.

ColdHeat:
The overall suspicion is that SQLAlchemy 1.3.0 will be released soon.

ColdHeat:
The correct value is JSON. So we mostly just need the backwards-compat hack to get JSON.

ColdHeat:
Closed by #931

PR #881, opened 2019-02-10
Remove SQLiteJSON from 8369118943a1_initial_revision.py

ColdHeat:
* Remove `SQLiteJson` from `8369118943a1_initial_revision.py`


ColdHeat:
Related to #867

PR #909, opened 2019-03-20
fix 2 problems caused by hidden user

xmsec:
1. Dynamic challenge value will be changed when hidden user submit flag, and it come to be normal when user not hidden submit flag. Another way to make value normal is to update the challenge in admin page.
2. Place in normal user's profile page is not equal to that in scoreboard.

### DIFF
--- a/CTFd/models/__init__.py
+++ b/CTFd/models/__init__.py
@@ -433,13 +433,6 @@ class Users(db.Model):
             ) \
                 .join(sumscores, Users.id == sumscores.columns.user_id) \
                 .order_by(sumscores.columns.score.desc(), sumscores.columns.id)
-        elif self.hidden:
-            standings_query = db.session.query(
-                Users.id.label('user_id'),
-            ) \
-                .join(sumscores, Users.id == sumscores.columns.user_id) \
-                .filter(Users.banned == False) \
-                .order_by(sumscores.columns.score.desc(), sumscores.columns.id)
         else:
             standings_query = db.session.query(
                 Users.id.label('user_id'),
--- a/CTFd/models/__init__.py
+++ b/CTFd/models/__init__.py
@@ -433,12 +433,19 @@ class Users(db.Model):
             ) \
                 .join(sumscores, Users.id == sumscores.columns.user_id) \
                 .order_by(sumscores.columns.score.desc(), sumscores.columns.id)
-        else:
+        elif self.hidden:
             standings_query = db.session.query(
                 Users.id.label('user_id'),
             ) \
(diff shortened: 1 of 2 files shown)


ColdHeat:
This code looks incorrect. I'm pretty sure you can just apply `Users.banned == False, Users.hidden == False` if the user isn't an admin. 

Secondly please provide a test.

ColdHeat:
You could also invert your if statement in the dynamic challenges code to eliminate the unnecessary pass statement.

ColdHeat:
Also in your code to detect if the user is hidden, I do not think you can rely on the team object. I believe you need to check both the user and team objects.

xmsec:
For the second problem, i think you are right.  If the user isn't an admin just apply what as you said.
For example, http://139.199.100.82:8001, for user test3:test3. And admin is my account name, password the same.
For the first one, i'm checking it.

ColdHeat:
By test I mean that we need to have a unit test to make sure this bug doesn't show up later down the road for some reason. But the example is also useful.

xmsec:
OK, i misunderstood it. But i am not familiar with unit test of CTFd.
And for the first one, i think `if user.hidden or (team and team.hidden) :` maybe right.

ColdHeat:
Hi @xmsec, think you can update this PR soon? I would like to release a new bugfix version in a few days.

ColdHeat:
Reading through this code, how does the score for the challenge change if a hidden user solves it? I'm not seeing the logic that this fails? Can you provide a test that fails?

ColdHeat:
I just tried this myself and the value does not change when a hidden user submits a solve.

ColdHeat:
Fixing the hidden users place issue in #918. If the dynamic challenge issue is valid please provide a minified test case and re-open.

xmsec:
First, create 2 user: test1 is visible and test2 is hidden.
Second, create a dynamic challenge with 1000 value, 50 min and 20 decay.
Third, test1 submit flag with challenge value not changed.
Finally, test2 submit flag with challenge value changed to 998.

As value is comluputed and updated but solve_count is not increased.

ColdHeat:
I don't think your interpretation of the issue is incorrect. The query here does not count hidden users. Try having more hidden users solve your test challenge and you'll notice that the score does not decrease.

ColdHeat:
Thinking on it a little more there is a sort of idea here but it's hard to identify what the solution is. Any PR to fix this will need to have a test case testing the fix.

xmsec:
I have noticed that before. But if a challenge is hard and only one solve it, then it may be a terrible experience. 
I understand your new reply and maybe what i can do is to let you know this little bug.  :)

PR #913, opened 2019-03-20
Fix email confirmation log line

grant-h:
It looks like this log output was copied and pasted from the reset password route. Also, the `{name}` format specifier resolves to `None` in cases where the user is not logged in (`session.get('name')` is None), but confirms the CTFd email.

Before:
```
[03/16/2019 18:59:30] x.x.x.x -  successful password reset for None
```
After:
```
[03/16/2019 19:10:39] x.x.x.x - successful confirmation for Test Team
```

### DIFF
--- a/CTFd/auth.py
+++ b/CTFd/auth.py
@@ -48,7 +48,7 @@ def confirm(data=None):
 
         user = Users.query.filter_by(email=user_email).first_or_404()
         user.verified = True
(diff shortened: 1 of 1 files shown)


ColdHeat:
Thanks for the PR! 

This looks reasonable, not sure whats up with the Travis checks though. Will merge as soon as all checks pass.

PR #918, opened 2019-03-22
Dont calculate hidden users in place

ColdHeat:
* Users that are hidden shouldn't be factored into the calculation of place for other players. 
* Clean up a few other tests.


ColdHeat:
@xmsec does this PR resolve your issue?

xmsec:
Yes, it does.

PR #919, opened 2019-03-22
Record a new solve before calculating the new dynamic value.

ColdHeat:
* Record a new solve before calculating the new dynamic value.


ColdHeat:
@xmsec does this resolve your other issue with dynamic challenges?

xmsec:
Great idea! Sure it is.

PR #922, opened 2019-03-27
Improve SSE code to detect disconnected clients

ColdHeat:
* Improve SSE code to detect disconnected clients
* Poll clients every second with an empty SSE
* Bump default `WORKERS` value to 4 in docker-compose and Dockerfile


ColdHeat:
This was ultimately a little too experimental. gevent proved to be a cleaner solution https://github.com/CTFd/CTFd/pull/924

PR #923, opened 2019-03-28
Return 404 error when viewing hidden/banned users

C0deH4cker:
Users marked as hidden (such as admins when testing challenges) are not fully hidden. It is still possible to view their user profile and even see challenges that this user has solved before the competition has started. This PR fixes this by returning a 404 Not Found error upon trying to view anything about hidden users, even via the API.

### DIFF
--- a/CTFd/api/v1/users.py
+++ b/CTFd/api/v1/users.py
@@ -32,7 +32,7 @@ users_namespace = Namespace('users', description="Endpoint to retrieve Users")
 class UserList(Resource):
     @check_account_visibility
     def get(self):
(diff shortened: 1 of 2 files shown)


ColdHeat:
@C0deH4cker awesome, thanks for the PR? Mind fixing up the test that's failing here? 

This looks mostly correct but in terms of the /user/solves and /user/fails endpoints, I think you're probably going to cause failures for admins when they need user data from the API.

ColdHeat:
Thanks for the PR, I'm going to take this over at #932 but thanks for sharing your initial changes.

Issue #852, opened 2019-01-25
Replace Flask-SocketIO with custom SSE code


ColdHeat:
Original issue with some discussion on rationale: #600 

The following are libraries and links that implement SSE behavior in Flask. However none of these should really be used without modifications. The reason is that we should be able to support the debug server AND a production server with redis configured. 

https://github.com/singingwolfboy/flask-sse
https://stackoverflow.com/a/51969441
http://flask.pocoo.org/snippets/116/

PR #924, opened 2019-03-28
Improve SSE code to detect disconnected clients

ColdHeat:
* Improve SSE code to detect disconnected clients
    * Use gevent to timeout blocking calls for new SSE
    * Poll clients every ten seconds with a "ping" SSE
* Bump default `WORKERS` value to 4 in docker-compose and Dockerfile


ColdHeat:
https://github.com/CTFd/CTFd/issues/852

PR #937, opened 2019-04-07
Update the challenge name field to required

s-nirali:
In the `admin/challenges` page, if a challenge was created without a
name, then the only way to open the challenge is by using its ID in the
URL as `admin/challenges/<ID>` since the "Name" column contains the
link to the challenge and "Name" of a challenge is not a required
field.

### DIFF
--- a/CTFd/plugins/challenges/assets/create.html
+++ b/CTFd/plugins/challenges/assets/create.html
@@ -6,7 +6,7 @@
 				The name of your challenge
 			</small>
 		</label>
-		<input type="text" class="form-control" name="name" placeholder="Enter challenge name">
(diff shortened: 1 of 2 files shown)


ColdHeat:
This is an interesting issue but mostly cosmetic. We have previously allowed challenges without names but based on the aesthetic I think I'm okay with this. 

Can you target this for the 2.1.0-dev branch?

s-nirali:
I've rebased the commit on branch `2.1.0-dev`.

ColdHeat:
I thought about this for some time and I would rather not enforce that names are provided. This is better solved with an aesthetic change making the rows themselves clickable.

PR #942, opened 2019-04-07
Make awards polymorphic

### LINKED ISSUE #793: Polymorphic awards table
The awards table should probably be polymorphic so that plugins can create new kinds of Awards or manipulate them as needed.

ColdHeat:
* Add a `type` column to Awards
* Add a migration to add `type` to Awards for existing installations
* Closes #793


ColdHeat:
I manually verified that the `type` column is added and set to standard when importing an old export.

PR #944, opened 2019-04-08
from six import string_types

cclauss:


migrations/1_2_0_upgrade_2_0_0.py
```diff
@@ -15,19 +15,15 @@
...
         return int(value)
-    elif value and isinstance(value, six.string_types):
-        if value.lower() == 'true':
-            return True
-        elif value.lower() == 'false':
-            return False
-        else:
-            return value
+    elif value and isinstance(value, string_types):
+        return {'true': True, 'false', False}.get(value.lower(), value)
```
ColdHeat: Please don't alter code here. This file does not have test coverage and is sensitive for upgrading old CTFd instances.

migrations/1_2_0_upgrade_2_0_0.py
```diff
@@ -55,7 +51,7 @@ def cast_bool(value):
         old_data.pop('alembic_version')
 
     print('Current Tables:')
-    for table in old_data.keys():
+    for table in old_data:
```
ColdHeat: This likely is fine but I would rather you not modify it.

migrations/1_2_0_upgrade_2_0_0.py
```diff
@@ -248,7 +244,7 @@ def cast_bool(value):
     manual = []
     not_created = []
     print('MIGRATING extra tables')
-    for table in old_data.keys():
+    for table in old_data:
```
ColdHeat: This likely is fine but I would rather you not modify it.

PR #939, opened 2019-04-07
Fix missing import and undefined names

cclauss:
[flake8](http://flake8.pycqa.org) testing of https://github.com/CTFd/CTFd on Python 3.7.1

$ __flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics__
```
./CTFd/__init__.py:22:5: F821 undefined name 'reload'
    reload(sys)
    ^
./CTFd/utils/__init__.py:22:26: F821 undefined name 'unicode'
    string_types = (str, unicode)
                         ^
./CTFd/utils/__init__.py:23:17: F821 undefined name 'unicode'
    text_type = unicode
                ^
./migrations/1_2_0_upgrade_2_0_0.py:24:38: F821 undefined name 'six'
    elif value and isinstance(value, six.string_types):
                                     ^
4     F821 undefined name 'reload'
4
```
__E901,E999,F821,F822,F823__ are the "_showstopper_" [flake8](http://flake8.pycqa.org) issues that can halt the runtime with a SyntaxError, NameError, etc. These 5 are different from most other flake8 issues which are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

PR #945, opened 2019-04-08
Use feature detection instead of version detection 

cclauss:
Redo #939

### DIFF
--- a/CTFd/utils/__init__.py
+++ b/CTFd/utils/__init__.py
@@ -3,15 +3,9 @@ import six
 from flask import current_app as app
 from CTFd.cache import cache
 
-if six.PY2:
-    string_types = (str, unicode)  # noqa: F821
-    text_type = unicode  # noqa: F821
-    binary_type = str
-else:
-    string_types = (str,)
-    text_type = str
-    binary_type = bytes
-
+string_types = six.string_types
+text_type = six.text_type
+binary_type = six.binary_type
 markdown = mistune.Markdown()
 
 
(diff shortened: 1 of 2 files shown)


CTFd/utils/__init__.py
```diff
@@ -28,20 +22,12 @@ def _get_config(key):
         if value and value.isdigit():
             return int(value)
         elif value and isinstance(value, six.string_types):
-            if value.lower() == 'true':
-                return True
-            elif value.lower() == 'false':
-                return False
-            else:
-                return value
+            return {'true': True, 'false', False}.get(value.lower(), value)
```
ColdHeat: SyntaxError here causing your test failures I believe. I don't see this as more readable, is there some gain to be had here besides the condensed code? If not please revert this change.

CTFd/utils/__init__.py
```diff
@@ -28,20 +22,12 @@ def _get_config(key):
...
 
 
 def get_config(key, default=None):
     value = _get_config(key)
-    if value is None:
-        return default
-    else:
-        return value
+    return default if value is None else value
```
ColdHeat: This reads okay but I am not entirely sure what the benefit is here besides slightly more condensed code. If there's no benefit besides that please revert.

cclauss: Would __return value or default__ work in this situation?

cclauss:
Add truth_dict
When we can cut the number of lines of code in half without loss of readability, we should attempt to do so.  Dict lookups should also be encouraged for performance reasons.

PR #950, opened 2019-04-12
Disable native asynchronous IO for MariaDB

thocou:
Running on windows hosts with docker, the database fails to initialize
as the native asynchronous IO is not supported on the host.

See https://dev.mysql.com/doc/refman/5.5/en/innodb-parameters.html#sysvar_innodb_use_native_aio

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,7 +36,7 @@ services:
     networks:
         internal:
     # This command is required to set important mariadb defaults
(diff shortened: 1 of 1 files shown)


ColdHeat:
Thanks for pointing this out. I feel like this is something that should be handled upstream by MySQL or its Docker image itself. I don't really think it belongs in CTFd itself. 

I'm confused though, shouldn't Docker for Windows properly virtualize this since it'd be running in a virtual machine anyway?

PR #984, opened 2019-05-06
Optimize docker entrypoint script

ProNic-QY:
Optimize docker entrypoint script, only check mysql validity when specify $DATABASE_URL is mysql connection url. After optimized, $DATABASE_URL can config for sql lite db file path in container.


ColdHeat:
I'm not sure what the benefit is here? Is this just so you can redirect the sqlite database somewhere else? Where would you want it to be? 

I'd rather not promote the usage of sqlite beyond development/testing...

PR #997, opened 2019-05-17
added specific error message when doing requests without csrf nonce

eyJhb:
Seems reasonable to give this error message, when the csrf nonce is not provided for a request.

### DIFF
--- a/CTFd/utils/initialization/__init__.py
+++ b/CTFd/utils/initialization/__init__.py
@@ -211,7 +211,13 @@ def init_request_processors(app):
                     abort(403)
             if request.content_type != "application/json":
                 if session["nonce"] != request.form.get("nonce"):
-                    abort(403)
+                    return (
+                        render_template(
+                            "errors/403.html",
+                            error="Invalid csrf nonce provided for this request",
+                        ),
+                        403,
+                    )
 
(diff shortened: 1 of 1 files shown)


ColdHeat:
I'm open to having a more descriptive error message but this is too descriptive. Something vaguer asking the user to try again would make more sense.

eyJhb:
But, why does it need to be vague at all? This is mostly because, as a developer debugging this error really really sucks. Maybe just a log statement or something?

ColdHeat:
Well a non technical user could see this page and wonder what happened. As I've said before, if you want to develop on CTFd it is expected that you be able to navigate the CTFd code and debug for yourself. A logging message also sounds fine.

eyJhb:
It is a CTF platform, and `CSRF` is a Google search away. The current error message isn't even valid, to what actually happens. I do believe in not throwing technical stuff in the face of the users, but this isn't all the technical at all :)

Debugging middleware isn't that fair for entrace users tbh. when the error messages are as vague as they are

ColdHeat:
CTFd is also used in schools and by non-programmers. CSRF is technical and asking users to Google it is not part of the design goals for CTFd.

PR #1029, opened 2019-06-19
Documentation: plugin creation walkthrough

lodi-g:
As I discussed with you on Slack, a small walkthrough on basics to create a plugin; to help beginners getting started.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,6 +25,7 @@ CTFd is written in Python and makes use of the Flask web framework.
    configuration
    themes
    plugins
+   plugin walkthrough
    contributing
 
 Indices and tables
--- a/docs/plugin walkthrough.rst
+++ b/docs/plugin walkthrough.rst
@@ -1,0 +1,339 @@
+Real-life plugin walkthrough
+============================
+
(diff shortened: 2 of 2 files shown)


ColdHeat:
The way this is written it seems more like a blog post than an official documentation entry. Is there a better way to reorganize these thoughts so that they fit into the docs and have less of a personal perspective?

lodi-g:
That was mostly the point. The docs looks complete to me, but it lacks of examples, and I personally prefer (and know other people that does too) a walkthrough/long guide to give insights.

If it doesn't meet what you'd like to have in the docs, I'll simply blog-post it on Medium.

ColdHeat:
I thought about this a bit and I like the idea and I learned a bit from your examples, but I don't think this belongs in the docs. I think a better idea would be to post it on medium or similar and then I can add an external resources section linking to these externally hosted resources.

lodi-g:
Here you go! https://medium.com/@gregoire.lodi/writing-ctfd-plugins-a-beginner-walkthrough-7d0b93eb89e7

Thank you!

Issue #257, opened 2017-05-05
how to fix CommandError: Path doesn't exist: 'migrations'?


Colbert337:
I use Apache/2.4.7 (Ubuntu14.04) mod_wsgi/4.1.3 Python/2.7.6 configured，but it failed and the http Status Code is 500 (Internal Server Error)

the Apache2  error_log is :  

[Fri May 05 11:35:21.288751 2017] [wsgi:error] [pid 3973:tid 140195138877184] [remote 14.17.22.33:38337] mod_wsgi (pid=3973): Target WSGI script '/var/www/html/CTFd/wsgi.py' cannot be loaded as Python module.
[Fri May 05 11:35:21.288843 2017] [wsgi:error] [pid 3973:tid 140195138877184] [remote 14.17.22.33:38337] mod_wsgi (pid=3973): Exception occurred processing WSGI script '/var/www/html/CTFd/wsgi.py'.
[Fri May 05 11:35:21.288936 2017] [wsgi:error] [pid 3973:tid 140195138877184] [remote 14.17.22.33:38337] Traceback (most recent call last):
[Fri May 05 11:35:21.289031 2017] [wsgi:error] [pid 3973:tid 140195138877184] [remote 14.17.22.33:38337]   File "/var/www/html/CTFd/wsgi.py", line 5, in <module>
[Fri May 05 11:35:21.289167 2017] [wsgi:error] [pid 3973:tid 140195138877184] [remote 14.17.22.33:38337]     application = create_app()
[Fri May 05 11:35:21.289252 2017] [wsgi:error] [pid 3973:tid 140195138877184] [remote 14.17.22.33:38337]   File "/var/www/html/CTFd/CTFd/__init__.py", line 49, in create_app
[Fri May 05 11:35:21.289341 2017] [wsgi:error] [pid 3973:tid 140195138877184] [remote 14.17.22.33:38337]     migrate_upgrade()
[Fri May 05 11:35:21.289436 2017] [wsgi:error] [pid 3973:tid 140195138877184] [remote 14.17.22.33:38337]   File "/usr/local/lib/python2.7/dist-packages/flask_migrate/__init__.py", line 244, in upgrade
[Fri May 05 11:35:21.289529 2017] [wsgi:error] [pid 3973:tid 140195138877184] [remote 14.17.22.33:38337]     command.upgrade(config, revision, sql=sql, tag=tag)
[...]

PR #1030, opened 2019-06-20
Fixed bug that caused apache2+wsgi deployment to break

CunninghamPatrick:
stamp() call missing the migrations directory causes apache2 w/ mod_wsgi support to break.

Issues #257 seems to be directly related to this issue, but was never fixed because the user moved onto another deployment method.

### DIFF
--- a/CTFd/__init__.py
+++ b/CTFd/__init__.py
@@ -151,7 +151,11 @@ def create_app(config="CTFd.config.Config"):
             db.create_all()
             # stamp call needs the migrations directory passed to it.
             migrations_dir = utils.__file__.split("CTFd")[0] + "CTFd/migrations"
-            stamp(directory=migrations_dir)
+            if os.path.isdir(migrations_dir):
+                stamp(directory=migrations_dir)
+            else:
+                stamp()
+                
         else:
             # This creates tables instead of db.create_all()
             # Allows migrations to happen properly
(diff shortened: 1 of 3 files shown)


ColdHeat:
@CunninghamPatrick thanks for doing the investigation into this! 

On one hand,  you probably shouldn't be using sqlite in production (what this PR is implying) but I can recognize this as an issue overall. That being said, you should probably switch to MySQL before doing much more with CTFd in production.

CTFd/__init__.py
```diff
@@ -149,7 +149,13 @@ def create_app(config="CTFd.config.Config"):
         # Alembic sqlite support is lacking so we should just create_all anyway
         if url.drivername.startswith("sqlite"):
             db.create_all()
-            stamp()
+            # stamp call needs the migrations directory passed to it.
+            migrations_dir = utils.__file__.split("CTFd")[0] + "CTFd/migrations"
+            if os.path.isdir(migrations_dir):
```
ColdHeat: I don't think you actually need this if statement. 

The definition for `migrations_dir` is quite ugly. I think you could probably cut it out entirely and just do something with stamp() and `app.root_path`. Also you should use `os.path.join`.

ColdHeat:
It'd be great if you could resolve the concerns I have above and also modify the other `stamp()` call that happens in the exports code. https://github.com/CTFd/CTFd/blob/master/CTFd/utils/exports/__init__.py#L281

CTFd/__init__.py
CunninghamPatrick: I agree it was quite ugly, at work I'm unable to push to github because of the corporate network, so I was editing directly in the GitHub editor. I'll look into those other options today.

Issue #973, opened 2019-04-30
Standardize to one database type


ColdHeat:
Supporting multiple database types is becoming more trouble than it's worth. It's time to choose what is the completely official supported database. 

An important note is that MariaDB's JSON support is lacking and it technically treats them as strings. 
https://mariadb.com/kb/en/library/json-data-type/

It's unlikely that we should ever remove SQLite support given its usefulness in tests.

ColdHeat:
Some consideration should be given to removing the dataset and datafreeze dependencies in favor of custom written solutions.

ColdHeat:
After a lot of deliberation, I am not sure this is still worth it. The issue is that MySQL and MariaDB should be very close to each other and looking at other applications, they also tend to support both. 

Postgres also has spiked in popularity and some graphs show it having more popularity than MySQL (not sure how true this is). There's also a few projects that are Postgres only and they could one day be used. 

The issue is that outright deprecating Postgres will lock out a bunch of people/projects/code when in actuality the most weird janky things happened b/c of MySQL. Postgres warts are really only in the import/export code. 

This doesn't really address the point that testing on all of these databases costs time. Instead I think Postgres will be spun out into a separate status check that doesn't need to pass and documentation will clarify the position that Postgres is tested against but not officially supported and not used by any CTFd developer. If you use CTFd with Postgres, you are on your own.

ColdHeat:
Closed in #1494

PR #1046, opened 2019-06-29
Fix several issues in Vagrant provisioning

joshuadugie:
1. Fixed various error output from `dpkg-preconfigure` due to a noninteractive shell:
```
default: dpkg-preconfigure: unable to re-open stdin: No such file or directory
```
The fix is by providing `DEBIAN_FRONTEND=noninteractive` to `apt-get install` commands.

2. Install the latest MariaDB server from their own repo since Ubuntu 16.04 only has [version 10.0](https://packages.ubuntu.com/search?keywords=mariadb-server&searchon=names&suite=xenial&section=all), which does not support the JSON type, causing database population to fail on provisioning.

3. The `install_docker.sh` (renamed from `install_docker_ubuntu.sh`) script assumed the user "ubuntu", which isn't the case for a Vagrant-provisioned machine.  The error output for this was `usermod: user 'ubuntu' does not exist`.  This is fixed by running the script as a non-privileged user and using the `${USER}` environment variable.

4. `pip` was being unnecessarily invoked via `sudo` in the `install_docker.sh` script, causing the following error output (truncated):
```
The directory '/home/vagrant/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled...
The directory '/home/vagrant/.cache/pip' or its parent directory is not owned by the current user and caching wheels has been disabled...
```

5. The `install_docker.sh` script forgot a backslash line-continuation character on its `apt-get` command, causing the following error output:
```
/tmp/vagrant-shell: line 10: apt-transport-https: command not found
```

6. In order to run tests, one needs `nose` installed.  Added to `development.txt`.


ColdHeat:
Thanks for the PR! 

A few points:

- MariaDB's JSON type is sort of a lie. Do you know if the MySQL version in the repos for 16.04 supports JSON? It likely would be better long term to use that. See #973. 
- Tests are now run using `pytest`. See https://github.com/CTFd/CTFd/blob/master/Makefile#L6-L8

ColdHeat:
Actually, let me do some soul searching about what database makes the most sense. The only concrete answer I know is that it isn't Postgres. But it's tough to say whether MySQL or MariaDB will be  the more "defacto" database.

joshuadugie:
Removed nose from development.txt.  (I was going off https://github.com/CTFd/CTFd/wiki/Development; maybe update that?)  I rarely interact with databases and have seen, in passing, good reviews about PostgreSQL -- I'm curious what about it makes you certain it isn't the right solution for CTFd?

ColdHeat:
The wiki front page links to https://docs.ctfd.io/en/latest/ but that's a good point. I've updated it on the development page. 

Well often times a project/company picks a database and goes with it until there's a problem before switching. There should be a few good engineering blog posts about this. 

In this case, CTFd started out with MySQL and eventually aimed to support the three major databases (SQLite, MySQL, Postgres). The problem is that even with an ORM, standardizing between them all is a lot more headache than it's worth. Two bugs come to mind that had different behavior depending on database (submission time precision in MySQL and import/exports failing in Postgres). I think some bugs could have been avoided with Postgres but I also think MySQL is easier to work with and also likely helped avoid bugs that could have manifested in Postgres. 

At this point, Postgres support gets hacked on or worked around with the main concern being MySQL, MariaDB, and SQLite. Right now, Postgres just slows down tests and doesn't get much usage.

PR #1069, opened 2019-07-29
Add support for custom OAuth2 providers

ldruschk:
This pull requests allows to use CTFd with any OAuth2 provider. By default everything is still configured work with MLC. The custom provider can be configured through the admin interface or through environment variables.


ColdHeat:
This will be supported in a future version of CTFd. As this is implemented, it's far more complex than necessary.

PR #1070, opened 2019-07-29
(Optionally) require an auth key for registration

ldruschk:
This pull requests adds an "auth key" to the registration form which can be configured via the admin panel or environment variables. If an auth key is configured, it is required to register a new user account.

This allows to leave the registration open but still restrict who is able to register.


ColdHeat:
This is far better implemented as a plugin in my opinion. Directly in the core seems unnecessary.

PR #1083, opened 2019-08-12
[WIP] i18n for the core theme

frankli0324:
wants to solve #570


frankli0324:
better squash the commits for that XD

ColdHeat:
This is a good idea that I hadn't considered. However it'll be some time before it can be worked on entirely. Primarily the JS in CTFd needs a large rewrite which is being worked on locally.

Issue #1099, opened 2019-09-05
CTFd crash right after install


noraj:
## Git + python 3.7.2 (3b1b82b9a0fbcb8731d7a3a3bbac99499c466c99)

[![asciicast](https://asciinema.org/a/iHDMVNRWSYJDTZUgIy4wIBsOK.svg)](https://asciinema.org/a/iHDMVNRWSYJDTZUgIy4wIBsOK)

## Git + python 2.7.15 (3b1b82b9a0fbcb8731d7a3a3bbac99499c466c99)

[![asciicast](https://asciinema.org/a/JRqfe0rMl0QeZAEklyvV1LQX9.svg)](https://asciinema.org/a/JRqfe0rMl0QeZAEklyvV1LQX9)

## 2.1.4 + python 3.7.2

[![asciicast](https://asciinema.org/a/wErebw8ZN2HFER9P71tLW4FBv.svg)](https://asciinema.org/a/wErebw8ZN2HFER9P71tLW4FBv)

**Environment**:

  - CTFd Version/Commit: see titles
  - Operating System: ArchLinux
  - Web Browser and Version: Firefox

**What happened?**

CTFd crash right after install.

**What did you expect to happen?**

Working or at least a more comprehensible error.

**How to reproduce your issue**

See the asciinema, but basically

- Download CTFd
- Install dependencies
- run flask
- crash

PS : with the untouched default config

**Any associated stack traces or error logs**

See the asciinema

noraj:
In fact this is the just the README that is wrong

![image](https://user-images.githubusercontent.com/16578570/64371584-339df980-d021-11e9-8e2e-928ef7c31a5d.png)

One must not do `flask run` but rather as said on the wiki (https://github.com/CTFd/CTFd/wiki) `python serve.py`.

PR #1100, opened 2019-09-05
fix the run command

noraj:
see #1099


ColdHeat:
This isn't really the right solution since it'd be much better to continue to support `flask run`. This seems to be related to monkey patching which should be required for other worker setups but perhaps not for the flask dev server.

PR #1117, opened 2019-09-18
Improve notification datetime display

grant-h:
Bare ISO time is hard to read.


ColdHeat:
Good idea! Unfortunately this will need to target the `2.2.0-dev` branch where the moment calls are handled in the page entrypoints.

grant-h:
It looks like with the introduction of `data-time` from the new javascript rewrite, this patch isn't necessary any more?

ColdHeat:
Oh you're right. It's missing in the admin panel though.

ColdHeat:
Resolved in https://github.com/CTFd/CTFd/commit/870384f89fdc3fe4ef890a56571f4cbac587591e

PR #1122, opened 2019-09-28
Set up semaphore

itechyboy:


ColdHeat:
I don't think this is actually intended for upstream so I'm closing this.

PR #1123, opened 2019-09-29
Add challenge id in flag submission log

Hedroed:
## Use-case
I use a SIEM to collect CTFd logs and i want to make the top 10 challenges with the most incorrect submissions.

## Solution
Adding the challenge identifier to submission logs will allow to get the challenge linked to the submission.

### DIFF
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -382,6 +382,7 @@ class ChallengeAttempt(Resource):
 
         chal_class = get_chal_class(challenge.type)
 
+
         # Anti-bruteforce / submitting Flags too quickly
         kpm = current_user.get_wrong_submissions_per_minute(user.account_id)
         if kpm > 10:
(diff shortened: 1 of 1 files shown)


ColdHeat:
Thanks for the PR! This is a good idea, never realized that this information was missing. Two things:

- Can you rename `{challenge}` to `{challenge_id}`
- Can you use `challenge.id` instead of `challenge_id`

Hedroed:
I made the modifications for `{challenge_id}` but i can't use `challenge.id` because methods `solve()` and `fail()` close the database session.

CTFd/api/v1/challenges.py
```diff
@@ -382,6 +382,7 @@ def post(self):
 
         chal_class = get_chal_class(challenge.type)
 
+
```
ColdHeat: Looks like you're failing lint checks here.

PR #1124, opened 2019-09-30
Fix #1079 feature toast

### LINKED ISSUE #1079: Add a toast to ezq
The alert modal is a bit jarring. We should switch to a toast because it doesn't require a click and should expire on its own. 
https://getbootstrap.com/docs/4.3/components/toasts/

Shimosora:
We fixed #1079.
We added toast feature.
The toast will disappear after 1 second.


ColdHeat:
Thanks for the PR!

A few things:
- Can you not commit your generated JS files and your package and yarn files? The generated JS is hard to verify and I haven't got a solution for it just yet. 
- Can you move the toast to the bottom right? 
- Is it possible to pull this off without making changes to `base.html`?

ColdHeat:
Abandon in favor of #1135

PR #1140, opened 2019-10-19
Dockerfile save 100MB image and clean-up

sir-ragna:
I replaced `COPY . /opt/CTFd` because it potentially includes a lot of things you do not want in the image (such as the .git repository and your local editor preferences.) This is probably better then attempting to keep `.dockerignore` up-to-date.

I completely removed the following command because there are no `requirements.txt` files in those directories anymore.
```
RUN for d in CTFd/plugins/*; do \
      if [ -f "$d/requirements.txt" ]; then \
        pip install -r $d/requirements.txt; \
      fi; \
    done;
```

I switched to the basic `alpine` image and use `apk add python3` instead of using the `python:3.7-alpine` image and this saves about a 100MB. On my system I built the following containers with `buildah`.
- 587 MB, 2.2.0-dev branch with python:3.7-alpine
- 473 MB, 2.2.0-dev branch with alpine and `apk add python3`
- 564 MB, master branch with python:3.7-alpine
- 457 MB, master branch with alpine and `apk add python3`
- 532 MB, docker.io/ctfd/ctfd:latest

I added some environment variables that are used for configuration in `docker-entrypoint.sh` to the Dockerfile. This should help people find those configuration options.

In `docker-entry.sh` I also added a password on the `mysqladmin ping` command. It didn't work on my system otherwise. (I was testing with the `mariadb:latest` container.)

------

I'm using this to run a pod with podman. I haven't tested this yet with docker-compose.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,29 @@
-FROM python:3.7-alpine
-RUN apk update && \
-    apk add python python-dev linux-headers libffi-dev gcc make musl-dev py-pip mysql-client git openssl-dev
-RUN adduser -D -u 1001 -s /bin/bash ctfd
+FROM alpine
+
+RUN apk update && apk add python3 python3-dev linux-headers libffi-dev gcc \
+    make musl-dev mysql-client git openssl-dev --no-cache
+
+RUN adduser -D -s /bin/bash ctfd ctfd
 
 WORKDIR /opt/CTFd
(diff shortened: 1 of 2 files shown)


ColdHeat:
A lot of the things you removed were chosen for specific reasons that I agreed with when they were suggested. Maybe there's some gain by switching to the base alpine image but I don't really see 100 MB as game changing. I would rather just go with the python base image.

Issue #1116, opened 2019-09-18
Add support for displaying country flags to CTFd


grant-h:
Seeing the name of countries is nice, but its easier to recognize flags at a glance, plus they hold a lot of pride for many participants. Country code to flag libraries are available for free (MIT licensed) at https://flag-sprites.com/

![image](https://user-images.githubusercontent.com/3030665/65165102-1814f480-da0c-11e9-996b-0185ffd56230.png)

If CTFd is interested, adding support for these flags is very easy. Something along the lines of:

```diff
--- a/CTFd/themes/core/templates/base.html
+++ b/CTFd/themes/core/templates/base.html
@@ -10,6 +10,7 @@
 	<link rel="stylesheet" href="{{ url_for('views.themes', path='css/vendor/font-awesome/fontawesome-fonts.css') }}" type='text/css'>
 	<link rel="stylesheet" href="{{ url_for('views.themes', path='css/vendor/font-awesome/fontawesome-all.min.css') }}" type='text/css'>
 	<link rel="stylesheet" href="{{ url_for('views.themes', path='css/vendor/font.css') }}"  type='text/css'>
+	<link rel="stylesheet" href="{{ url_for('views.themes', path='css/flags.min.css') }}"  type='text/css'>
 	<link rel="stylesheet" href="{{ url_for('views.themes', path='css/jumbotron.css') }}">
 	<link rel="stylesheet" href="{{ url_for('views.themes', path='css/sticky-footer.css') }}">
--- a/CTFd/themes/core/templates/users/users.html
+++ b/CTFd/themes/core/templates/users/users.html
@@ -60,7 +60,11 @@
 								{% endif %}
 							</td>
 							<td class="d-none d-md-table-cell d-lg-table-cell">
-								<span>{% if user.country %}{{ lookup_country_code(user.country) }}{% endif %}</span>
+								<div>{% if user.country %}<img src="{{ url_for('views.themes', path='img/blank.gif') }}"
+									class="flag flag-{{user.country.lower()}}"
+									alt="{{ lookup_country_code(user.country) }}"
+									style="vertical-align: top;" />
+									<span>&nbsp;&nbsp;{{ lookup_country_code(user.country) }}</span>{% endif %}</div>
 							</td>
 						</tr>
 					{% endfor %}
```

ColdHeat:
[...]

PR #1148, opened 2019-10-22
Country flag Support added

jimender2:
I added Support for Country flags to be added when the users country is set as described in #1116 .  I needed to modify the code they gave a little bit to work with teams and users.

I added the CSS file into a vendor folder named flag-sprites due to it coming from an outside source.  They say that "FamFamFam flag icons are "available for free use for any purpose with no requirement for attribution""

@ColdHeat If I need to do anything else, please let me know.


ColdHeat:
This needs to target the `2.2.0-dev` branch and use the updated front end structure there as well.

ColdHeat:
Implemented this for 2.2.0 in #1152

PR #1155, opened 2019-10-31
Strip password before length check

dwillie:
Prevents users from submitting all whitespace for their password and subsequently having an empty password.


dwillie:
Not sure why this is breaking in CI, looks like there was a problem with installing dependencies

ColdHeat:
Thanks for the PR!

PR #1165, opened 2019-11-11
Add show argument to /api/v1/challenges

jedevc:
The show argument can be set to "all" to display all challenges,
including hidden and locked challenges and ones for which their
dependencies have not been completed. This functionality is only
available to admins.

I've added this because I've been working a lot with the CTFd API recently, and we need the ability to list all the challenges, including ones that may have dependencies. As it was before, hidden challenges and ones with unmet dependencies were not shown, even to admin accounts.

### DIFF
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -58,31 +58,38 @@ class ChallengeList(Resource):
         # This can return None (unauth) if visibility is set to public
         user = get_current_user()
 
-        challenges = (
-            Challenges.query.filter(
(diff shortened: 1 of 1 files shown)


jedevc:
Looks like the build has failed for unrelated reasons.

ColdHeat:
Now that I was able to get 2.2 out I can see more about how this is necessary. In general the API needs more filtering capabilities. I'm not sure if the show parameter makes the most sense but I'll mark this for future work. Most of the API improvements can be made pretty easily but it hasn't been a focus for awhile.

ColdHeat:
Thanks for the PR here, I implemented something very similar in #1451.

PR #1174, opened 2019-11-30
Only convert values in date columns to dates

argilo:
While importing a backup file, I encountered the following error:

> ValueError('unconverted data remains: Z',)

The error occurs in this code, which attempts to convert date values for use in SQLite:

https://github.com/CTFd/CTFd/blob/97f5dcaf8c408a68ef0062d410b6c1599976688f/CTFd/utils/exports/__init__.py#L210-L230

While importing `submissions.json`, this code tries and fails to convert a flag submission which by coincidence happens to be in ISO8601 format:

`{"id": 4585, "challenge_id": 41, "user_id": 24, "team_id": 1, "ip": "", "provided": "2019-10-26T17:49:32Z", "type": "incorrect", "date": "2019-11-29T14:07:41.095729"}`

Clearly flag submissions should not be treated as dates.

To fix this problem, I've added in a check to ensure that only date fields (`created` and `date`) are converted.


ColdHeat:
Does this behave properly under MySQL? I don't really want to add more hacks into an existing hack because of some edge case.

argilo:
I would presume so, because this code is inside an `if sqlite:` block.

This change isn't so much adding a hack as constraining an existing hack so that it doesn't touch things it's not meant to.

ColdHeat:
I'm abandoning this for #1270 which checks the column to see if it's a DateTime column

Issue #1043, opened 2019-06-26
Score of a challenge should be recalculated whenever scoring parameters are changed.


pedromigueladao:
**Environment**:

  - CTFd Version/Commit: 2.1.3
  - Operating System: Linux
  - Web Browser and Version: any

**What happened?**
When the decay limit is updated, current value should be updated to accommodate the new values.

**How to reproduce your issue**
1) Create a challenge
2) submit a correct answer
3) update the decay limit (similarly for max and min values)

ColdHeat:
Can you clarify? The code already handles updating of the challenge value if the decay changes. 
https://github.com/CTFd/CTFd/blob/master/CTFd/plugins/dynamic_challenges/__init__.py#L121-L133

Do you mean the scoreboard value should change or the challenge value should change?

ColdHeat:
I'm closing this because I need a little more information here. Feel free to update with more information and re-open in the future.

ColdHeat:
This should now be fixed in the 2.2.0-dev branch which is due out soon.

pedromigueladao:
Hi @ColdHeat thank you. I missed this one. Just to add some context the issue is: suppose challenge X is worth 500, and users A, B, and C solve it.
If you then update the decay limit, the scoring of users A, B and C are not updated accordingly. Only whenever a new used D solves the challenge. I hope it is clearer now.

PR #1177, opened 2019-12-07
Fix: dynamic challenge points decrease when hidden user submit first ly

fengkx:
# What this pr do
When I use CTFd, I find that when a hidden user solve a dynamic challenge first(Before any others), It will cause the points of this challenge to change.

# Detail of what happend

We should look at some lines of code

https://github.com/CTFd/CTFd/blob/master/CTFd/plugins/dynamic_challenges/__init__.py#L203-L222

Logic go through:
L203-L211: Get the solve_count with `hidden = False` and `banned = False` Because the hidden user is the first person solve the challenge and he is hidden, the solve_count should be 0.

L214: minus solve_count by 1. solve_count become -1. Which cause `((chal.minimum - chal.initial) / (chal.decay ** 2)) * (solve_count ** 2)` to be no zero. And change the value.

# The way to reproduce

In most of the case we don't notice the problem. Because when we hold a race and have a dynamic challenge the `((chal.minimum - chal.initial) / (chal.decay ** 2))` should be small than one at most of the cases. And it got `math.ceil` in L222

1. Create a dynamic challenge with a **Large** initial and a **Small** minimum and a **Small** decay
such as `10000` `10` `3`

2. Create a hidden user and submit the flag.

3. We can find the points changed.

Bug

# Way to fix

Add a condition in L214

### DIFF
--- a/CTFd/plugins/dynamic_challenges/__init__.py
+++ b/CTFd/plugins/dynamic_challenges/__init__.py
@@ -211,7 +211,7 @@ class DynamicValueChallenge(BaseChallenge):
         )
 
         # We subtract -1 to allow the first solver to get max point value
-        solve_count -= 1
+        solve_count -= 1 if solve_count != 0 else solve_count
 
         # It is important that this calculation takes into account floats.
         # Hence this file uses from __future__ import division


ColdHeat:
I have a suspicion this is related to #1043?

fengkx:
I don't understand what is going on in #1043. But it seem to be related to ##1166

ColdHeat:
I consolidated the PRs into one PR that should resolve all the known issues here.

Issue #820, opened 2019-01-04
Python 3 Only 


ColdHeat:
It's very likely that by the time CTFd 3.0 comes out we will have reached the Python 2 EOL or at least be very close. There won't be any great reasons to support Python 2 at that time in an open source capacity.

fengnian-xu:
现在支持的还是python2.7版本吧

ColdHeat:
Basically closed in #1454. To be released with v3.0.

PR #1214, opened 2020-01-12
Try to convert to python3(Python2 grandpa is retire)

99Kies:
**First of all, I think this project is a very good open source project, and our team is using this framework.But now that python2 is retired, I think it's a good project to continue with, and we think it's necessary to upgrade to python3 to facilitate later maintenance and future operations.**

**1. Only the.py file has been modified (it can be successfully built).
2. Successful build means (can register, can log in and can put questions), because I cannot fully test in a short time due to my limited ability alone, I need more friends to check the python3 code**


ColdHeat:
CTFd already supports Python 3 and Travis already tests for it. See #820

PR #1218, opened 2020-01-20
Strip spaces on registration and refine password reset

ColdHeat:
* Usernames are now properly stripped before being used in registration checks
* Reset password function now uses email addresses instead of user names for tokens
* Prevent MLC users from resetting their password


stefan2904:
This is now [CVE-2020-7245](https://nvd.nist.gov/vuln/detail/CVE-2020-7245)

fridayraymond:
Sup guys

PR #1223, opened 2020-01-28
fix spelling errors

hjung4:


ColdHeat:
In general cosmetic changes like these aren't worthwhile. https://github.com/CTFd/CTFd/blob/master/CONTRIBUTING.md#did-you-fix-whitespace-format-code-or-make-a-purely-cosmetic-patch

None of this is particularly user-visible. If you could perhaps create systems or lints that could do spellchecking at build time as part of the linters I think it would be more useful.

PR #1237, opened 2020-02-15
Update to fix #1002

### LINKED ISSUE #1002: Make e-mail text customizable(to improve spam rating)
I noticed that the built-in e-mail text are producing a spam warning. Two provider moved them straight to the spam folder.

https://www.mail-tester.com complained about:
- not enough text
- wrong encoding for special characters I'm using([admindōjō](https://admindojo.org/))[means I cant use `ctf_name`]

It would be nice to be able to be able to customize subject and text for all e-mails without the need out touching CTFd directly.
Additionally it would be nice to be able to use all defined variables in all three mail types.

**Proposal**

- make email text and subjects customizable via `config.py`
- add `ctf_name=get_config('ctf_name')` to all three
- add `name=name` to all three to greet the user with username
- use variable names instead of `{0}`,`{1}` in `def forgot_password(email, team_name)`

jimender2:
Add in feature to change emails that are sent to the user.

### DIFF
--- a/CTFd/themes/admin/templates/configs/email.html
+++ b/CTFd/themes/admin/templates/configs/email.html
@@ -8,7 +8,11 @@
 			<li class="nav-item">
 				<a class="nav-link" href="#mailgun" role="tab" data-toggle="tab">Mailgun</a>
 			</li>
+			<li class="nav-item">
+				<a class="nav-link" href="#emailtext" role="tab" data-toggle="tab">Email Text</a>
+			</li>
 		</ul>
+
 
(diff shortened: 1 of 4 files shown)


ColdHeat:
You're missing one of the other emails that's available in the email module. 

Also looks like you're sending the wrong default subject for one of the emails.

ColdHeat:
Actually sorry I think this is a feature that it's better for me to implement. This needs to handle adding in migrations to add in the new default config values.

PR #1260, opened 2020-02-24
Fix issue where password contains % symbol

novafacing:
With unescaped % symbols in an SQL password like with the following URI:

mysql+pymysql://root:p@ssw%rd@127.0.0.1/ctfd

Alembic will throw an error complaining about the %, as it is a valid control character in an SQL URI. This replace escapes the URI % symbol so that doesn’t happen.


novafacing:
I should clarify that yes I tested this. Fixed my issue immediately.

ColdHeat:
The way I see this, this should be resolved upstream in Flask-Migrate or alembic. I don't think it makes a lot of sense inside of CTFd. Maybe you can find a different version of something where this resolved that we could pin to?

novafacing:
Sure, I can ask the alembic guy if he'd like to have a fix for that pulled in, I have a pretty good idea of what needs to change.

Issue #1274, opened 2020-03-03
Investigate flask-restx


ColdHeat:
Flask-Restplus isn't really maintained so it might be worth investigating popular forks: https://github.com/python-restx/flask-restx.

ColdHeat:
The Flask-Restplus README outlines what happened https://github.com/noirbizarre/flask-restplus/commit/e9d5632b3e62b49a7fd2e57fb8606a6db229a6b6. I skimmed the issue and it's fairly reasonable so it makes sense to switch over sooner rather than later.

ColdHeat:
Closed in #1282

PR #1282, opened 2020-03-07
Replace flask_restful with flask_restx

cydave:
This PR contains all the changes required to migrate from [flask-restplus](https://github.com/noirbizarre/flask-restplus) to [flask-restx](https://github.com/python-restx/flask-restx).
See issue: #1274


cydave:
ah, looks like the requirements.txt is not up to date in this current state.
Maybe using the original one would pass the checks?

ColdHeat:
Thanks for writing this. Looks like it works but I haven't really investigated `flask-restx` and what benefits it might have. Do you happen to know how it improves on `Flask-Restplus`?

cydave:
No I don't. I think the main motivation behind the fork is to simply get more changes in more frequently as the original author didn't keep up with requests.

PR #1286, opened 2020-03-10
Prune down docker image

### LINKED ISSUE #1285: Dockerfile problems
The `Dockerfile` has a handful of improvements available to it.  Mainly...

##### `COPY . /opt/CTFd` is overkill, even with existing `.dockerignore`

Specifically, by copying the `.git` directory, your Docker image is guaranteed to break Docker's layer caching on *every commit*.  Every layer after that also cannot be matched in cache.  

No software should require its version control history to run.  So `.git` should be excluded.

Also consider this sequence:
```Dockerfile
WORKDIR /opt/CTFd
...
COPY requirements.txt .
RUN pip install -r requirements.txt
COPY . /opt/CTFd
```
This copies the `requirements.txt` file twice.  

From the inside of `ctfd/ctfd` currently, here's files that are not required for the image to run:
```
.codecov.yml
.dockerignore
.git
.github
.gitignore
.prettierignore
.travis.yml
CHANGELOG.md
CONTRIBUTING.md
Dockerfile
README.md
Vagrantfile
development.txt
docker-compose.yml
docs 
requirements.txt # delete it after install
setup.cfg
tests
```

There's probably more, but I would propose to exclude them all.

atz:
Avoid adding things the runtime doesn't need and use fewer commands to produce fewer layers.

Note: `/bin/bash` is NOT installed on `python:3.7-alpine`, so it does not make sense as a default shell for the ctfd user.

Fixes #1285


Dockerfile
```diff
@@ -1,26 +1,20 @@
...
+    apk add python python-dev linux-headers libffi-dev gcc make musl-dev py-pip mysql-client git openssl-dev
 
 COPY . /opt/CTFd
 
-RUN for d in CTFd/plugins/*; do \
-      if [ -f "$d/requirements.txt" ]; then \
-        pip install -r $d/requirements.txt; \
-      fi; \
+RUN pip install -r requirements.txt
+RUN for d in CTFd/plugins/*/requirements.txt; do \
```
ColdHeat: This isn't as readable as it was before. Please revert this chunk.

Dockerfile
```diff
@@ -1,26 +1,20 @@
...
-RUN adduser -D -u 1001 -s /bin/bash ctfd
-
 WORKDIR /opt/CTFd
 RUN mkdir -p /opt/CTFd /var/log/CTFd /var/uploads
 
-COPY requirements.txt .
-
-RUN pip install -r requirements.txt
+RUN apk update && \
+    apk add python python-dev linux-headers libffi-dev gcc make musl-dev py-pip mysql-client git openssl-dev
```
ColdHeat: Might as well break this into multiple lines as well.

.dockerignore
```diff
@@ -5,3 +5,19 @@ CTFd/*.db
...
+.prettierignore
+.travis.yml
+*.md
+Dockerfile
+Vagrantfile
+development.txt
+docker-compose.yml
+docs
+setup.cfg
+tests
```
ColdHeat: I don't really agree with the premise of the container should only contains what it needs to run. What if I develop CTFd using Docker? 

Will just putting `.git` in here prevent it from being copied during the COPY command?

ColdHeat:
Thanks for writing this. I didn't realize that `.git` was preventing the cache from working and I was investigating that myself awhile ago. 

I don't know if I completely agree with the idea that the container should only contains what is needed to run. The extra files don't hurt and it provides more benefit than not I would say. 

These changes look fairly sane so as long as this builds and works, should be okay to merge.

ColdHeat:
Hi @atz any concern with the comments I've raised?

PR #1295, opened 2020-03-23
fix scoreboard top 10 graph's title

restran:
scoreboard top 10 graph's title should be `Top 10 Users`  when user mode is not teams.


ColdHeat:
Can you rebase this against master and run `yarn build`?

PR #1297, opened 2020-03-25
Update to Python v3.8

ckane:
Python version 3.8 is the most current stable version of Python, and I was able to successfully build and start a Docker image using this version of Python. Will continue to test on 3.8 to validate that it functions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.7-alpine
+FROM python:3.8-alpine
 WORKDIR /opt/CTFd
 RUN mkdir -p /opt/CTFd /var/log/CTFd /var/uploads
 


ColdHeat:
Thanks for filing this. Version changes are a little tricky though. They can introduce weird breakages that aren't noticed for awhile. Unless there's a significant benefit from moving to 3.8 I think it's better to hold this off until something like v2.4 or v3.0 (when Python 2 is deprecated).

ckane:
That sounds fine - I figured if I dropped a PR here, at least there's be the TravisCI data point, and I'll continue testing it on the newer Python version in my own deployment, and report back with how that works

xavierholt:
I was able to run with Python 3.8, but needed to update FreezeGun to the latest version (`freezegun==0.3.15`) to run all the unit tests: the old version relied on the deprecated `time.clock()` function that was [removed](https://bugs.python.org/issue36895) in 3.8.

ckane:
Oh thanks! It doesn't look like the existing repo imports that directly, so for a new build I presume that will get added as the latest version automatically?

xavierholt:
@ckane - It's referenced in `development.txt` (like `requirements.txt` but for the dev dependencies).  I was able to bump the version without anything else needing to be updated, so I think it'll just be a one-line (one character!) change.

ckane:
Ah thanks - I see it now. I searched my fork of the repo not realizing that it won't find matches in a forked repo. I'll add it to my branch for this PR.

ColdHeat:
I looked through Python 3.8 and the upcoming 3.9 and I don't see anything that would benefit from a version bump. I think it's better to stay at 3.6 compatibility.

ckane:
Ok thanks, that's what I've done in the meantime (just forked the project and made the change)

PR #1299, opened 2020-03-29
bugfix: affix @authed_only to /notifications

0x48piraj:
I am a student using CTFd to create a small CTF for my university, I found the platform very efficient and easy to setup. CTFd is simply amazing!

While exploring the codebase, I found that anonymous users can also access notifications which, in most cases, is not what organizers want.

Thus this pull simply adds `@authed_only` decorator to `/notifications` route to make it view-able by authenticated users only.

### DIFF
--- a/CTFd/views.py
+++ b/CTFd/views.py
@@ -266,6 +266,7 @@ def integrations():
 
 
 @views.route("/notifications", methods=["GET"])
+@authed_only
 def notifications():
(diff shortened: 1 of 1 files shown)


ColdHeat:
This is intentional. At this time, notifications are intended to be global announcements to all. Perhaps in later versions it will be revisited.

PR #1300, opened 2020-04-01
Add CTFtime.org as OAuth provider

kyprizel:
First of all - thanks for this amazing CTF board engine!
I'm pretty sure it made CTF running much easier for a lot of people.
And I understand that MajorLeagueCyber was part of the project, but still - people use CTFtime.org and they have to create and manage teams in multiple places.
So I decided to make their life a bit easier and run OAuth engine at CTFtime and added some hacks to integrate it to CTFd. Not sure if I used correct way, but I was not able to do it only with currect plugin system.

Thanks.

### DIFF
--- a/CTFd/auth.py
+++ b/CTFd/auth.py
@@ -11,7 +11,7 @@ from CTFd.utils import config, email, get_app_config, get_config
 from CTFd.utils import user as current_user
 from CTFd.utils import validators
 from CTFd.utils.config import is_teams_mode
-from CTFd.utils.config.integrations import mlc_registration
+from CTFd.utils.config.integrations import oauth_registration
 from CTFd.utils.config.visibility import registration_visible
 from CTFd.utils.crypto import verify_password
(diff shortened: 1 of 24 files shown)


ColdHeat:
I don't see this getting merged to master. MLC is a priority as far as CTFd is concerned regardless of usage. The scoreboard formats already align and all integrations are already there. Integration with CTFtime would already be possible via a plugin and is outlined [in the docs](https://docs.ctfd.io/en/latest/scoring.html#ctftime) with regard to scoring.

kyprizel:
@ColdHeat ColdHeat the problem is - teams do not use MLC and register aliases for an event.
Then aliases got imported to CTFtime and results of existing teams should be merged.
This integration allows to logon with CTFtime if event organizers decides to use CTFtime as oauth provider. MLC is by default anyway.

irdkwmnsb:
*ahem* 
image

PR #1308, opened 2020-04-10
Re-enable the Jinja LRU Cache

### LINKED ISSUE #662: Re-enable jinja cache
```
    @app.before_request
    def disable_jinja_cache():
        app.jinja_env.cache = {}
```

Explore how we can re-enable the Jinja cache by perhaps storing it on Redis. This may require very custom code.

ColdHeat:
Re-enable the Jinja LRU Cache by overriding the `Environment._load_template` function and adding a theme namespace. 

This is not a great solution but it's likely that this is causing a lot of slowdown because we're continually reparsing and compiling templates. 

We still cannot completely clear the jinja cache but this way theme changes don't trample each other. 

Closes #662

### DIFF
--- a/CTFd/__init__.py
+++ b/CTFd/__init__.py
@@ -1,6 +1,7 @@
 import datetime
 import os
 import sys
+import weakref
 from distutils.version import StrictVersion
 
 from flask import Flask, Request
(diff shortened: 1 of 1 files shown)


ColdHeat:
```
from CTFd import create_app
from flask import render_template
from CTFd.utils.scores import get_standings
from CTFd.utils import config
import time

app = create_app()
app.config['SERVER_NAME'] = 'localhost'

with app.app_context(), app.test_request_context():
    standings = get_standings()
    start = time.time()
    for _ in range(1000):
        render_template(
            "scoreboard.html",
            standings=standings,
            score_frozen=config.is_scoreboard_frozen(),
        )
    end = time.time()
    print(end - start)
```

Running three times on each branch on my rinky dink laptop from 2013:
```
python testtime.py; python testtime.py; python testtime.py
master: 
93.42874217033386
82.99229288101196
72.27301716804504

branch:
18.520366668701172
20.439661979675293
18.175461292266846
```

Issue #1215, opened 2020-01-13
Switch from Alpine to python:3.7-slim-buster


ColdHeat:
https://pythonspeed.com/articles/base-image-python-docker-images/

Alpine is good for making a small image but bad for overall development flow because you need to rebuild Python wheels. https://github.com/docker-library/docs/issues/904

This wastes a lot of time and CPU during the build process. 

We should switch over to `python:3.7-slim-buster` as recommended or figure out some other way to make builds go faster. 

Another point it appears alpine is also slower: https://github.com/tao12345666333/docker-python-perf

ColdHeat:
Closed in #1461

PR #1312, opened 2020-04-11
Questions about optimizing Dockerfile play

99Kies:
## optimize Dockerfile, make the container lighter, mainly by removing PIP and apk download file cache to achieve a goal

**Environment**:

  - CTFd Version/Commit:
  - Operating System:
  - Web Browser and Version:

**What happened?**

Let the container lighter!

-----------------
I feel sorry about my English

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,33 @@
 FROM python:3.7-alpine
-RUN apk update && \
-    apk add --no-cache python python-dev linux-headers libffi-dev gcc make musl-dev py-pip mysql-client git openssl-dev
-RUN adduser -D -u 1001 -s /bin/bash ctfd
-
 WORKDIR /opt/CTFd
 RUN mkdir -p /opt/CTFd /var/log/CTFd /var/uploads
 
-COPY requirements.txt .
-
-RUN pip install -r requirements.txt --no-cache-dir
+RUN apk update && \
+    apk add --no-cache \
+        python \
+        python-dev \
+        linux-headers \
+        libffi-dev \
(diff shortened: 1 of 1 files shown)


ColdHeat:
Looks like a reasonable change. I'll need to do a quick comparison to see exactly what the size benefits are. I'll also have to carry this forward when we move to debian.

99Kies:
> Looks like a reasonable change. I'll need to do a quick comparison to see exactly what the size benefits are. I'll also have to carry this forward when we move to debian.

I think the alpine than debian/ubuntu and lighter, more suitable for the construction of the python web services, what do you think?

99Kies:
I often use alpine as the flask deployed docker image, hahaha

ColdHeat:
This is summarized here: #1215 

Alpine is probably smaller but it leads to other odd issues like random performance losses. You also can't download wheels from pypi which makes builds slow and hard to cache. In general it just sucks.

99Kies:
Sorry, I only focus on the container size, I got it. So the future plan to put the docker image replacement for debian?

ColdHeat:
Yes the future Docker base image will be Debian. It would've been done sooner but making a big change like that needs to be done in a major release. People can still use a custom Debian base image if they want the benefits though. 

Regarding this PR it looks like there's a very slight size savings:
```
ctfd/ctfd                                                   latest                         ad31929b2012        56 minutes ago      532MB
ctfd-branch                                                 latest                         367c115bb36c        5 minutes ago       497MB
```

Seems worth it overall though.

PR #1333, opened 2020-04-18
Add webargs to requirements and improve SubmissionsList to use webargs and mass deletions

ColdHeat:
* Add `webargs==6.0.0` module
* Improve `SubmissionsList` to paginate responses
* Improve `SubmissionsList` to support bulk delete
* Specify GET args to `SubmissionsList`


ColdHeat:
Don't like this approach as much anymore. Not worth the significant investment in rewriting the APIs for the 2.4 release.

Issue #1342, opened 2020-04-25
Rewrite populate.py to use data from Faker


ColdHeat:
I copied the populate.py data from lots of random lorem ipsum sites but instead we can probably just use faker. 

https://faker.readthedocs.io/en/master/

ColdHeat:
Closed in #1344

PR #1344, opened 2020-04-25
feat: add Faker and change populate.py

developerfred:
#1342

https://faker.readthedocs.io/en/master/fakerclass.html#examples


ColdHeat:
Wow fast PR! Can you move the Faker dep to `development.txt` instead of `requirements.txt` though?

ColdHeat:
Also looks like the last version to support Py2 was 3.0.1 so please pin that instead since Py2 is still supported by CTFd. 
https://github.com/joke2k/faker/blob/master/CHANGELOG.rst#301---14-january-2020

ColdHeat:
The emails you're generating appear to be invalid. 

image

populate.py
```diff
@@ -1177,35 +199,35 @@
 
 
 def gen_sentence():
-    return " ".join(random.sample(lorems, 50))
+    return fake.text()
 
 
 def gen_name():
-    return random.choice(names)
+    return fake.name()
```
ColdHeat, line 206: Can you make this return a first name only inline with previous behavior?

developerfred: yes

developerfred:
```
def gen_name():
    from random import shuffle, seed
    from faker.providers.person.en import Provider

    first_names = list(set(Provider.first_names))
    seed(1551)
    shuffle(first_names)
    
    return first_names[0:1000]
```

with this new script I can print a list of 1000 first names without repetition.

populate.py
```diff
@@ -1177,35 +199,42 @@
 
 
 def gen_sentence():
-    return " ".join(random.sample(lorems, 50))
+    return fake.text()
 
 
 def gen_name():
-    return random.choice(names)
+    from random import shuffle, seed
```
ColdHeat, line 206: This seems odd, do you need to do all of this for each function call? Is there not a direct method in Faker to generate a random first name?

requirements.txt
```diff
@@ -26,4 +26,4 @@ pathlib2==2.3.5
 flask-marshmallow==0.10.1
 marshmallow-sqlalchemy==0.17.0
 boto3==1.10.39
-marshmallow==2.20.2
+marshmallow==2.20.2
```
ColdHeat, line 29: You shouldn't have any changes in this file, can you revert this?

populate.py
developerfred: exists but can repeat itself at some point

requirements.txt
developerfred: yeap

developerfred:
@ColdHeat what command do i use to remove the change in the requirements of my pull request?

PR #1349, opened 2020-04-27
Remove datafreeze dependency

### LINKED ISSUE #1348: Remove datafreeze depenency
Datafreeze is archived https://github.com/pudo/datafreeze/ and was already a difficult library. This needs to be removed going forward.

ColdHeat:
* Remove `datafreeze` dependency
* Closes #1348


ColdHeat:
I think test coverage is okay to pass here given the amount of code that's being pruned out. This will need more of a manual test.

PR #1378, opened 2020-05-04
Add function mass clear user/team sessions

ColdHeat:
* Add function to mass clear the cached user/team attributes for all users/teams


ColdHeat:
Code doesn't have an accompanying test but just tested the functionality manually.

PR #1380, opened 2020-05-05
Add SQLALCHEMY_ENGINE_OPTIONS with a slightly higher default max_overflow

ColdHeat:
* Add `SQLALCHEMY_ENGINE_OPTIONS` to `config.py` to specify a slightly higher default `max_overflow`

### DIFF
--- a/CTFd/config.py
+++ b/CTFd/config.py
@@ -247,7 +247,7 @@ class Config(object):
     APPLICATION_ROOT = os.getenv("APPLICATION_ROOT") or "/"
     SERVER_SENT_EVENTS = not os.getenv("SERVER_SENT_EVENTS")  # Defaults True
     SQLALCHEMY_ENGINE_OPTIONS = {
-        "max_overflow": int(os.getenv("SQLALCHEMY_MAX_OVERFLOW", 20)),
+        "max_overflow": int(os.getenv("SQLALCHEMY_MAX_OVERFLOW", 20))
     }
 
     """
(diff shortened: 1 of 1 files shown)


banditaf:
The 'max_overflow` limit of `20` is still not sufficient in my opinion. This is because of the fact that the total number of concurrent sessions that can be initiated and maintained are the sum of `pool_size` and `max_overflow`. By default, both are `10`, which makes the total number of possible concurrent sessions to be `20` by default. However, after this commit, it was increased to `30` with `20` as `max_overflow`.

According to this source: https://docs.sqlalchemy.org/en/20/core/pooling.html
setting `max_overflow` to `-1` enforces no limit on the total number of possible concurrent sessions which should be the ideal case.

TLDR: `max_overflow` must be set to `-1` for no limits on concurrent sessions.

ColdHeat:
@banditaf You are able to configure `SQLALCHEMY_MAX_OVERFLOW` in config.ini. To have more connections, you can add more workers, or change the setting in `config.ini`/`config.py`. 

I did consider whether `-1` made more sense but based on my research it did not make sense as the default. There is still the max connection limit imposed by the database.  See something like https://stackoverflow.com/a/38534463. 

If there's a good rationale for setting the default to -1 we can consider changing it. 

Also see this quote from https://docs.sqlalchemy.org/en/20/errors.html:

> Given the above, the connection pool can be looked at as a safety valve for connection use, providing a critical layer of protection against a rogue application causing the entire database to become unavailable to all other applications. When receiving this error message, it is vastly preferable to repair the issue using up too many connections and/or configure the limits appropriately, rather than allowing for unlimited overflow which does not actually solve the underlying issue.

banditaf:
@ColdHeat Now that you have mentioned it and I gave it a thought, limiting the `max_overflow` size is indeed required. But that would vary from deployment to deployment. But I really believe that this should be part of the documentation so users can decide upon the best configuration according to their need.

PR #1384, opened 2020-05-06
admin theme successful confirmation email fix

ilanarbi:
unable to save changes in admin panel for Account Registration email


ColdHeat:
Thanks for fixing this!

PR #1390, opened 2020-05-07
Convert Dockerfile to Multi-Stage

jjasonkal:
- Decrease Dockefile image from 497MB to 391MB, leveraging Docker Multi-Stage technique.


ColdHeat:
I don't really think the size savings here are worth the added complexity. Also the Dockerfile is due for changes later and I don't really want to have to port this later.

PR #1393, opened 2020-05-10
Add Dropbox uploader support and fix file deletion issue

erseco:
Add Dropbox uploader backend support to store files using Dropbox API. 

To use it you need to generate a
_oauth2 token_ following the [Dropbox API tutorial](https://www.dropbox.com/developers/documentation/python#tutorial), set it to the `DROPBOX_OAUTH2_TOKEN`  variable and set `UPLOAD_PROVIDER`  to `dropbox`

You can specify the `DROPBOX_ROOT_PATH`, by default will be /CTFd

Also fixed an issue in CTFd API: CTFd/api/v1/files.py was not deleting files.

### DIFF
--- a/CTFd/api/v1/files.py
+++ b/CTFd/api/v1/files.py
@@ -61,6 +61,7 @@ class FilesDetail(Resource):
     def delete(self, file_id):
         f = Files.query.filter_by(id=file_id).first_or_404()
 
+        uploads.delete_file(file_id=f.id)
         db.session.delete(f)
         db.session.commit()
         db.session.close()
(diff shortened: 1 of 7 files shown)


ColdHeat:
Thanks for flagging the file deletion issue!

As for Dropbox based uploads, I don't really want to support that use case in master. That seems very far away from the way most web applications operate. I think you can still support that behavior with a plugin if you really need it though.

erseco:
Agree with the non essential use-case, I can split the PR if you want

ColdHeat:
Splitting out the file deletion changes would be great thanks!

erseco:
Split in #1396, closing this PR

Issue #1310, opened 2020-04-10
Move testing to Github Actions if possible


ColdHeat:
Not a fan of Travis and I hope that there's tighter integration with Github Actions.

ColdHeat:
Closed with #1394. Travis and Python 2.7 have been made an optional status check.

PR #1394, opened 2020-05-10
Added github actions workflow

erseco:
Added first GitHub Actions pipeline to solve #1310 and #1321


ColdHeat:
Thanks for the PR! I had some trouble figuring out the right format for Github Actions so this is really welcome!

PR #1396, opened 2020-05-11
Fix file deletion. Related to #1393

erseco:
Fixed an issue in CTFd API: CTFd/api/v1/files.py was not deleting files.


ColdHeat:
Perhaps you could also update this test? https://github.com/CTFd/CTFd/blob/22c132358ef15e89b93af84326962ab1c53cb029/tests/api/v1/test_files.py#L99-L119

ColdHeat:
Thanks for updating that test! I think a cleaner solution would be to update `gen_file` to have a parameter like `on_disk=False` or something like that which would automatically create the files that it generates. That way this behavior can be copied to other tests that might need it.

PR #1399, opened 2020-05-12
Fix bug which causes user creation email to never be sent out

### LINKED ISSUE #1398: Wrong method is used to check for request data in user creation API endpoint
CTFd version:  2.4.2

The following line uses the wrong method on the `flask.Request` instance, causing user creation e-mails to never be sent out:
https://github.com/CTFd/CTFd/blob/1a57a33fe712d3b449b9075b7587012f56ceb38f/CTFd/api/v1/users.py#L59

From the [flask documentation](https://flask.palletsprojects.com/en/1.1.x/api/#flask.request) about `request`: 
```
 property args

    The parsed URL parameters (the part in the URL after the question mark).
```

This causes `request.args` to always be an empty `ImmutableMultiDict` here which means the `if` statement will never be true, and thus the creation e-mail will never be sent either.

Instead we can use the `req` variable to access `notify` just like you did inside the statement to access the `password`:

`if req.get("notify"):`

jbrinksma:
This makes sure that e-mails will be sent when the user ticks the `Email account credentials to user` box on user creation.

The request data should be accessed through `req` instead of `request.args`, since `request.args` will always return an empty `ImmutableMultiDict` here.

Please see the issue attatched below for more details.

Fix #1398

### DIFF
--- a/CTFd/api/v1/users.py
+++ b/CTFd/api/v1/users.py
@@ -56,7 +56,7 @@ class UserList(Resource):
         db.session.add(response.data)
         db.session.commit()
 
-        if request.args.get("notify"):
+        if req.get("notify"):
             name = response.data.name
             email = response.data.email
             password = req.get("password")


ColdHeat:
Please see my comment in #1398

PR #1400, opened 2020-05-13
Have EventManagers yield before timer code to force SSE response head…

ColdHeat:
Some reverse proxies will buffer SSE responses until they see a valid SSE response header. Thus we need to force the EventManagers to yield a ping event immediatley to force Flask to set right headers. 

* EventManagers should send an initial ping event to force `text/event-steam` header to be set


ColdHeat:
This actually wasn't the problem but I still think it's valid based on what other applications have done.

PR #1397, opened 2020-05-11
Add support for limited privilege databases (ej. Heroku)

erseco:
I'm running CTFd on Heroku for testing purposes. Heroku provides a PostgreSQL database without permission to create/drop the database. Also, revoke privileges in `postgres` database, used by `sqlalchemy-utils` to check if the database exists. 

You can duplicate this permission behavior in a database running the next queries:

```
CREATE DATABASE ctfd;
CREATE USER ctfd WITH PASSWORD 'ctfd';
REVOKE CONNECT ON DATABASE postgres FROM PUBLIC;
REVOKE ALL PRIVILEGES ON DATABASE postgres from ctfd;
```
and using this url for database `DATABASE_URL=postgres://ctfd:ctfd@localhost/ctfd`

To fix this we can try-catch the comprobation of the database. Also fixed the import. Because we don't have drop support we just truncate all CTFd tables like in `reset` feature.

Added a basic Procfile and app.json for use with Heroku. We can easily deploy with the button `deploy to heroku` in Readme.

- Hosting CTFd on Heroku is intended for **testing** purposes.
- Heroku provides a free PostgreSQL database with a **limitation of 10K rows**.
- Heroku provides a free Redis database with a **limit of 25MB**.
- Heroku provides an ephemeral filesystem, so you **must** configure _s3_ as `UPLOAD_PROVIDER`

This PR refers to #1127 and #1327

PR #1403, opened 2020-05-13
Fix constraints to use .drop_all() instead of dropping database

erseco:
Fix constraints to use `use_alter=True` in the foreign key between _Users_ and _Teams_, this allows to use the sqlalchemy's built-in` .drop_all()` function instead of dropping database via sqlalchemy-utils' `drop_database()`.

Drop a running database in production is a totally discouraged practice, so a better approach is dropping all database objects. This is the main reason why sqlalchemy don't have built-in methods for dropping and creating databases.

This PR improves PostgreSQL support to try to maintain currently supported databases in the upcoming 3.x versión

This PR also fixes #1397 so I can close that PR

More info about why using `use_alter=True` to avoid constraint errors: https://docs.sqlalchemy.org/en/13/core/constraints.html#creating-dropping-foreign-key-constraints-via-alter


ColdHeat:
Maintaining database compatibility causes nothing but problems and I do not see any benefit to keeping Postgres support inside of CTFd. 

For example, by specifying the name you likely stop the automatic foreign key name generation so there's probably a weird oddly named foreign key inside of MySQL (the actual CTFd database). 

You can already deploy MySQL/MariaDB on Heroku so I don't see why you need to fight to support Postgres.

erseco:
PostgreSQL on Heroku brings a 10k rows database for free. The options for MySQL/MariaDB only brings a 5Mb databases for free, this is around 1K rows, maybe less so is impossible to use in a minimal CTFd setup.

Also, the resource usage of PostgreSQL in Docker is lower than MySQL/MariaDB so my students can develop in our low resource machines. PostgreSQL docker container is taking 8Mb of RAM while MySQL/MariaDB is taking more than 200Mb in idle.

These are the main reasons to want to try to maintain the compatibility with PostgreSQL, I would like to collaborate with this project giving patches, but if it's a problem I can maintain my own fork.

Best regards!

ColdHeat:
Thanks for your PR, I'm appreciative that you're using CTFd for education. There isn't an issue with sending a PR but I take issue with PRs that solve one-off problems specific to one particular use case because they end up very disorganized. In this case this tramples on a general expectation that CTFd only supports Postgres unofficially and starts adding warts on MySQL. 

I admittedly have considered retaining Postgres support because I have read the literature that it's "superior" but then I remember the hours that I've spent investigating weird differences between how MySQL and Postgres handle data. There's a lot of added complexity for a database that's not really very widely used when the recommendation is MySQL. 

To retain Postgres support I would have to see most if not all of the Postgres specific code removed from the CTFd code base and replaced with db-agnostic code.

erseco:
Thanks to you for your project, I started the course doing an easy CTF about informatics, later they learn how to deploy locally, later to deploy on the cloud, and finally how to modify the code. This project fits all needs of the course.

I agree about don't add complexity to support different databases, that is why I did this new PR, the constraint problem is also in **MySQL**, so this fixes it, you can see the raised error when calling to `app.db.drop_all()` function.

```
sqlalchemy.exc.CircularDependencyError: Can't sort tables for DROP; an unresolvable foreign key dependency exists between tables: teams, users.  Please ensure that the ForeignKey and ForeignKeyConstraint objects involved in the cycle have names so that they can be dropped using DROP CONSTRAINT.
```
This PR fixes this error and seems a cleaner approach because is the recommended usage in sqlachemy docs. More info about why using use_alter=True to avoid constraint errors: https://docs.sqlalchemy.org/en/13/core/constraints.html#creating-dropping-foreign-key-constraints-via-alter 

In the future, my idea is removing the `sqlachemy-utils` dependency using just `sqlalchemy` for all the tasks.

I checked the code and maybe the messiest problem with different database engines is in the import/export system, that can be changed for an easier approach using the built-in `sqlalchemy` serializer exporting to pickle. Another approach is loading the data in the ORM and let it set the correct ids instead of doing tricky things to reset the auto-increments.

About your reads: I worked as DevOps for years and yes, PostgreSQL is so much better, that's why it is the default engine in other projects like Django.

CTFd/utils/exports/__init__.py
```diff
@@ -165,16 +159,11 @@ def import_ctf(backup, erase=True):
 
     members = first + members
 
-    upgrade(revision=alembic_version)
```
ColdHeat, line 168: By removing this line, tables are created to the latest migration and imports will fail if they are from a previous version that had different data.

CTFd/utils/migrations/__init__.py
```diff
@@ -20,22 +19,24 @@ def create_database():
...
 
-    # Creates database if the database database does not exist
-    if not database_exists_util(url):
+    # Creates database if the database does not exist
+    try:
+        # If the user don't have permission to access to "postres" databasen  this
+        # will fail: https://github.com/kvesteri/sqlalchemy-utils/pull/372
+        # in that case we asume that the database was previously created and we can't
+        # drop or create the database
+        exists = database_exists_util(url)
```
ColdHeat, line 28: Even if we ignore the Postgres support discussion, I don't know if I would accept this change. Too hacky for my tastes and probably could affect other things. Maybe better to only exception handle under Postgres. 

We would need to wait for https://github.com/kvesteri/sqlalchemy-utils/pull/372

CTFd/utils/exports/__init__.py
```diff
@@ -123,8 +118,7 @@ def import_ctf(backup, erase=True):
         )
 
     if erase:
-        drop_database()
-        create_database()
+        app.db.drop_all()
```
ColdHeat, line 121: While I was working on #1419, I experimented with this functionality to see if it resolved an issue I was running into but it didn't. This might be better practice but I am not convinced that it has an actual benefit in CTFd.

CTFd/models/__init__.py
```diff
@@ -244,7 +244,10 @@ class Users(db.Model):
     verified = db.Column(db.Boolean, default=False)
 
     # Relationship for Teams
-    team_id = db.Column(db.Integer, db.ForeignKey("teams.id"))
+    # use_alter=True along with name='' adds this foreign key after Teams has been created to avoid circular dependency
+    team_id = db.Column(
+        db.Integer, db.ForeignKey("teams.id", use_alter=True, name="users_team_id_fkey")
```
ColdHeat, line 249: I don't like pinning the name here. If we can tap into the Sqlalchemy built-in name generator, I would be more willing to merge this in but I still am not certain I want to go down the path of supporting multiple databases overall.

erseco: Removed the name and let sqlalchemy to set the name of the constraint using its internal conventions: https://alembic.sqlalchemy.org/en/latest/naming.html

CTFd/utils/exports/__init__.py
erseco: Restored that line

erseco:
Closing this PR because don't seems more needed

Issue #856, opened 2019-01-28
Config.py Variable Documentation


Helithumper:
Within the `CTFd/config.py` file, there are a few variables defined under the `TestConfig` class which have no documentation surrounding them. What do they do?

```
PRESERVE_CONTEXT_ON_EXCEPTION = False
TESTING = True
DEBUG = True
SERVER_NAME = 'localhost'
SAFE_MODE = True```

If someone could explain exactly what each of these do, it would be useful to add documentation surrounding their use.

ColdHeat:
Hi @Helithumper, this seems related to #742 but overall the top 4 are documented by Flask. The only one custom to CTFd is the last one which defines whether plugins are loaded by CTFd. 

`SAFE_MODE = True` means don't load plugins. In terms of the tests, disabling plugin loading helps ensure the consistency of generated output. 

I'm not sure if there really needs to be documentation for the Flask specific values but perhaps there should be some for the CTFd specific values. Ultimately we should have that documentation in Sphinx first I think.

PR #1410, opened 2020-05-14
detailed documentation issue #856 in comments 

abhinavprkash:
Within the CTFd/config.py file, there are a few variables defined under the TestConfig class which have no documentation surrounding them. What do they do?

PRESERVE_CONTEXT_ON_EXCEPTION = False
TESTING = True
DEBUG = True
SERVER_NAME = 'localhost'
SAFE_MODE = True
I tried to add some information in the form of comments at the bottom class TestingConfig(Config) in config.py file. Please check it out thanks

### DIFF
--- a/CTFd/config.py
+++ b/CTFd/config.py
@@ -273,3 +273,8 @@ class TestingConfig(Config):
     CACHE_TYPE = "simple"
     CACHE_THRESHOLD = 500
     SAFE_MODE = True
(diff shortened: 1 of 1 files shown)


ColdHeat:
This is in the wrong format and copy pasted from around the internet. Please clean it up and provide better documentation about each option.

PR #1412, opened 2020-05-14
HTTP Request Smuggling in Gunicorn v19.9.0

socketz:
Upgrade gunicorn to version 19.10.0 or 20.0.1.

Affected versions of this package are vulnerable to HTTP Request Smuggling. It fails to properly process the `Transfer-Encoding` and `Content-Length` headers when both are present in a package request. This allows for conflicting information to be sent regarding the length of the package, which when processed by back-end servers under certain configurations would allow for malicious users to bypass security controls, gain unauthorized access to sensitive data, and directly compromise other application users.


ColdHeat:
Interesting bug class... I doubt a lot of people use CTFd with chunked requests but this is a good PR anyway.

Issue #1435, opened 2020-05-23
Unable to remove members from team


itsmingjie:
**Environment**:

  - CTFd Version/Commit: 2.4.1
  - Operating System: Ubuntu 18.04
  - Web Browser and Version: Firefox 76

**What happened?**

When trying to remove a user from a team, clicking on the remove shows the modal for a split second but the page is then subsequently redirected to the user detail page.

**What did you expect to happen?**

User is successfully removed from the page.

**How to reproduce your issue**

Head to `/admin/teams/1` (or any valid team), scroll down to "Team Members", and click remove on any team member.

PR #1436, opened 2020-05-23
Remove redundant overriding link

itsmingjie:
See #1435 for issue description. This is an attempted quick fix. 

The link on the `<tr>` element is not necessary since there is already a link on line 215 (on the username column).


ColdHeat:
The reason for this link is that it lets you click on the rest of the `tr` and link to the user's profile. Looks like this is scoped too widely though.

ColdHeat:
I'm going to close this in favor of #1437

Issue #467, opened 2017-11-18
Lost connection to MySQL server during query


FreakyFreddie:
**Environment**:

  - CTFd Version/Commit: Latest as of 11/11/2017
  - Operating System: Ubuntu 16.04 LTS, deploying platform in Docker containers, using docker-compose up
  - Web Browser and Version: Firefox: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:56.0) Gecko/20100101 Firefox/56.0 and Chrome: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36

**What happened?**
When adding a new challenge, I took some time to fill in the form. When sending it, I received an internal server error. I tried it again and it worked normally.

**What did you expect to happen?**
The challenge to be created normally.

**How to reproduce your issue**
Click the button to create a new challenge, fill in the form. Then, leave the page open for around 15 minutes and click the button.

**Any associated stack traces or error logs**
Error.log
```
[2017-11-13 08:28:13,623] ERROR in app: Exception on /admin/chal/new [POST]

File "/usr/local/lib/python2.7/site-packages/pymysql/connections.py", line 1033, in _read_bytes
    CR.CR_SERVER_LOST, "Lost connection to MySQL server during query")
OperationalError: (pymysql.err.OperationalError) (2013, 'Lost connection to MySQL server during query') [SQL: u'SELECT config.id AS config_id, config.`key` AS config_key, config.value AS config_value \nFROM config \nWHERE config.`key` = %(key_1)s \n LIMIT %(param_1)s'] [parameters: {u'param_1': 1, u'key_1': 'setup'}]
```

Access.log
```
10.0.4.2 - - [13/Nov/2017:08:28:13 +0000] "POST /admin/chal/new HTTP/1.1" 500 2700 "https://10.0.7.4/admin/chal/new" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36"
```

The following errors may also be related to the same issue:
Error.log
```
[2017-11-13 09:12:36,352] ERROR in app: Exception on /challenges [GET]
File "/usr/local/lib/python2.7/site-packages/pymysql/connections.py", line 1033, in _read_bytes
    CR.CR_SERVER_LOST, "Lost connection to MySQL server during query")
[...]

PR #1438, opened 2020-05-24
Bugfix: Improved MySQL server connectivity

nomuken:
CTFd does not send queries to MySQL for a certain amount of time, so it loses connection to the MySQL server.
It doesn't happen with Docker deployments, but it does frequently occur on-premises.
In this PR, CTFd's connection pool sends a ping to the MySQL server using the SQLAlchemy feature.

Related Issue: #1395 #467
SQLAlchemy document: https://docs.sqlalchemy.org/en/13/core/pooling.html#dealing-with-disconnects

### DIFF
--- a/CTFd/models/__init__.py
+++ b/CTFd/models/__init__.py
@@ -8,7 +8,7 @@ from sqlalchemy.orm import column_property, validates
 from CTFd.utils.crypto import hash_password
 from CTFd.utils.humanize.numbers import ordinalize
 
-db = SQLAlchemy()
+db = SQLAlchemy(engine_options=dict(pool_pre_ping=True))
 ma = Marshmallow()
 
 


ColdHeat:
I kind of doubt that this is best solution since it would add more work onto the database. What is the database connection timeout on your database? 
https://flask-sqlalchemy.palletsprojects.com/en/2.x/config/#timeouts

nomuken:
MariaDB config: (I set the same value in `docker-compose.yml `)
```
MariaDB [(none)]> show variables like "%timeout%";
+---------------------------------------+----------+
| Variable_name                         | Value    |
+---------------------------------------+----------+
| connect_timeout                       | 28800    |
| deadlock_timeout_long                 | 50000000 |
| deadlock_timeout_short                | 10000    |
| delayed_insert_timeout                | 300      |
| idle_readonly_transaction_timeout     | 0        |
| idle_transaction_timeout              | 0        |
| idle_write_transaction_timeout        | 0        |
| innodb_flush_log_at_timeout           | 1        |
| innodb_lock_wait_timeout              | 50       |
| innodb_rollback_on_timeout            | OFF      |
| interactive_timeout                   | 28800    |
| lock_wait_timeout                     | 86400    |
| net_read_timeout                      | 30       |
| net_write_timeout                     | 60       |
| rpl_semi_sync_master_timeout          | 10000    |
| rpl_semi_sync_slave_kill_conn_timeout | 5        |
| slave_net_timeout                     | 60       |
| thread_pool_idle_timeout              | 60       |
| wait_timeout                          | 28800    |
+---------------------------------------+----------+
```

> https://flask-sqlalchemy.palletsprojects.com/en/2.x/config/#timeouts
I don't know, because we didn't set `SQLALCHEMY_POOL_RECYCLE` in our CTF, but the same problem occurred with `db = SQLAlchemy(engine_options=dict(pool_recycle=5))`.

> I kind of doubt that this is best solution since it would add more work onto the database.
Honestly, I think so too.
I tried many solutions(e.g. change timeout, change pool recycle time) during held our CTF. It only worked was when I added this code.

jedevc:
If adding more work to the database is a concern, would it be possible to gate this functionality behind an environment variable? Then it can be left up to the system administrator to enable it, if they judge the additional cost worth a reduction in 500 errors.

ColdHeat:
@jedevc, I think you can use `SQLALCHEMY_ENGINE_OPTIONS` to specify this. See https://flask-sqlalchemy.palletsprojects.com/en/2.x/config/#timeouts. 

Based on your investigation in #1509 though I think this could be more valuable to merge in some form.

ColdHeat:
I enabled `pool_pre_ping` to be the default in v3 in #1520 so closing this.

Issue #1411, opened 2020-05-14
mysql question


binLep:
I don't know how to sovle it.

**Environment**:

  - CTFd Version/Commit:CTFd-2.4.2
  - Operating System:Ubuntu 16.04/Ubuntu 18.04
  - Web Browser and Version:

**What happened?**
mysql is down

**Any associated stack traces or error logs**

```
cache_1  | 1:C 14 May 07:52:33.668 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
cache_1  | 1:C 14 May 07:52:33.668 # Redis version=4.0.14, bits=64, commit=00000000, modified=0, pid=1, just started
cache_1  | 1:C 14 May 07:52:33.668 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
cache_1  | 1:M 14 May 07:52:33.670 * Running mode=standalone, port=6379.
cache_1  | 1:M 14 May 07:52:33.671 # WARNING: The TCP backlog setting of 511 cannot be enforced because /proc/sys/net/core/somaxconn is set to the lower value of 128.
cache_1  | 1:M 14 May 07:52:33.672 # Server initialized
cache_1  | 1:M 14 May 07:52:33.672 # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
cache_1  | 1:M 14 May 07:52:33.674 * DB loaded from disk: 0.002 seconds
cache_1  | 1:M 14 May 07:52:33.675 * Ready to accept connections
ctfd_1   | Waiting for db: to be ready
db_1     | 2020-05-14 07:41:41+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 1:10.4.13+maria~bionic started.
db_1     | 2020-05-14 07:41:42+00:00 [Note] [Entrypoint]: Switching to dedicated user 'mysql'
db_1     | 2020-05-14 07:41:42+00:00 [Note] [Entrypoint]: Entrypoint script for MySQL Server 1:10.4.13+maria~bionic started.
db_1     | 2020-05-14 07:41:42+00:00 [Note] [Entrypoint]: Initializing database files
db_1     | 
db_1     | 
db_1     | PLEASE REMEMBER TO SET A PASSWORD FOR THE MariaDB root USER !
[...]

PR #1443, opened 2020-05-25
Update docker-compose.yml

fjh1997:
As detailed in #1411
mariadb 10.4.13 may cause a meaningless migrating

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
         internal:
 
   db:
-    image: mariadb:10.4
+    image: mariadb:10.4.12
     restart: always
     environment:
       - MYSQL_ROOT_PASSWORD=ctfd


ColdHeat:
To be honest, I don't see this as an issue in CTFd. MariaDB mostly follows semver so this kind of a regression shouldn't really have happened. 

But I suppose this is still worth merging because ultimately it pins the version. From some reading it seems like 10.3 would be a better pin because it has good MySQL compatibility but we can't just downgrade the MariaDB version so I guess this is the next best thing.

fjh1997:
> To be honest, I don't see this as an issue in CTFd. MariaDB mostly follows semver so this kind of a regression shouldn't really have happened.
> 
> But I suppose this is still worth merging because ultimately it pins the version. From some reading it seems like 10.3 would be a better pin because it has good MySQL compatibility but we can't just downgrade the MariaDB version so I guess this is the next best thing.

If MariaDB really follows semver, patch version 10.4.13 should fix bugs instead of making more bugs. And pinning is  surely the temporary solution before we could figure out how does the regression happend. Anybody has the idea？

Issue #663, opened 2018-08-14
Use Pathlib? 


ColdHeat:
Pathlib is a big improvement over the os.path calls. The only major concern here is how to maintain backwards compatibility with Python 2.

ColdHeat:
This isn't really blocking for 2.0.0.

PR #1444, opened 2020-05-25
Use Pathlib instead of os module refactor

braheezy:
Implementation of #663

Took a stab at moving the project to Pathlib support. Via `make test` on my local machine, I ensured the test results and coverage were the same before and after the refactor.


ColdHeat:
Hmm not so sure I like the pathlib api but it looks like an improvement. Do you mind removing the pathlib2 fallbacks and the addition to .gitignore?

mbraha:
Agreed. In some places, using `Pathlib` over `os.path` was a trivial gain to the code. (Like `Path(path1, path2)` instead of `os.path.join(path1, path2)`. Pathlib also requires the developer knows the Pathlib API (I like this [cheatsheet](https://github.com/chris1610/pbpython/blob/master/extras/Pathlib-Cheatsheet.pdf)). 

There were a few spots I liked after the refactor, like those changes in `/admin/__init__ :: plugins()` and `/plugins/__init__ :: init_plugins()`

.gitignore
```diff
@@ -75,4 +75,4 @@ CTFd/uploads
 node_modules/
 
 # Flask Profiler files
-flask_profiler.sql
+flask_profiler.sql
```
ColdHeat, line 78: This change should be removed.

PR #1447, opened 2020-05-26
Support SQLALCHEMY URLS with "%" sign

snurilov:
Sometimes the SQLALCHEMY URL must contain a percent sign (%). One example of this is when your password contains a %. Another example is when deploying to azure mysql. The username contains an @ sign, which must be encoded as "%40" to pass to sqlalchemy as a URL. Having a percent sign in the URL breaks line 27, because the Python ConfigParser uses % as a special character for evaluating interpolated values.
In order to place the value into the Configuration, the % sign must be escaped with %%.

### DIFF
--- a/migrations/env.py
+++ b/migrations/env.py
@@ -24,7 +24,7 @@ config = context.config
 logger = logging.getLogger("alembic.env")
 
 config.set_main_option(
-    "sqlalchemy.url", current_app.config.get("SQLALCHEMY_DATABASE_URI")
+    "sqlalchemy.url", current_app.config.get("SQLALCHEMY_DATABASE_URI").replace("%","%%")
 )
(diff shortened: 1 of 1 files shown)


ColdHeat:
This code is generated by Flask Migrate. Looks like this is an upstream change that we probably should copy. 

https://github.com/miguelgrinberg/Flask-Migrate/blob/master/flask_migrate/templates/flask/env.py#L25-L27

Overall this code probably needs to be re-generated.

ColdHeat:
@snurilov, your tests didn't pass because you need to run `black` on the files you changed.

ColdHeat:
@snurilov it would be great if you could try out #1447 to see if it also resolves your issues. Thanks for filing this PR to flag the issue.

PR #1449, opened 2020-05-29
Add a session invalidation strategy inspired by Django

### LINKED ISSUE #1430: Session invalidation
We need a better way to invalidate sessions per user. Django has an approach that we will likely copy because of its flexibility: https://docs.djangoproject.com/en/3.0/topics/auth/default/#session-invalidation-on-password-change

ColdHeat:
* Use a session invalidation strategy inspired by Django. https://docs.djangoproject.com/en/3.0/topics/auth/default/#session-invalidation-on-password-change
* Closes #1430


ColdHeat:
Needs tests

PR #1462, opened 2020-05-31
Get basic implementation of HTML sanitization working for pages

### LINKED ISSUE #1332: HTML Sanitization
Since CTFd's beginning, admins inserting HTML was considered valid. Instead we are going to take a different approach and only allow a sanitized subset of HTML in both the Pages editor and the challenge UI. 

Some investigation needs to be done to see what's allowed by admins and what isn't. 

Example libs that will work for our use case:
https://github.com/mozilla/bleach
https://lxml.de/lxmlhtml.html#cleaning-up-html

ColdHeat:
* Closes #1332
* Pages by default now strip script tags and other potential XSS vectors
* lxml and html5lib are now pinned dependencies
* Challenge plugins rewritten to allow for better re-useability of template content and allow more control from the theme side


ColdHeat:
Blocked on #1478 but the page sanitization works

Issue #1467, opened 2020-06-03
Explore a Form global


ColdHeat:
Something like `Forms.LoginForm` from within Jinja could make it easy to customize a theme to look different. wtforms is popular but I am not necessarily a huge fan. Needs investigation as to how far we will use a form library. 

Related to #1234

ColdHeat:
Many of these forms were rewritten. Some of them remain but they are admin ones and can be slowly converted over time. The main user facing ones are converted and can be used in themes.

PR #1469, opened 2020-06-03
WIP: Add form globals

ColdHeat:
* Work on #1467 
* Creates the Form global which will contain all forms.

Whether or not these forms will actually use wtforms to validate the input remains to be seen. They should as a first pass be used to generate the HTML.


CTFd/forms/auth.py
```diff
@@ -0,0 +1,32 @@
+from wtforms import PasswordField, StringField, SubmitField
+from wtforms.fields.html5 import EmailField
+from wtforms.validators import InputRequired
+
+from CTFd.forms import BaseForm
+
+
+class RegistrationForm(BaseForm):
+    name = StringField("User Name", validators=[InputRequired()])
```
ColdHeat, line 9: These weren't required before but it makes sense to keep them as required. If you want to modify the behavior of a form you should make another form.

CTFd/forms/auth.py
```diff
@@ -0,0 +1,32 @@
...
+    email = EmailField("Email", validators=[InputRequired()])
+    password = PasswordField("Password", validators=[InputRequired()])
+    submit = SubmitField("Submit")
+
+
+class LoginForm(BaseForm):
+    name = StringField("User Name or Email", validators=[InputRequired()])
```
ColdHeat, line 16: Same for here

ColdHeat:
The user facing side looks fine but the admin panel still needs to be converted

PR #1498, opened 2020-06-19
API Pagination

ColdHeat:
* Add pagination to the API endpoints that return large amounts of db rows


ColdHeat:
I think this is too much. I think it's better to restrict this to just users and teams b/c those are the endpoints that return the most significant amount of data.

memir0:
I have implemented pagination in my docker plugin for ctfd. I use a url parameter for current page. It might be useful for you.
https://github.com/memir0/CTFd-Docker/blob/master/__init__.py 
see function admin_overview() on line 130

PR #1501, opened 2020-06-21
Experiment with integrating webargs into flask-restplus

ColdHeat:
Experimenting with getting webargs to play nice with flask-restx


ColdHeat:
webargs is too much of a fight to merge with Flask-restx. This is my overall problem with marshmallow aside from possible speed improvements. It seems like you're either in the ecosystem entirely or you're not. Because pydantic is mostly json schema, it's a lot easier to marry with Flask-restx and openapi.

PR #1507, opened 2020-06-23
Display current attempts in challenge view when max attempts is enabled

### LINKED ISSUE #1477: Show current attempts in challenge view
Showing how many attempts the user has tried on a given challenge. Helps with max attempts as well.

ColdHeat:
* Display current attempts in challenge view when max attempts is enabled
* Closes #1477


ColdHeat:
image

PR #1529, opened 2020-07-04
Fix challenge filter in /api/v1/statistics/challenges/solves

hakatashi:
Since challenge's state cannot be "hidden" and "locked" simultaneously, this statement is pretty useless.
I bet the intention is to filter out challenges that is not visible to end users, so I fixed this to follow that spec.


ColdHeat:
Your assumption is probably right. Locked is a holdover from when CTFd v2 was being developed and challenge requirements were being written. I'll merge this after I can confirm that this is the right behavior.

ColdHeat:
This behavior looks like the correct intention of not showing hidden challenges in challenge solve count statistics. I think this is the right idea so I'm going to merge this in.

PR #1533, opened 2020-07-07
Spinner centering bootstrap css

ColdHeat:
* Fix spinners to not use a fixed height CSS rule to vertically center. Instead accomplish this with flexbox CSS rules built into Bootstrap


ColdHeat:
I'm merging this so that I can merge upstream fixes into the branch that this is failing on.

PR #1561, opened 2020-07-21
Add a raw_query parameter to the standings functions

ColdHeat:
* Add a `raw_query` parameter to the standings functions to get the raw standings query


ColdHeat:
I think I remember why I didn't do this before. It's difficult to cache the result, resulting in errors when you query the function. 

Leaving this as an open item because I think it's actually a useful idea.

PR #1571, opened 2020-07-24
Fix issue with previewing certain pages

ColdHeat:
* Fix previewing pages when page attributes are set


ColdHeat:
This should have a test so I agree with codecov here

Issue #756, opened 2018-11-26
Generic location attribute for users/teams


ColdHeat:
We need a solution to allow groups smaller than countries to denote where they come from. E.g. State, City, County, School, etc. 

Should we store states for countries? 

This gets complicated because what if we need to get deeper (county, school, grade, etc.)

ColdHeat:
After some thought, I think this is best implemented as a User/Team "tag". Similar to what exists for Challenges.

ColdHeat:
The way that Wordpress approaches this is it uses a table much like the Config table that just maps a key to a value. It basically lets you store a one layer deep dictionary for users. 

I don't *really* like that approach because it isn't very constrained and kind of difficult to edit. I would argue that if we didn't already have the affiliation, country, website fields, etc, it would make more sense to put those there. 

![image](https://user-images.githubusercontent.com/166333/84941400-5a5c8680-b0af-11ea-99bd-aef56bcaffc5.png)

ColdHeat:
Looking at Discourse, they have a more flexible approach that involves two tables, one to specify the field requested, and one to contain the user's submission. 

It looks like a better approach that I think is worth prototyping. The only issue really is that it involves basically building some kind of form builder.

ColdHeat:
With #1608, there are now custom user fields. All that's missing is team fields now.

ColdHeat:
Closed with #1615

PR #1608, opened 2020-08-18
756 generic data fields

ColdHeat:
* Works on #756 
* Adds the ability to add custom fields for registration

TODO: 

----

- ~Need a lot of tests in a separate PR against this one~
- ~Generate migration~
- ~Add affiliation, website, and country(?) field special cases. Maybe?~
- ~Admin form extra labels are not bolded~
- ~Add support for ingesting and updating checkbox fields~
- ~Add checkbox field type~
- ~Move imports into the right place~
- ~Properly show which fields are required during registration~
- ~Remove model comments~
- ~Review code for simplification possibilities~
  - ~Simplified form.extra~
  - ~Simplified jinja form extra parsing~
- ~Separate FieldEntries into different types~

Future TODO:
- Add custom fields for teams


CTFd/forms/auth.py
```diff
@@ -4,13 +4,39 @@
...
+        validators = []
+        if field.required:
+            validators.append(InputRequired())
+        setattr(
+            _RegistrationForm,
+            f"fields[{field.id}]",
+            StringField(
+                field.name, description=field.description, validators=validators
+            ),
+        )
```
ColdHeat, line 37: Could this be seperated out into a decorator?

ColdHeat: I actually don't think so

PR #1540, opened 2020-07-13
Fix error on flag submission when using dynamic challenge.

h4ckd0tm3:
The reason this happens is because the view.js in this plugin uses #submission-input as id for the input field when it's in fact #challenge-input.

* Closes #1542

PR #1609, opened 2020-08-18
Match challenges and dynamic challenges

ConnorNelson:
Both default challenge types overload the input block and submit block in the same way. This changes the default template to match these, so that other plugins don't have to overload it.

[/CTFd/plugins/challenges/assets/view.html](https://github.com/CTFd/CTFd/blob/12857797bbe6b7b9645e715138acee01db6e79e5/CTFd/plugins/challenges/assets/view.html)
[/CTFd/plugins/dynamic_challenges/assets/view.html](https://github.com/CTFd/CTFd/blob/12857797bbe6b7b9645e715138acee01db6e79e5/CTFd/plugins/dynamic_challenges/assets/view.html)

With this change, we could go the extra step of replacing both templates simply with `{% extends "challenge.html" %}`.

This also fixes an issue related to #1540, since the default template uses `submission-input` instead of `challenge-input` and `submit-key` instead of `challenge-submit`.

### DIFF
--- a/CTFd/themes/core/templates/challenge.html
+++ b/CTFd/themes/core/templates/challenge.html
@@ -96,13 +96,13 @@
 						<div class="row submit-row">
 							<div class="col-md-9 form-group">
 								{% block input %}
-									<input id="challenge-id" type="hidden" value="{{ challenge.id }}">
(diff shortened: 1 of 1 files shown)


ColdHeat:
Thanks for raising this PR! This looks legit, I'll play around with it when I get a chance and merge after that.

PR #1623, opened 2020-08-28
Fix config default logic

ConnorNelson:
`empty_str_cast` allows for a default value. This is probably how defaults should be managed. The current logic is definitely incorrect for `true` defaults, since when trying to override it to `false`, it ultimately resolves to `false or true`, which is `true`.


ColdHeat:
Thanks for the fix! Yes I definitely think that the `empty_str_cast` default should be used more but I wasn't sure what the side effects would be so it's good to know that it's the better approach. 

I will probably switch over to default entirely in a separate PR.

PR #1625, opened 2020-08-29
Compress notification sound more

ColdHeat:
* Compress `notification.webm` to save some file size

I didn't perceive any difference in sound. If there is it's tough to hear.


ColdHeat:
This seems to increase the size of the individual page bundles which is unfortunate.

ColdHeat:
This approach isn't necessary. Perhaps the settings to compress the notification sound is useful but this shouldn't be merged.

Issue #1622, opened 2020-08-27
redis: max number of clients reached


ConnorNelson:
Hello, we have several hundred users concurrently using a CTFd instance. After approximately 8 hours of usage, we faced internal server errors for nearly everything. This trace back is the underlying cause. I would be happy to look into fixing it in a couple of days, but I figured I'd report the issue now in case it is an already known issue with a known solution. Thanks.

```
[2020-08-27 17:45:31 +0000] [29] [ERROR] Error handling request /
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 2446, in wsgi_app
    ctx.push()
  File "/usr/local/lib/python3.7/site-packages/flask/ctx.py", line 390, in push
    self.session = session_interface.open_session(self.app, self.request)
  File "/opt/CTFd/CTFd/utils/sessions/__init__.py", line 77, in open_session
    val = cache.get(self.key_prefix + sid)
  File "/usr/local/lib/python3.7/site-packages/flask_caching/__init__.py", line 243, in get
    return self.cache.get(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/flask_caching/backends/rediscache.py", line 114, in get
    self._read_clients.get(self._get_prefix() + key)
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 1606, in get
    return self.execute_command('GET', name)
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 901, in execute_command
    return self.parse_response(conn, command_name, **options)
  File "/usr/local/lib/python3.7/site-packages/redis/client.py", line 915, in parse_response
    response = connection.read_response()
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 730, in read_response
    response = self._parser.read_response()
  File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 337, in read_response
    raise error
redis.exceptions.ConnectionError: max number of clients reached
```

We are using commit 976b191b535018827e8d8f1eadefb2e352179e19 (3.0.1).

ColdHeat:
I believe Flask-caching pools redis connections but the Notifications system could potentially consume pubsub clients. You can disable it by toggling `SERVER_SENT_EVENTS` in `config.ini`. 

To figure this out, you'd need to know what redis's max client amount is, what type of clients they are, and what factor is increasing the client list. https://redis.io/topics/clients

[...]

PR #1631, opened 2020-09-06
Added filters for notifications for users/teams

calebstewart:
I'm unsure whether this was an planned/abandoned feature, however I've added the ability to filter notifications by user/team ID. The information was already in the `Notification` database model, but appeared to be unused. As some background, I'd like to use this feature to allow per-user notifications. I'm developing a plugin to integrate a Kubernetes cluster with CTFd to develop active challenges, and notifying users when their deployments are being reset or shutdown would be a nice feature.

As it stands, I simply added filtering for both events and notifications which removes notifications not destined for the current user/team (depending on if teams are enabled). I did not add any UI mention of this (e.g. in the admin panel notifications dialog), however the ability to create notifications tagged for a user or team existed implicitly in the notifications API, because any of the properties of the `Notification` model could be specified in the post request. The UI side is not my forte, and the these modifications don't break or change any existing functionality that I'm aware of, so I left the old notification creation page unchanged.

Summary/explanation of changes per file:

- `CTFd/api/v1/notifications.py` - modified the `get` handler to insert extra filters on SQLAlchemy query to filter out user and team IDs if present. Also checks for current user mode.
- `CTFd/events/__init__.py` - based what I read in the notifications API post handler, this is how async notifications are sent to the users. Here, I intercepted the notification in the generator and applied similar logic to the above to ignore events not applicable to our current user.
- `CTFd/views.py` - Similar to the notifications API `get` handler. Added the same filters to SQLAlchemy query to remove non-applicable notifications.

If I misunderstood the purpose of the fields in the `Notifications` model, please let me know. I tested locally and was able to receive notifications destined for my user, and was unable to see notifications destined for other users, and was still able to see notifications destined for all users.

### DIFF
--- a/CTFd/api/v1/notifications.py
+++ b/CTFd/api/v1/notifications.py
@@ -2,6 +2,7 @@ from typing import List
 
 from flask import current_app, request
 from flask_restx import Namespace, Resource
+from sqlalchemy import or_
 
 from CTFd.api.v1.helpers.request import validate_args
 from CTFd.api.v1.helpers.schemas import sqlalchemy_to_pydantic
@@ -11,6 +12,9 @@ from CTFd.models import Notifications, db
 from CTFd.schemas.notifications import NotificationSchema
 from CTFd.utils.decorators import admins_only
 from CTFd.utils.helpers.models import build_model_filters
+from CTFd.utils.modes import TEAMS_MODE
+from CTFd.utils.user import is_admin, get_current_user, get_current_team
+from CTFd.utils import get_config
 
 notifications_namespace = Namespace(
     "notifications", description="Endpoint to retrieve Notifications"
@@ -68,9 +72,26 @@ class NotificantionList(Resource):
         field = str(query_args.pop("field", None))
         filters = build_model_filters(model=Notifications, query=q, field=field)
 
-        notifications = (
-            Notifications.query.filter_by(**query_args).filter(*filters).all()
+        notifications = Notifications.query.filter_by(**query_args).filter(*filters)
+
+        # Filter on user/team id if present in the notification
+        notifications = notifications.filter(
+            or_(
+                Notifications.user_id == None,
+                Notifications.user_id == get_current_user().id,
+            )
         )
+        if get_config("user_mode") == TEAMS_MODE:
+            notifications = notifications.filter(
+                or_(
+                    Notifications.team_id == None,
+                    Notifications.team_id == get_current_team().id,
+                )
+            )
+
+        # Grab the full list
+        notifications = notifications.all()
+
         schema = NotificationSchema(many=True)
         result = schema.dump(notifications)
         if result.errors:
--- a/CTFd/events/__init__.py
+++ b/CTFd/events/__init__.py
@@ -1,7 +1,9 @@
 from flask import Blueprint, Response, current_app, stream_with_context
 
-from CTFd.utils import get_app_config
+from CTFd.utils import get_app_config, get_config
+from CTFd.utils.modes import TEAMS_MODE
 from CTFd.utils.decorators import authed_only, ratelimit
+from CTFd.utils.user import get_current_user, get_current_team
 
 events = Blueprint("events", __name__)
 
@@ -13,6 +15,19 @@ def subscribe():
     @stream_with_context
     def gen():
         for event in current_app.events_manager.subscribe():
+            # Filter notifications based on user and team ID
+            if event.type == "notification":
+                if (
+                    event.data["user"] is not None
+                    and event.data["user"] != get_current_user().id
+                ):
+                    continue
+                if get_config("user_mode") == TEAMS_MODE:
+                    if (
+                        event.data["team"] is not None
+                        and event.data["team"] != get_current_team().id
+                    ):
+                        continue
             yield str(event)
 
     enabled = get_app_config("SERVER_SENT_EVENTS")
--- a/CTFd/views.py
+++ b/CTFd/views.py
@@ -5,6 +5,7 @@ from flask import current_app as app
 from flask import redirect, render_template, request, send_file, session, url_for
 from flask.helpers import safe_join
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy import or_
 
 from CTFd.cache import cache
 from CTFd.constants.config import (
@@ -55,6 +56,9 @@ from CTFd.utils.security.signing import (
 )
 from CTFd.utils.uploads import get_uploader
 from CTFd.utils.user import authed, get_current_user, is_admin
+from CTFd.utils.modes import TEAMS_MODE
+from CTFd.utils.user import is_admin, get_current_user, get_current_team
+from CTFd.utils import get_config
 
 views = Blueprint("views", __name__)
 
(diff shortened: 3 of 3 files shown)


calebstewart:
I've never used `codecov`, so I'm not exactly sure why it's not passing the check. There's a lot of information in that dump, and none of it is very clear. If anyone can point me in the right direction, I'd be happy to fix whatever is wrong there. :+1:

ColdHeat:
Per-user/team notifications was an idea but the original intention was to cover announcements. The user_id and team_id columns were just there to cover the potential idea to reduce the amount of migrations needed later. Only announcements were implemented b/c the tech do do real-time notifications has been a big fat hassle and has probably contributed to some negative CTFd sentiment b/c of how hard it is to get right. 

I am not against user-team notifications but it has to be implemented very carefully. 

- You should look at the changes in the 3.1.0-dev branch for the changes to the underlying EventManager system for #1622. 
- You should investigate how other real time systems implement this behavior. Discourse comes to mind with its real time popups for replies to threads. If it turns out that they've implemented this behavior by having all events get filtered out by the web worker instead of having separate channels I would be more open to merging an implementation like this. 

Regarding code, codecov is failing b/c your code doesn't have any tests. You can look at the diff tab in codecov if you can see it. Red lines aren't covered by any tests.

calebstewart:
That all makes sense. This was basically a naive first stab at a feature I would like to see. I'm taking a look at the Discourse source code now. From an initial pass through the code, it looks like their push notification service uses the browser's serviceWorker interface to enable desktop notifications. They use the Ruby gem Web-Push. There is an equivalent pywebpush module for Python [here](https://github.com/web-push-libs/pywebpush), and a crash course on Medium [here](https://medium.com/@suryasankar/how-to-setup-basic-web-push-notification-functionality-using-a-flask-backend-1251a5413bbe). 

I'm not intimately familiar with the `serviceWorker` interface, but it seems that each browser uses some third-party server for caching notifications. The notifications are encrypted using a private key and decrypted by the browser. From what I understand, the issue with this construct would be that for each client, CTFd would need to store their subscription details somewhere accessible. For push notifications going to all clients, all client details would need to be looked up and each notification sent individually. In the case of discourse, sending a notification to a single user looks like this:

```ruby
    subscriptions(user).each do |subscription|
      subscription = JSON.parse(subscription.data)
      send_notification(user, subscription, message)
    end
```

Where `send_notification` calls the `Web-Push` gem, and `subscriptions(user)` is a database query for `PushSubscription` objects which describe each subscription for that user. Sending a notification for all users (similar to what the current events api does), would require you to wrap this loop in a loop over all users to get the notification out everywhere. However, Discourse also appears to limit subscriptions to one per user (based on [this](https://github.com/discourse/discourse/blob/master/app/services/push_notification_pusher.rb#L42)). On the upside, the processing on the JS end becomes asynchronous. Simply subscribing to push events after registering via the `navigator.serviceWorker.register` and `PushManager.subscribe`.

From what I can see, doing this in CTFd would require the following:

- Generating a VAPID public/private key (could be specified in the config)
- Creating a "service worker" js file.
- Providing an endpoint to subscribe to push notifications which receives the `PushManager.subscribe().toJSON()` output from the client and stores it in the database.
- Create helper methods to send push notifications which look up the corresponding subscription information for the intended recipients and utilize `pywebpush` to encrypt and dispatch the notifications.
  - Probably methods such as `push_to_user`, `push_to_all`, and `push_to_team` or something similar.

The contents of these messages are abstract similar to the contents of the current events system. I don't think the notification system itself would need to be changed at all, unless I'm missing something. It would essentially look like a replacement (or modification?) of the EventManager.

ColdHeat:
I think you're looking at push notifications which isn't really what CTFd Notifications are. If you want push notifications, there is technically an archived plugin that does it, but it's not really the right functionality. You would probably be looking for polling type functionality?

calebstewart:
Is that not what the `EventManager` is essentially implementing? I may be misunderstanding some piece of this, but it seems that the events manager is essentially implementing push notifications. Granted, they aren't system notifications, but it's implementing a asynchronous server-driven event API, which is what the browser's push events are for. The actual events received by the browser don't have to be actual desktop notifications, but can be. They are simply events delivered to clients asynchronously. They show desktop notifications for the user via `self.registration.showNotification` or dispatch messages back to the DOM for update via `self.client.postMessage` function. 

It's very possible I'm misunderstanding the underlying purpose EventManager interface, but that's what I understood from reading the CTFd source.

ColdHeat:
You're confusing Push Notifications for any kind of JS that receives data from a backend. CTFd Notifications don't require any  keys, workers, or worker registration and thus aren't technically push notifications. To receive a CTFd Notification, your browser has to be open with CTFd site open. 

If you think push notifications fit your use case better, you can take a look at the old CTFd-Notify plugin which is the precursor to the current Notifications system: https://github.com/CTFd/CTFd-Notify. 

Push notifications as your examples describe is a much larger discussion (as evidenced by the amount of additional data you need).

ColdHeat:
I did some research here and it looks like the core idea you have is right but it needs some refinement and the upcoming improvements from 3.1.0. I'll leave a review here but to get merged you'll at least need some tests.

CTFd/events/__init__.py
```diff
@@ -13,6 +15,19 @@ def subscribe():
...
+                    event.data["user"] is not None
+                    and event.data["user"] != get_current_user().id
+                ):
+                    continue
+                if get_config("user_mode") == TEAMS_MODE:
+                    if (
+                        event.data["team"] is not None
+                        and event.data["team"] != get_current_team().id
+                    ):
+                        continue
```
ColdHeat, line 30: This code should be moved into the event managers and allow for subscribing to events that have the specified user/team IDs. You should use `data["user_id"]` instead of `["user"]`. Same with team.

CTFd/events/__init__.py
```diff
@@ -13,6 +15,19 @@ def subscribe():
     @stream_with_context
     def gen():
         for event in current_app.events_manager.subscribe():
+            # Filter notifications based on user and team ID
+            if event.type == "notification":
+                if (
+                    event.data["user"] is not None
+                    and event.data["user"] != get_current_user().id
```
ColdHeat, line 22: get_current_user requires you to query the db. Not worthwhile if you only need the user's ID.

CTFd/events/__init__.py
```diff
@@ -13,6 +15,19 @@ def subscribe():
...
+                ):
+                    continue
+                if get_config("user_mode") == TEAMS_MODE:
+                    if (
+                        event.data["team"] is not None
+                        and event.data["team"] != get_current_team().id
```
ColdHeat, line 28: get_current_team requires you to query the db. Not worthwhile if you only need the team's ID.

CTFd/api/v1/notifications.py
```diff
@@ -68,9 +72,26 @@ def get(self, query_args):
...
+            notifications = notifications.filter(
+                or_(
+                    Notifications.team_id.is_(None),
+                    Notifications.team_id == get_current_team().id,
+                )
+            )
+
+        # Grab the full list
+        notifications = notifications.all()
+
```
ColdHeat, line 94: This kind of looks unfortunate. Can you add onto the existing filters? And how does the user distinguish between notifications to them vs globally?

CTFd/views.py
```diff
@@ -271,7 +272,25 @@ def integrations():
...
+    if get_config("user_mode") == TEAMS_MODE:
+        notifications = notifications.filter(
+            or_(
+                Notifications.team_id.is_(None),
+                Notifications.team_id == get_current_team().id,
+            )
+        )
+
+    notifications = notifications.all()
+
```
ColdHeat, line 293: Same as above. Can this be consolidated?

ColdHeat (changes requested):
This appears to be missing the code in the API to publish notifications? It would also be nice to have some kind of UI to work with.

PR #1632, opened 2020-09-07
TOS and Privacy Policy Pages

### LINKED ISSUE #1621: TOS and Privacy Policy pages
We should add a config entry to allow users to upload their own Terms of Service and Privacy Policy pages. 

This should probably be a simple textarea for people to specify the basic text. We should also make changes to the registration fields. Discourse has a very similar feature that we should draw inspiration from.

ColdHeat:
* Adds a legal section where users can add a terms of service and privacy policy
* Optionally show links to the TOS and Privacy Policy on the registration page
* Closes #1621 

image


ColdHeat:
88.23% of diff hit (target 89.72%) 

This is good enough

Issue #1639, opened 2020-09-09
Sending email to user fails due to bad request handling


kreegahbundolo:
**Environment**:

- CTFd Version/Commit: 3.0.2/5b2926b18b57d25903b16611e7c05ff95b06e857
- Operating System: Ubuntu Linux 18.04
- Web Browser and Version: Firefox 80.0.1

**What happened?**
Tried to send email to a user from CTFd admin. Pressing send results in the message "Email text cannot be empty"

**What did you expect to happen?**
Expected the message "Email sent successfully"

**How to reproduce your issue**
Set up email deliveru (I used setup for gmail smtp and enabled unsecure apps on that account)

**Any associated stack traces or error logs**
No.

**How to fix it**
The python code reads the wrong json request variable "text" instead of "message". This can be seen by viewing the HTML source for the email form. The textarea element containing email body is called "message".

Changed a single line of code:

Old: CTFd/api/v1/users.py:455:        text = req.get("text", "").strip()
New: CTFd/api/v1/users.py:455:        text = req.get("message", "").strip()

Restarted CTFd and now it works.

ColdHeat:
Thanks for filing this! 

I think the best fix is actually to change the underlying form so that we don't have to change the API. 

https://github.com/CTFd/CTFd/blob/master/CTFd/forms/email.py#L8-L10
https://github.com/CTFd/CTFd/blob/master/CTFd/themes/admin/templates/modals/mail/send.html#L4-L5

If you're interested you can PR the fix.

kreegahbundolo:
Added a PR, continuing the discussion there.

PR #1640, opened 2020-09-09
Fix for email send issue

kreegahbundolo:
Fixing issue #1639 , bug that prevented email from being sent.
Applied suggested fixes from @ColdHeat , tested it on my setup, works for me.


ColdHeat:
Thanks for reporting this and fixing it!

PR #1657, opened 2020-09-19
Add the ability to override the sender header of email via SMTP

### LINKED ISSUE #1644: Add email sender address override
Right now the email sender is set to the From address which isn't right in all situations. We need a way to override the email sender if it's not supposed to be the same as the From address. 

https://help.mailgun.com/hc/en-us/articles/202236494-What-is-the-difference-between-the-From-and-Sender-
https://stackoverflow.com/questions/4367358/whats-the-difference-between-sender-from-and-return-path

ColdHeat:
* Add the ability to override the sender header of email via SMTP with the `MAILSENDER_ADDR` config value
* Closes #1644


CTFd/utils/email/smtp.py
```diff
@@ -56,7 +57,7 @@ def sendmail(addr, text, subject):
         msg["From"] = mailfrom_addr
         msg["To"] = addr
 
-        smtp.send_message(msg)
+        smtp.send_message(msg, from_addr=mailsender_addr)
```
ColdHeat, line 60: This is incomplete because it can't be safely overriden by the admin panel.

PR #1670, opened 2020-09-27
Update dependencies

cbk914:


ColdHeat:
Looks like this updates a bunch of dependencies. Is there some kind of benefit to these dependencies?

ColdHeat:
This also seems to have other older dependencies in it, so I'm guessing this PR probably wasn't intended to be opened?

ColdHeat:
I'm going to close this because it's not clear what the goal is. If you want to upgrade dependencies, I'd prefer it was done more piece-meal. One PR per upgraded feature if possible. Upgrading everything at the same time is likely to break something.

PR #1672, opened 2020-09-28
i18n localization (JavaScript)

GregorSpagnolo:
Simple JavaScript localization. 
In this way, we don't affect any backend functionality.


ColdHeat:
Related to #570 

Thanks for writing this up. This looks like a pretty good PR and it's similar to #1083. The JS that you're vendoring would need to be added into `package.json` and properly imported by the main JS. Also though I'm unsure of which library or translation format to go with. I think I had some earlier discussion/investigation in #570.

GregorSpagnolo:
I added inside package.json. 
I have also read other issues but as almost every translation program supports *.json files I would vote for them.

Issue #1667, opened 2020-09-26
Error with docker RuntimeWarning: greenlet.greenlet size changed


whoo:
**Environment**:

- CTFd Version/Commit: ed64b48
- Operating System: (docker-compose build / docker-compose up)
- Web Browser and Version:

**What happened?**
ctfd_1   | INFO  [alembic.runtime.migration] Context impl MySQLImpl.
ctfd_1   | INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
ctfd_1   | [2020-09-26 00:19:26 +0000] [104] [INFO] Booting worker with pid: 104
ctfd_1   | /usr/local/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
ctfd_1   |   return f(*args, **kwds)
ctfd_1   | INFO  [alembic.runtime.migration] Context impl MySQLImpl.
ctfd_1   | INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
ctfd_1   | [2020-09-26 00:19:28 +0000] [106] [INFO] Booting worker with pid: 106
ctfd_1   | /usr/local/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
ctfd_1   |   return f(*args, **kwds)
ctfd_1   | INFO  [alembic.runtime.migration] Context impl MySQLImpl.
ctfd_1   | INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
ctfd_1   | [2020-09-26 00:19:29 +0000] [108] [INFO] Booting worker with pid: 108

**What did you expect to happen?**

**How to reproduce your issue**

**Any associated stack traces or error logs**

Fix:
I've add "greenlet==0.4.16" in requirements.txt to solve this issue.

PR #1677, opened 2020-09-30
Bump gevent version

TheSpeedX:
The gevent version 20.5.2 is not compatible with greenlet generating a Runtime error and not letting the CTFd start. 

This PR fixes Issue #1667 
This was fixed by gevent version bump.The version 20.9.0 works fine.


ColdHeat:
Thanks for the explanation and PR here.

PR #1696, opened 2020-10-23
Binary Compatibility Issue

xFaraday:
Added greenlet to requirements.txt to fix this problem when running a local 'docker-compose up'


ColdHeat:
Thanks for the PR but this was fixed with #1722 which pins all the dependencies in use.

PR #1699, opened 2020-10-26
reduce ping events (Fix #1685)

### LINKED ISSUE #1685: Unnecessary ping event
**Environment**:

- CTFd Version/Commit: 3.1.1, latest commit
- Operating System: any
- Web Browser and Version: any

in the comment you said "Immediately yield a ping event to force Response headers to be set", but this event seems to lies inside the while True loop, which results to an unnecessary ping event every 5 seconds.
I believe that's an unintended behavior, though it doesn't break anything.

https://github.com/CTFd/CTFd/blob/4c31dc23e8cfa0308367732d603b16e01871b00e/CTFd/utils/events/__init__.py#L57-L67

frankli0324:


ColdHeat:
This seems reasonable. Doesn't this also need to modify the other event manager?

frankli0324:
oh

Issue #1379, opened 2020-05-05
Uploading Large Files to CTFd


danielgarner64:
This doesn't necessarily have to be fixed but I think it is useful to be attached to the project for people who run into this issue.

**What happened?**
Uploading a file greater than 100MB causes the upload to hang indefinitely.
This was done through selecting the Challenge through the admin view selecting files and uploading.
I also tested it through an API upload using python requests package.

**Cause**
The Cause of the problem was that the Gunicorn thread was killed during the upload process.
This is because I was using the default syncronous threads as my worker class for Gunicorn.

**What did you expect to happen?**
If the upload timeout the UI should display an error message that connection was lost to the server.

**How to reproduce your issue**
Attempt to upload a file greater than 100MB to a CTFd server with the default configuration.

** Fix I applied **
In the docker-entrypoint.sh I added the following 3 lines
  --worker-class "gthread" \
  --keep-alive 10 \
  --threads 5

This allows the worker to report back to the Gunicorn master on a separate thread and notify the status of the thread (it is still alive just processing a large file upload.)

I suggest as a minimum documenting that CTFd will limit file size due to Gunicorn killing workers that take longer than 30seconds to upload a file.

**Environment**:
  - CTFd Version/Commit:
  - Operating System: Ubuntu 18.04
  - Web Browser and Version: Firefox 75.0

danielgarner64:
So when I did these uploads I primarily used the Restapi (using python code to encode and uplaod the file).
And for very large files 2GB+ I used a very large timeout (in the hours range) it was seomthing like --timeout 6000

1podracer1:
> This doesn't necessarily have to be fixed but I think it is useful to be attached to the project for people who run into this issue.
> 
> **What happened?** Uploading a file greater than 100MB causes the upload to hang indefinitely. This was done through selecting the Challenge through the admin view selecting files and uploading. I also tested it through an API upload using python requests package.
> 
> **Cause** The Cause of the problem was that the Gunicorn thread was killed during the upload process. This is because I was using the default syncronous threads as my worker class for Gunicorn.
> 
> **What did you expect to happen?** If the upload timeout the UI should display an error message that connection was lost to the server.
> 
> **How to reproduce your issue** Attempt to upload a file greater than 100MB to a CTFd server with the default configuration.
> 
[...]

PR #1702, opened 2020-10-27
Add TIMEOUT environment variable

syn-bit:
Add support for a timeout variable for gunicorn
to prevent WORKER timeouts when importing a CTFd export on a
slow platform like a synology nas.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
+TIMEOUT=${TIMEOUT:-30}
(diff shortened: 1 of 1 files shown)


syn-bit:
This PR will also (partly?) solve issue #1379

SmylerMC:
There is indeed a need to increase the Gunicorn timeout value when uploading very large files to S3, to avoid the worker being killed.

An alternative to this PR is simply to use the `GUNICORN_CMD_ARGS` environment variable to add Gunicorn arguments. Setting it to `--timeout=X` effectively sets the  timeout to X seconds.

PR #1707, opened 2020-11-06
Updated vagrant with slightly newer guest os

muttiopenbts:
On first clone and run, many errors were reported with software dependencies and base os defaulted to python2.
After many hours of testing, I was able to get ctfd to work with this base os and python3 default.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,11 +1,19 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+BOX_IMG = "generic/ubuntu1810"
+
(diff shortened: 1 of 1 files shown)


ColdHeat:
I think this looks pretty reasonable. Is it possible to use https://app.vagrantup.com/ubuntu/boxes/bionic64 instead because it seems to be a more official source of Ubuntu?

muttiopenbts:
> I think this looks pretty reasonable. Is it possible to use https://app.vagrantup.com/ubuntu/boxes/bionic64 instead because it seems to be a more official source of Ubuntu?

Tried and failed when running "vagrant ssh".

PR #1719, opened 2020-11-16
Vagrant Ubuntu 18.04

muttiopenbts:
Changed vagrant base box image to Ubuntu 18.04.
18.10 might not be reliable enough.


ColdHeat:
> # BOX_IMG = "ubuntu/bionic64" # psycopg2 problem

Hmm I think I've also run into a psycopg2 installation issue but I suspect it can be avoided by removing that dependency and relying on `psycopg2-binary`. I'll make that change in master shortly.

ColdHeat:
See #1720. I rebased your PR on master. I suspect `ubuntu/bionic64` will work now.

ColdHeat:
Just so you know you can create a branch and then PR against that branch so you don't have to keep deleting PRs and recreating them. This looks like a good tutorial: https://opensource.com/article/19/7/create-pull-request-github

muttiopenbts:
Having trouble ssh. ssh pubkey authc fails.

PR #1724, opened 2020-11-17
New Vagrant file

muttiopenbts:
Based on Official Ubuntu Bionic.

Tested:
vagrant up
vagrant ssh
and 
http://localhost:8000/

Working.


Vagrantfile
```diff
@@ -1,18 +1,32 @@
...
+python -m pip install --upgrade pip
+python -m pip install virtualenvwrapper
 
 # As per instructions at https://downloads.mariadb.org/mariadb/repositories
 apt-get install -y software-properties-common
 apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0xF1656F24C74CD1D8 2>&1
 add-apt-repository -y 'deb [arch=amd64,arm64,i386,ppc64el] http://mirror.lstn.net/mariadb/repo/10.4/ubuntu xenial main'
 apt-get update
 apt-get install -y mariadb-server
+apt-get install -y tmux virtualenvwrapper
```
ColdHeat, line 29: This code probably works but it seems to be running more apt-gets calls than necessary. Can you reorder it so that fewer apt-get calls are made? Perhaps by moving your changes to beneath the mariadb installation and consolidating all the installation commands into one apt-get command.

Vagrantfile
```diff
@@ -56,6 +70,10 @@ Vagrant.configure("2") do |config|
   # and docker or gunicorn (8000) to host machine
   config.vm.network "forwarded_port", guest: 4000, host: 4000
   config.vm.network "forwarded_port", guest: 8000, host: 8000
+  config.vm.synced_folder ".", "/vagrant"
+  
+  # Fix ssh problem
+  config.vm.provision :shell, :inline => "sudo sed -i 's/PasswordAuthentication no/PasswordAuthentication yes/g' /etc/ssh/sshd_config; sudo systemctl restart sshd;", run: "always"
```
ColdHeat, line 76: I feel like this shouldn't be necessary. Don't most vagrant images come pre-installed with vagrant's default SSH key?

ColdHeat (changes requested):
Sorry for the delay. I think this code looks good and probably works but I have some nitpicks about things before merging.

PR #1728, opened 2020-11-19
No longer set "filesystem" as UPLOAD_PROVIDER in config.ini

laurrentt:
Setting `filesystem` here actually prevent the config to be overwritten by an environment variable. This prevent stateless setup such as running CTFd from the official docker image.

This is not a breaking change since the `UPLOAD_PROVIDER` config already has `filesystem` as fallback in `CTFd/config.py` and `CTFd/utils/uploads/__init__.py`.

PS: I've used Github to create the patch and it automatically added a new line at the end of the file, I haven't removed it as it is best practise anyway.


ColdHeat:
Seems legit. Thanks for the PR.

Issue #1706, opened 2020-11-05
None initiated a confirmation email resend


davidhcefx:
**Environment**: 

- CTFd Version/Commit: `3.1.1`
- Operating System: `Linux linux 5.4.0-52-generic #57~18.04.1-Ubuntu SMP Thu Oct 15 14:04:49 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

- Web Browser and Version: `Google Chrome  86.0.4240.183`

**What happened?**
When I pressed the `Resend` button for resending a verification email, CTFd logged a line (in `logs/registrations.log`) like this:
```
[11/05/2020 16:12:32] 10.0.2.2 - None initiated a confirmation email resend
```
From my inspection, `session` in `utils/logging/__init__.py` didn't contain a key called 'email', as well as key 'name' and other stuff.

**What did you expect to happen?**
The `None` been replaced by my email which I used to register. If this is not a bug, then I apologize for the inconvenience.

**How to reproduce your issue**
- Start CTFd with `docker-compose up`.
- Register an account and press the `Resend` button for email verification.

**Any associated stack traces or error logs**
No

alperb:
In addition to the current issue, registrations processes are also not interpreted correctly in the logs.
Here's a log from `logs/registrations.log`:
`[11/22/2020 12:32:03] 127.0.0.1 - None registered with None`

Issue #844, opened 2019-01-21
Rewrite application logging


ColdHeat:
These logs are a holdover from before CTFd was even public. I have significant doubts they're even used and in many situations they're stuck inside of a transient docker container. @tecknicaltom if you do make use of these logs it would be useful to understand how. 

First thoughts:
 - Get these logs off of the fs by default and into something more inline with how docker does logs 
 - Take better advantage of log levels throughout the application

tecknicaltom:
I was developing a plugin related to login and was looking for the log messages to help to debug it. I didn't use the log files, but since the `log()` function prints as well as writes to file, the messages make their way to the container's log. I wasn't tied to the log files on disk, just wanted to make sure that the points in the code where I saw `log()` called were reached. Looking back, I'm not even sure why I discovered that `init_logs()` wasn't called, because I assume now that the message would have made it into the docker log regardless of if the handler was installed or not cause the print is done right in `log()`.

tl;dr: I've got no use case that would prevent you from removing those log files and refactor the logging code

ColdHeat:
We likely want to rewrite logging to be in JSON versus in the currently unstructured text data. See https://github.com/CTFd/CTFd/pull/2580

PR #1734, opened 2020-11-22
Fixed registration and confirmation logs

alperb:
As the problem stated in issue #1706, it was just a problem of format function. Tested and working!


ColdHeat:
Thanks for the PR but I don't think this fix is entirely correct. It looks like the core issue is that the session no longer has the name and email attributes because those aren't being exposed in the session anymore. 

https://github.com/CTFd/CTFd/blob/da4ee4021afae4a98173ed33b15efc997c402b68/CTFd/utils/logging/__init__.py#L12-L18
https://github.com/CTFd/CTFd/blob/da4ee4021afae4a98173ed33b15efc997c402b68/CTFd/utils/security/auth.py#L14-L20

Can you update this PR to remove the old data? And perhaps have a test for the functionality?

ColdHeat:
You might also be interested in https://github.com/CTFd/CTFd/issues/844.

PR #1767, opened 2020-12-30
Speed up docker builds

acepace:
This change first copies and sets up the python requirement file and only then copies the rest of the working directly.

This change can speed up docker rebuilds in a dev or CI environment as code changes that do not impact requirements.txt will not force redownloading all the python dependencies.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,9 +14,12 @@ RUN apt-get update \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-COPY . /opt/CTFd
+COPY requirements.txt /opt/CTFd/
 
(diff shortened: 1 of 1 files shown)


acepace:
Note dockerfile linting fails because of an internal CI error rather than any problem. A manual check with an online linter passed

acepace:
@ColdHeat any idea why linting fails in the CI?

ColdHeat:
Sorry I've been behind on reviewing PRs. I'm trying the lint step again. The error it gave wasn't helpful but I think it was probably something transient. 

I think your PR looks good and is probably the right approach b/c it will cache pip deps. But let me double check.

PR #1771, opened 2021-01-07
feat: Add option to fallback to the default theme

maybe-sybr:
This is intended to make it easier to develop themes in pieces rather
than having to copy the default one over and modifying it in place, or
having to create all pages before it will work sensibly. It may not also
be easier to construct minor modifications to the core theme which
aren't easy/possible with the pages mechanism or jinja template block
overrides.


maybe-sybr:
Hi CTFd folks! I've been working on some theming for an event and wanted to avoid copy pasting the core theme so I wrote this instead. I'll fix up the linter complaints now but let me know if you'd be happy to merge this or if you'd like to see any changes before considering further.

I'll note upfront the the introduction of the `CTFd.defaults` module is a bit clunky at the moment but unfortunately to eliminate all references to the string `"core"` in the codebase I couldn't jam it in `CTFd.utils.config` as a constant because that caused circular imports. Hopefully it's a trope you'd be happy to move forward with.

ColdHeat:
I haven't looked through your implementation yet but ruminating over it for awhile and I think this is a step in the right direction. We should view the core theme as the fallback/default and themes should override behavior as necessary.

ColdHeat:
Also `defaults` likely is better suited in the `constants` directory.

maybe-sybr:
@ColdHeat - I've moved the `DEFAULT_THEME` definition into `CTFd.constants`. It didn't have an obvious home AFAICT so I made `CTFd.constants.theme` rather than putting it in `CTFd.constants.config` which had some theme config bits already but didn't feel like the right place.

ColdHeat:
I'm going through this code now and I still like the idea. A few changes that I think I will make shortly:

- Move theme fallback into a `config.py`/`config.ini` level configuration. I don't think there's much need for this in the Admin Panel. Really feels like more of a development feature than one admins need to manipulate via UI, but maybe that will change in the future.
- There's gotta be a better place for `DEFAULT_THEME` than `CTFd/constants/themes.py`...
- The theme fallback code seems a little too much to understand. Maybe it's just me. I'll try to simplify it down.

ColdHeat:
I think silently returning the default theme data could lead to some confusion. I wonder if it's perhaps better to 302 to the right file instead?

ColdHeat:
I think so long as THEME_FALLBACK defaults to false, we can just do the fallback normally without the 302.

ColdHeat:
I cannot seem to improve upon the for loop for theme fallback... I don't like the pattern but I guess it's not really that bad?

maybe-sybr:
Those points all sound valid. I like the use of a 302 for static assets.

Making `THEME_FALLBACK` and config.ini is a great idea, it would definitely be fit for purpose since this would probably only be useful during dev or when the user knows they need it for the lifetime of the server because their theme is deliberately incomplete.

CTFd/__init__.py
```diff
@@ -114,10 +117,18 @@ def get_source(self, environment, template):
             template = "/".join(["admin", "templates", template])
             return super(ThemeLoader, self).get_source(environment, template)
 
-        # Load regular theme data
-        theme = str(utils.get_config("ctf_theme"))
-        template = "/".join([theme, "templates", template])
-        return super(ThemeLoader, self).get_source(environment, template)
+        # Load regular theme data with potential fallbacks
+        first_cand, *others = ctf_theme_candidates()
```
ColdHeat, line 121: Is there a reason to need to break this out into two lists? I think `themes = list(ctf_theme_candidates())` would work?

ColdHeat: Ah I see it's for the explicit exception!

PR #1773, opened 2021-01-08
Update docker

khiemdoan:


ColdHeat:
What's the purpose of this here? Just to bump versions?

khiemdoan:
@ColdHeat Yes, and one more purpose. Because we use python 3, we have to install python3-dev. The package python-dev is used with python 2 correspondingly.

docker-compose.yml
```diff
@@ -52,7 +52,7 @@ services:
     command: [mysqld, --character-set-server=utf8mb4, --collation-server=utf8mb4_unicode_ci, --wait_timeout=28800, --log-warnings=0]
 
   cache:
-    image: redis:4
+    image: redis:6
```
acepace, line 55: No breaking changes?

khiemdoan: I have tested it without any problem. Furthermore, the library redis-py is developed against redis 6.0.9 now.

ColdHeat (changes requested):
I'm not a fan of updating version numbers between major revisions without some kind of benefit. The MariaDB version was chosen specifically b/c there was some issue on later versions. Perhaps this is fixed but I'm not sure it's worth it. 

I think I'd only be okay with accepting the `python3-dev` change. 

The rest of the changes imo don't really represent enough value. Seems like its just bumping versions for the sake of it. And if you really want to use those higher version numbers you can always use them in your own docker-compose. 

If there's a specific reason for some of these version bumps/changes let me know and I'll reevaluate.

khiemdoan:
@ColdHeat OK, I have updated this PR.

Issue #866, opened 2019-02-01
Linux Headers missing from Dockerfile


nimnaij:
**Environment**:

  - CTFd Version/Commit: latest ( file in question is https://github.com/CTFd/CTFd/blob/master/Dockerfile found in commit https://github.com/CTFd/CTFd/commit/c8031b38c27926b8166174af331bc2b4ab614c1b )
  - Operating System: Ubuntu 18.04
  - Web Browser and Version: N/A

**What happened?**
When running `docker compose build` with a plugin that has additional python pip packages that require linux headers to install, pip fails to install the additional package.

**What did you expect to happen?**
I expected the plugin to install correctly. 

**How to reproduce your issue**
Add this package to a requirements.txt, either in the plugin or in the main requirements:
`python-openstackclient`
(see: https://git.cybbh.space/ben/ctfd_openstack_integration/blob/master/requirements.txt ) 

**Any associated stack traces or error logs**
netifaces.c:131:28: fatal error: linux/tipc.h: No such file or directory pip

**Proposed solution**
Modify Dockerfile from:
`apk add python python-dev libffi-dev gcc make musl-dev py-pip mysql-client git openssl-dev`
to
`apk add python python-dev libffi-dev gcc make musl-dev py-pip mysql-client git openssl-dev linux-headers`

I can fork and make a PR if necessary.

ColdHeat:
In general, the Dockerfile's purpose is not to have enough to build every single plugin. Developers are free to create child Dockerfiles that contain what they need. 

I feel like linux-headers may be generic enough that this would warrant an exception but I'm not sure yet.

PR #1782, opened 2021-01-26
[closes #1781] reduce docker image size

### LINKED ISSUE #1781: Dockerfile improvements
I recently see that there are some improvements to the Dockerfile, but there's a little more to do

#### 1. requirements.txt and caching

not only the main requirements.txt can be cached, but also the plugin requirements. this can be done with a "multi-stage" build.

```Dockerfile
# stage 1
FROM python:3.7-slim-buster
COPY . /somewhere
RUN <remove everything let alone requirements.txt>

# stage 2
FROM python:3.7-slim-buster
COPY --from=0 /somewhere .
# now there are only requirements.txt in the folder, and will be cached on next build.
```

#### 2. python:3.7-slim-buster and python3-dev

the image `python:3.7-slim-buster` doesn't contain a python3 package

```cli
root@7d5d3c29ec30:/# apt list --installed | grep python

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

root@7d5d3c29ec30:/# 
```

it copies a custom built python binary into the image instead. you can see that in the dockerhub build history

image

however, python3-dev depends on that python3 package, so if you `apt install python3-dev`, you are basically installing a second copy of the python3 binary.

I would suggest building from the debian:buster-slim image, but maybe there are better ways.

#### 3. dev headers

furthermore, the dev headers are somewhat not needed in the running container. so maybe the `pip install` should go to a building stage, leaving the `site-packages` folder to the actual env

--------

I'll try to implement this, if anyone have better ideas, please let me know

frankli0324:
1. caches plugin requirements.txt for faster re-building
2. removes dev headers and git for they are only required for the alpine base image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,32 @@ FROM python:3.7-slim-buster AS cache
 COPY . /opt/CTFd
 RUN find /opt/CTFd -not -name "requirements.txt" -type f -delete
 
+FROM python:3.7-slim-buster AS builder
+
+COPY --from=cache /opt/CTFd /opt/CTFd
+WORKDIR /opt/CTFd
+
+# hadolint ignore=DL3008
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        build-essential \
+        python3-dev \
+        libffi-dev \
+        libssl-dev \
+        git \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# hadolint ignore=SC1091
+RUN python -m venv /opt/venv && \
+    . /opt/venv/bin/activate && \
+    pip install -r requirements.txt --no-cache-dir && \
+    for d in CTFd/plugins/*; do \
+        if [ -f "$d/requirements.txt" ]; then \
+            pip install -r "$d/requirements.txt" --no-cache-dir; \
+        fi; \
+    done;
+
 FROM python:3.7-slim-buster
 WORKDIR /opt/CTFd
 RUN mkdir -p /opt/CTFd /var/log/CTFd /var/uploads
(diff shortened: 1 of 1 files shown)


ColdHeat:
This seems wrong. Having those deps helps with building requirements. Maybe now with more wheels it's less needed but there are still often deps that need to be built which libffi at least helps with. 

I'm also not sure if I agree with pulling the plugin requirements install out of the main image. 

I'll investigate whether this is useful or not but I suspect it's not too beneficial given the trade offs.

frankli0324:
I didn't catch up, how is there going to be more deps?

even if there are more deps, I really can't imagine a dependency for an app like this needs building...
on almost every scenario the python package is written in pure python, except for low level libraries like the socket handling gvent library.

moreover, wheel is the `standard` for python package distribution since almost ten years ago, I don't think there will be more packages without wheel package.

I think what you need is the runtime library, instead of the dev headers.

I promise, I thought through this

frankli0324:
I didn't catch up, how is there going to be more deps?

even if there are more deps, I really can't imagine a dependency for an app like this needs building...
on almost every scenario the python package is written in pure python, except for low level libraries like the socket handling gvent library.

moreover, wheel is the `standard` for python package distribution since almost ten years ago, I don't think there will be more packages without wheel package.

I think what you need is the runtime library, instead of the dev headers.

I promise, I thought through this

Dockerfile
```diff
@@ -1,32 +1,32 @@
+FROM python:3.7-slim-buster AS cache
+
+COPY . /opt/CTFd
+RUN find /opt/CTFd -not -name "requirements.txt" -type f -delete
```
frankli0324, line 4: # delete everything **except** requirements.txt, so when actually copied to the second stage, docker will get the same hashes even if the plugin source has changed, as long as the requirements.txt doesn't change. the packages still remains in the "main image", if I understanded your comment correctly.

frankli0324:
or how about venv? I'm going to have a try

frankli0324:
by adding this extra stage you can have the dev headers available for pip install, keeping the final image small. the binaries built usually stays in the virtual env.

maybe, maybe you would add some runtime libraries to the apt install list in the final stage

the final line is somewhat ugly.

ColdHeat:
I don't think this approach is going to work. Maybe it reduces the image size but it reduces capability as well. See issues like #866. I know for a fact that it will break some CTFd deployments because plugins currently have no way of adding in additional dependencies via apt/apk. 

Wheels might be a standard, but that doesn't mean every package has a wheel or is pure Python. See https://pythonwheels.com/. 

I'll leave your issue open b/c there could be improvements made to the Docker image but I don't think this PR is going down the right path. 

If the image is too big for your needs, you're always free to build your own Docker image.

frankli0324:
```Dockerfile
FROM ctfd/ctfd
RUN apt update && \
    apt install runtime libraries needed && \
    remove cache
```

while keeping all the dev headers that are already there. I don't think the last approach have any trade offs at all.

well then, PR closed

Issue #1786, opened 2021-01-28
Report errors when data is too long


ColdHeat:
The API endpoints need more validation. I suspect this is at least partially due to Marshmallow which overall is a bad library.

ColdHeat:
For the record, the size of a MySQL TEXT appears to be 64KB.

PR #1787, opened 2021-01-28
Add length error content that is too long

ColdHeat:
* Add length checking to some sensitive fields in Pages and Challenges.
* Works on #1786

This is enough to fix most of the issues but this is really a systemic problem for most of the API endpoints. We should have something that verifies data consistency. Marshmallow is not good enough at this. Pydantic seems like it would be superior here.


ColdHeat:
Merging this even without full test coverage because there shouldn't be any kind of break in functionality. Although thinking about it there should be a few tests but I think I will leave that for a more systemic solution.

frankli0324:
There **is** length validation in marshmallow

https://marshmallow.readthedocs.io/en/stable/marshmallow.validate.html?highlight=validate.Length#marshmallow.validate.Length

ColdHeat:
I believe that would require adding the field explicitly in the schema instead of relying on the model.

frankli0324:
you are not relying on the model anyway...😓

frankli0324:
Could be combined with https://stackoverflow.com/questions/1777814/how-to-get-sqlalchemy-length-of-a-string-column

frankli0324:
image
this is ur User schema

and maybe you are looking for ma.SQLAlchemyAutoSchema

frankli0324:
https://github.com/frankli0324/CTFd/commit/16d8f1c27c302717c3653b402f9478369e95d610

at least for the sake of code consistency

ColdHeat:
Ah okay, I apologize. I misunderstood what you meant but if I was honest I don't like Marshmallow and the code will eventually be removed. 

If you'd like you can PR the field based solution.

PR #1795, opened 2021-02-07
improv: Avoid 500 when error templates are missing

maybe-sybr:
By not considering that error templates might be missing, it's possible
for a 500 to leak out when we're rendering a more innocuous error like a
404. This change reworks the error handlers into a single function and
avoids misbehaviour when the template being rendered is not found.


maybe-sybr:
I also noticed that we have a template for 429 in the core theme but don't appear to actually register it

ColdHeat:
I think this leak would only happen if you were using the debug server right? I'm not sure this is necessary if that's the case

maybe-sybr:
You're right, I'm pretty sure I was using the flask/werkz server when I saw this. I think I'd still want the server to return the actual code I hit (e.g. 404) than a 500 because it is also missing the 404 template.

CTFd/__init__.py
```diff
@@ -271,10 +271,8 @@ def set_sqlite_pragma(dbapi_connection, connection_record):
 
         app.register_blueprint(admin)
 
-        app.register_error_handler(404, page_not_found)
-        app.register_error_handler(403, forbidden)
-        app.register_error_handler(500, general_error)
-        app.register_error_handler(502, gateway_error)
+        for code in {403, 404, 500, 502}:
+            app.register_error_handler(404, render_error)
```
ColdHeat, line 275: I think this should be code right?

ColdHeat:
This does seem to affect deployments under Gunicorn so I believe this code is good to merge. It avoids errors when themes freak out.

CTFd/__init__.py
maybe-sybr: Yes, of course. ^c^v fail on my part! Thanks for correcting it

PR #1806, opened 2021-02-19
Experiment with a faster /api/v1/scoreboard endpoint

ColdHeat:
* Experiment with a faster way to generate the scores of all members in the /api/v1/scoreboard endpoint

### DIFF
--- a/CTFd/api/v1/scoreboard.py
+++ b/CTFd/api/v1/scoreboard.py
@@ -55,7 +55,7 @@ class ScoreboardList(Resource):
                     }
 
             # Get user_standings as a dict so that we can more quickly get member scores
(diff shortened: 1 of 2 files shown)


ColdHeat:
I was able to get very significant speed ups with this approach. 20-30x speedups (sometimes 60x) with very large datasets. Reviewing the code the reason is clear that previously we were forcing the database to have to look at every row. 

```
(ctfd3) ❯ curl -o /dev/null -s -w 'Total: %{time_total}s\n' http://localhost:4000/api/v1/scoreboard
Total: 2.181876s

(ctfd3) ❯ curl -o /dev/null -s -w 'Total: %{time_total}s\n' http://localhost:4000/api/v1/scoreboard
Total: 62.941061s
```

PR #1808, opened 2021-02-20
improv: Add blocks to the core challenges template

maybe-sybr:
This will allow for more flexible injection of content into the
challenge page from other templates.


ColdHeat:
I like this idea but I'm confused a bit on how you use it? 

Are you trying to create new different types of challenge pages within the same theme or are you trying to extend the core theme?

I think the latter makes sense and is the most interesting but it doesn't seem to work in my local testing with things like:
```
{% extends "core/challenges.html" %}
```

ColdHeat:
For example what I'm thinking is if you add the following code to ThemeLoader.get_source():

```
        if template.startswith("core/"):
            template = template[5:]  # Strip out admin/
            template = "/".join(["core", "templates", template])
            return super(ThemeLoader, self).get_source(environment, template)
```

you can do things like:

```
{% extends "core/challenges.html" %}

{% block challenge_footer %}
Testing out this footer
{% endblock %}
```

Which would probably help out with making child themes.

Is this what you're going for?

maybe-sybr:
I've been working on a plugin for our event which adds some UI elements to the challenge page in the content div. Because the challenges template only allows you to override the entire content block, that was going to be a pain. One option I did consider was adding a tail element and some onready JS to reparent it in the right spot but that felt disgusting.

Here's a rough example of usage:

```py
# plugins/my_plugin/__init__.py
from flask import Blueprint, render_template

bp = Blueprint("my_plugin", __name__, template_folder="templates")

def view_challenges():
    return render_template("my_challenges.html")

def load(app):
    app.register_blueprint(bp)
    # The format used by the view_functions dictionary is blueprint.view_function_name
    app.view_functions["challenges.listing"] = view_challenges
```

```jinja2
{# my_challenges.html #}
{% extends "challenges.html" %}
{% block challenges_header %}
{{ super() }}
<div class="row">
  <div class="form-row">
    <div class="col-auto">Here's some content!</div>
  </div>
</div>
{% endblock challenges_header %}
```

As you can see, I'm actually hijacking a route to render my own challenges template, but being able to add to the search path for the theme loader and shadow names of existing templates (can I already do this? where in the template loader search path do new paths added by blueprints end up? append or prepend?) so I can avoid hijacking the route would also be a nice trick.

maybe-sybr:
I've actually hit a perfect use case for being able to refer to the template from a specific theme for inheritance purposes, per your previous comment @ColdHeat. This came up in some of our theming work rather than the plugin work I was thinking of and your example of being able to `{% extends core/challenges.html %}` would make that immensely easier.

If you don't mind, I'd like to make a change along those lines and add it to this PR. I'd think that would also make the case for actually merging this change a lot more compelling from you, since it would be useful for themers as well as plugin (ab)users like me :)

ColdHeat:
Yeah I agree! I also think `{% extends core/challenges.html %}` is a good feature to have. 

The blocks themselves are useful but I think using them in that template overriding strategy is a little too specific. We should still support it but I'll probably need to add blocks for every template.

maybe-sybr:
The theme prefixing got merged in #1817 so this PR remains a small change to add these blocks. I'm still using it to add stuff to the top of the challenge board in a plugin and it's worked fairly well for us so far.

ColdHeat:
I'm playing around with this idea and it's definitely useful but I think those are kinda specific scenarios that might be better suited by some other ideas I have. 

I'm going to leave this open for others to use if they want but take this off of the 3.4 milestone. 

I'm going to release a beta rewrite of the core theme that should make it easier to theme with https://github.com/CTFd/CTFd.js and other improvements. I might include this there but for now I'd rather stick to keeping it something to experiment with than in the main repo.

PR #1812, opened 2021-02-26
feat: Add solve count and status to challenge list

### LINKED ISSUE #1811: Add solve count to challenge list API
It would be nice if we could add the number of solves to the challenge list API so themes could render things like first blood or medal availability. We can currently request the solves for each challenge but that'd be extremely wasteful and since we already make queries for solves to manage challenge requirements, it doesn't seem unreasonable to also add a count on the challenge ID  in the solves table. Something like:

```
solve_counts = {
    chal_id: solves for (chal_id, solves) in db.session.query(
        Solves.challenge_id, sa_func.count(Solves.challenge_id),
    ).group_by(Solves.challenge_id).all()
}
```

We can also do some trickery to wrap up the check for if the current user/team has solved it like so:
```
solve_counts = {
    chal_id: (solves, user_solved) for (chal_id, solves, user_solved) in db.session.query(
        Solves.challenge_id, sa_func.count(Solves.challenge_id),
        sa_func.sum(Solves.account_id == user.account_id),
    ).group_by(Solves.challenge_id).all()
}
```

which would save us from doing more queries later on.

Would you be interested in a PR for this?

maybe-sybr:
This changeset reworks the challenge list API endpoint to count the
number of solves for each challenge and whether the current user (or
their team) have already solved it. The endpoint now return two new keys
per item in the response; the `solved` boolean which shows if the
challenge has been solved already, and the `total_solves` integer which
shows how many solves have been made by all users.

Closes #1811

### DIFF
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -50,7 +50,14 @@ from CTFd.utils.helpers.models import build_model_filters
 from CTFd.utils.logging import log
 from CTFd.utils.modes import generate_account_url, get_model
 from CTFd.utils.security.signing import serialize
-from CTFd.utils.user import authed, get_current_team, get_current_user, is_admin
+from CTFd.utils.user import (
+    authed,
+    get_current_team,
+    get_current_team_attrs,
+    get_current_user,
+    get_current_user_attrs,
+    is_admin,
+)
 
 challenges_namespace = Namespace(
     "challenges", description="Endpoint to retrieve Challenges"
@@ -156,6 +163,16 @@ class ChallengeList(Resource):
         location="query",
     )
     def get(self, query_args):
+        # Require a team if in teams mode
+        # TODO: Convert this into a re-useable decorator
+        # TODO: The require_team decorator doesnt work because of no admin passthru
+        if get_current_user_attrs():
+            if is_admin():
+                pass
+            else:
+                if config.is_teams_mode() and get_current_team_attrs() is None:
+                    abort(403)
+
         # Build filtering queries
         q = query_args.pop("q", None)
         field = str(query_args.pop("field", None))
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -3,7 +3,9 @@ from typing import List
 
 from flask import abort, render_template, request, url_for
 from flask_restx import Namespace, Resource
-from sqlalchemy.sql import and_
+from sqlalchemy import func as sa_func
+from sqlalchemy import types as sa_types
+from sqlalchemy.sql import and_, cast, false, true
 
 from CTFd.api.v1.helpers.request import validate_args
 from CTFd.api.v1.helpers.schemas import sqlalchemy_to_pydantic
@@ -75,6 +77,44 @@ challenges_namespace.schema_model(
 )
 
 
+def _build_solves_query(extra_filters=tuple(), admin_view=False):
+    # This can return None (unauth) if visibility is set to public
+    user = get_current_user()
+    # We only set a condition for matching user solves if there is a user and
+    # they have an account ID (user mode or in a team in teams mode)
+    if user is not None and user.account_id is not None:
+        user_solved_cond = Solves.account_id == user.account_id
+    else:
+        user_solved_cond = false()
+    # We have to filter solves to exclude any made after the current freeze
+    # time unless we're in an admin view as determined by the caller.
+    freeze = get_config("freeze")
+    if freeze and not admin_view:
+        freeze_cond = Solves.date < unix_time_to_utc(freeze)
+    else:
+        freeze_cond = true()
+    # Finally, we never count solves made by hidden or banned users/teams, even
+    # if we are an admin. This is to match the challenge detail API.
+    AccountModel = get_model()
+    exclude_solves_cond = and_(
+        AccountModel.banned == false(), AccountModel.hidden == false(),
+    )
+    # This query counts the number of solves per challenge, as well as the sum
+    # of correct solves made by the current user per the condition above (which
+    # should probably only be 0 or 1!)
+    solves_q = (
+        db.session.query(
+            Solves.challenge_id,
+            sa_func.count(Solves.challenge_id),
+            sa_func.sum(cast(user_solved_cond, sa_types.Integer)),
+        )
+        .join(AccountModel)
+        .filter(*extra_filters, freeze_cond, exclude_solves_cond)
+        .group_by(Solves.challenge_id)
+    )
+    return solves_q
+
+
 @challenges_namespace.route("")
 class ChallengeList(Resource):
     @check_challenge_visibility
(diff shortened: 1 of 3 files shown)


maybe-sybr:
Hi @ColdHeat, I've got the guts of the change here but will need to spend a bit of time bringing tests up to scratch and potentially adding some new ones. I'll also need to duplicate this logic into the `challenges/<id>` endpoint, but can you think of anywhere else I should be aware of? Happy to get some feedback on the changeset as it stands and I'll work on making it mergable over the next couple of days.

CTFd/api/v1/challenges.py
```diff
@@ -123,53 +124,49 @@ def get(self, query_args):
...
         # This can return None (unauth) if visibility is set to public
         user = get_current_user()
+        # Admins get a shortcut to see all challenges despite pre-requisites
+        admin_view = is_admin() and request.args.get("view") == "admin"
 
-        # Admins can request to see everything
-        if is_admin() and request.args.get("view") == "admin":
-            challenges = (
-                Challenges.query.filter_by(**query_args)
-                .filter(*filters)
```
maybe-sybr, line 131: I've accidentally dropped this from the challenge query in its new location. Luckily the linter caught me. I'll re-add it

CTFd/api/v1/challenges.py
```diff
@@ -201,6 +200,8 @@ def get(self, query_args):
                     "type": challenge_type.name,
                     "name": challenge.name,
                     "value": challenge.value,
+                    "solved": challenge.id in user_solves,
```
maybe-sybr, line 203: This will make it possible to avoid an extra query for solves when colouring the challenges. I won't do that in this changeset but definitely something to think about as follow on work.

CTFd/api/v1/challenges.py
```diff
@@ -201,6 +205,8 @@ def get(self, query_args):
...
+                    "total_solves": solve_counts.get(challenge.id, 0),
```
maybe-sybr, line 209: This key needs to be `"solves"` to match the existing behaviour on the `/challenges/<id>` endpoint. That makes me want to avoid using just `"solved"` for the key above since it's very easy to typo or misread `solved` and `solves`. I'm leaning toward `"solved_by_me"`.

CTFd/api/v1/challenges.py
```diff
@@ -123,53 +124,54 @@ def get(self, query_args):
...
-                else:
-                    if config.is_teams_mode() and get_current_team() is None:
-                        abort(403)
-            else:
-                solve_ids = set()
+            user_solved_cond = false()
+        # This query counts the number of solves per challenge, as well as the
+        # sum of correct solves made by the current user per the condition
+        # above (which should probably only be 0 or 1!)
+        solves_q = db.session.query(
```
maybe-sybr, line 139: This doesn't take into account any of the extra logic around hidden or banned solvers which the `/challenges/<id>` endpoint does :( Also I probably need to respect `score_visible() and accounts_visible()`, and freeze time for solve timestamps.

maybe-sybr:
I'm done with this changeset for the moment and would appreciate some review from y'all now, if possible. No real rush, I'm planning on just running off my own branch or picking this changeset onto a stable release in the meantime.

CTFd/api/v1/challenges.py
```diff
@@ -75,6 +76,43 @@ class ChallengeListSuccessResponse(APIListSuccessResponse):
...
+        AccountModel.banned == false(),
+        AccountModel.hidden == false(),
+    )
+    # This query counts the number of solves per challenge, as well as the sum
+    # of correct solves made by the current user per the condition above (which
+    # should probably only be 0 or 1!)
+    solves_q = db.session.query(
+        Solves.challenge_id, sa_func.count(Solves.challenge_id),
+        sa_func.sum(user_solved_cond),
+    ).join(AccountModel).filter(
```
maybe-sybr, line 108: Note that I don't bother joining the challenges table to check for hidden challenges. I'm assuming that the caller will take care of not using data this query returns for challenges which are hidden or don't meet other visibility criteria.

ColdHeat:
Sorry I've been ill the past week or so, so I'm behind on reviewing your PRs but I haven't forgotten about them. Fingers crossed, I'll be able to get to them by next week.

CTFd/api/v1/challenges.py
ColdHeat: I actually like `solved_by_me` a lot

maybe-sybr:
> Sorry I've been ill the past week or so, so I'm behind on reviewing your PRs but I haven't forgotten about them. Fingers crossed, I'll be able to get to them by next week.

All good! I appreciate the attention :)

This change should fix the lint I left. It'll need to be autosquashed prior to merge. The CI test failure also weirded me out but I re-ran `make test` locally and it seemed okay. Let me know if there's anything specific which I broke and how I can repro a failure, and I'll get onto it.

maybe-sybr:
Should this be targeting 3.3.0-dev?

CTFd/api/v1/challenges.py
```diff
@@ -75,6 +76,44 @@ class ChallengeListSuccessResponse(APIListSuccessResponse):
...
+        AccountModel.banned == false(), AccountModel.hidden == false(),
+    )
+    # This query counts the number of solves per challenge, as well as the sum
+    # of correct solves made by the current user per the condition above (which
+    # should probably only be 0 or 1!)
+    solves_q = (
+        db.session.query(
+            Solves.challenge_id,
+            sa_func.count(Solves.challenge_id),
+            sa_func.sum(user_solved_cond),
```
maybe-sybr, line 108: Turns out you can't sum booleans in postgres, that's what's causing the CI failures. I think making it `sum(case({true(): 1}, value=user_solved_cond, false()))` will basically do what I want, but `sum(cast(user_solved_cond, Integer))` may also work

ColdHeat:
I'm starting to go through this and I wanted to thank you for the care that you took while preparing this PR! It's rare to see this kind of quality and just wanted to let you know that. 

It probably should target 3.3.0-dev but it's not a big deal, I can change it if it's necessary.

CTFd/api/v1/challenges.py
```diff
@@ -376,25 +418,19 @@ def get(self, challenge_id):
...
+            maybe_row = solves_q.first()
+            if maybe_row:
+                _, solve_count, solved_by_user = maybe_row
+                solved_by_user = bool(solved_by_user)
+            else:
+                solve_count, solved_by_user = 0, False
         else:
-            response["solves"] = None
-            solves = None
+            solve_count, solved_by_user = None, False
```
ColdHeat, line 433: I feel as if this line is wrong? scores_visible and accounts_visible wouldn't affect whether the user can see if their own challenge is solved?

I haven't tested it myself yet though.

ColdHeat: Yeah solved_by_user should be independent of scores_visible() and accounts_visible() and always point out whether the user solved their challenge.

`/api/v1/users/me/solves`
```json
{
            "type": "correct",
            "challenge_id": 6,
            "user": 2,
            "team": 7,
            "challenge": {
                "value": 150,
                "category": "Forensics",
                "name": "light"
            },
            "date": "2021-02-25T10:00:37.041914+00:00",
            "id": 60
        }
```

compared to:

`/api/v1/challenges/6`
```json
        "solves": null,
        "solved_by_me": false,
```

maybe-sybr: Sure thing, that sounds entirely reasonable. So we'd want the response to say this following in that case?
```
        "solves": null,
        "solved_by_me": true,
```

CTFd/api/v1/challenges.py
```diff
@@ -121,55 +161,51 @@ def get(self, query_args):
...
-                .order_by(Challenges.value)
-                .all()
+            # This is necessary to match the challenge detail API which returns
+            # `None` for the solve count if visiblity checks fail
+            solve_count_dfl = None
+
+        # Build the query for the challenges which may be listed
+        chal_q = Challenges.query
+        # Admins can see hidden and locked challenges in the admin view
+        if not admin_view:
```
ColdHeat, line 191: nit: I prefer these as `if admin_view is False`
```suggestion
        if admin_view is False:
```

CTFd/api/v1/challenges.py
```diff
@@ -201,6 +239,8 @@ def get(self, query_args):
                     "type": challenge_type.name,
                     "name": challenge.name,
                     "value": challenge.value,
+                    "solves": solve_counts.get(challenge.id, solve_count_dfl),
+                    "solved_by_me": challenge.id in user_solves,
```
ColdHeat, line 243: This also does not behave correctly if scores are disabled.

CTFd/api/v1/challenges.py
ColdHeat: It would be dependent on whether the user actually solved the challenge. In the example data I showed above it would say true. 

You probably need to move the solve_q variable out of the if statement and then refer to it from within.

tests/api/v1/test_teams.py
```diff
@@ -860,7 +860,7 @@ def test_api_user_without_team_challenge_interaction():
         gen_flag(app.db, 1)
 
         with login_as_user(app) as client:
-            assert client.get("/api/v1/challenges").status_code == 403
+            assert client.get("/api/v1/challenges").status_code == 200
```
ColdHeat, line 863: Why did this change? 

I don't necessarily disagree but I don't think this PR would be the right place to change this.

CTFd/api/v1/challenges.py
```diff
@@ -121,55 +161,51 @@ def get(self, query_args):
...
-                    .all()
-                )
-                solve_ids = set([value for value, in solve_ids])
-
-                # TODO: Convert this into a re-useable decorator
-                if is_admin():
-                    pass
-                else:
-                    if config.is_teams_mode() and get_current_team() is None:
-                        abort(403)
```
ColdHeat, line 161: The lack of this code is causing you to modify the team check test for 403.

ColdHeat: I re-introduced this code and reverted the change to that teams mode test. Let me know if you felt like it should be 200.

CTFd/api/v1/challenges.py
ColdHeat: Yup I will handle that in a follow up PR

ColdHeat:
I think I was able to fix up the `solved_by_me` field and update the tests appropriately. I think this PR should be okay to merge but I think it makes the most sense in 3.3.0.

maybe-sybr:
This probably needs a rebase to get 843546b out of the changeset which I think is unrelated

CTFd/api/v1/challenges.py
```diff
@@ -182,22 +182,24 @@ def get(self, query_args):
...
-            solves_q = _build_solves_query(admin_view=admin_view)
-            # Aggregate the query results into the hashes defined at the top of
-            # this block for later use
-            for chal_id, solve_count, solved_by_user in solves_q:
-                solve_counts[chal_id] = solve_count
-                if solved_by_user:
-                    user_solves.add(chal_id)
         else:
+            # Empty out the solves_count if we're hiding scores/accounts
+            solve_counts = {}
```
maybe-sybr, line 202: This seems a bit unfortunately wasteful to do :( I'd guess that having scores off isn't that common though and since we need to find any solve for every challenge ID, making a simpler query to just get a value for `solved_by_me` seems like it's not worth the effort.

seems legit!

maybe-sybr:
The changes on you've put up look fine to me @ColdHeat - thanks for doing that! I'm happy with this as it stands if you're keen to merge this now

CTFd/api/v1/challenges.py
ColdHeat: Yeah I think double query isn't worth it but you're right it's not an extremely common configuration. 

I think I'll just see if it ends up causing problems and I'll optimize further when it gets there.

ColdHeat:
Yeah admittedly I'm not that great at git but I will try to clean up the changeset. I'll give it a shot and merge it shortly.

maybe-sybr:
Thanks for the merge @ColdHeat :) Much appreciated

PR #1816, opened 2021-03-08
feat: Allow challenge types to override visibility

maybe-sybr:
The goal of this change is to support plugins which implement "stealth"
challenges which are not shown to users until they have been solved.
Such challenge types would need some way for a user to discover the
challenge IDs to make use of the traditional challenge attempt API
endpoint, or have a paired endpoint which allows the submission of flags
without specifying a challenge ID. It is expected that these details
should be implemented entirely by plugins with CTFd simply doing what it
can to consult them where reasonable.

The change introduces a property on the base challenge type class which
is used as a final visibility check prior to data for that challenge
being returned to a user. In particular, the challenge list and detail
API endpoints have been taught to check this prior to returning data.
The attempt endpoint continues to rely solely on `.attempt()` so that
challenge types which refuse to be shown to the user may still be
attempted unless the challenge type also disallows that (in which case,
what's the point?).


maybe-sybr:
I'm using this to implement a class of "stealth" challenges which would be visible only after having been successfully solved by players. The plugin is also going to implement a variant of the `/attempt` endpoint which will check if the submission is correct for any challenge not currently hidden, locked or blocked by requirements. This is just the enabling feature but that plugin will go public along with the rest of our code and I'll think about upstreaming it if you care for it at that point. Just wanted to provide some context for how I'm actually making use of this change.

maybe-sybr:
Just rebased this back onto master after the 3.3.0 tag, should be mergeable again without the old stuff from 3.3.0-dev

frankli0324:
seeing this I think CTFd needs a documented abstraction of a "challenge", such as challenge methods that can be overridden, necessary properties.

maybe-sybr:
> seeing this I think CTFd needs a documented abstraction of a "challenge", such as challenge methods that can be overridden, necessary properties.

The `BaseChallenge` kind of is that, I suppose. I do think something that would be worth considering for 4.0 would be a revisit of the CTFd data model to see what's be useful and what isn't. Specifically, I think that having the base challenge implementation be plugin based is more annoying that it is helpful because ultimately the only practical way to make challenge extensions is to subclass or modify that base class, and also grok how that interacts with the `Challenges` model which is where a bunch of the stuff one might care to override actually lives. It's all possible with the current layout, it just feels pretty complicated.

frankli0324:
I mean docs...

Issue #1194, opened 2019-12-23
Theme preview for admins


ColdHeat:
A way to preview a theme as an admin would be useful.

ColdHeat:
I think perhaps we could re-use the preview page from the Page Editor.

ColdHeat:
This is unfortunately not very easy. You would either need to create some kind of isolated Jinja environment or allow for the hacking in of the theme name via url parameter. Neither of which I really want to grapple with right now.

ColdHeat:
This is now possible thanks to #1817! There is still some trickiness to do but it is now technically feasible.

ColdHeat:
The Ghost project doesn't do theme previews in-app. They instead link to some kind of static site or example page where it can be viewed. I feel like this is probably the right approach.

PR #1817, opened 2021-03-08
refac: Allow loading theme templates by prefix

maybe-sybr:
This change reworks the template loaders so that we make more liberal
use of the ordered preferential loading provided by `ChoiceLoader` to
simplify the `ThemeLoader` so it only needs to be aware of a single
theme name to load templates from. This allows us to more easily
implement loading of templates from specific themes by providing an
extra `PrefixLoader` which is aware of all of the themes available to
the application.

### DIFF
--- a/CTFd/__init__.py
+++ b/CTFd/__init__.py
@@ -118,6 +118,10 @@ class ThemeLoader(FileSystemLoader):
 
     def get_source(self, environment, template):
         # Refuse to load `admin/*` from a loader not for the admin theme
+        # Because there is a single template loader, themes can essentially
+        # provide files for other themes. This could end up causing issues if
+        # an admin theme references a file that doesn't exist that a malicious
+        # theme provides.
         if template.startswith(self._ADMIN_THEME_PREFIX):
             if self.theme_name != ADMIN_THEME:
                 raise jinja2.TemplateNotFound(template)
--- a/CTFd/__init__.py
+++ b/CTFd/__init__.py
@@ -13,10 +13,10 @@ from jinja2.sandbox import SandboxedEnvironment
 from werkzeug.middleware.proxy_fix import ProxyFix
 from werkzeug.utils import cached_property
 
+import CTFd.utils.config
 from CTFd import utils
-from CTFd.constants.themes import DEFAULT_THEME
+from CTFd.constants.themes import ADMIN_THEME, DEFAULT_THEME
 from CTFd.plugins import init_plugins
-from CTFd.utils.config import ctf_theme_candidates
 from CTFd.utils.crypto import sha256
 from CTFd.utils.initialization import (
     init_events,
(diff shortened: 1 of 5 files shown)


maybe-sybr:
Per discussion on #1808. I haven't added any tests for this yet because I think it might be a bit fiddly to do so. Happy for any input on that.

ColdHeat:
Thanks for PR'ing this as well! Looks a little complicated so it'll take some time to get through. 

But just so I'm clear, this allows for things like `{% extends core/challenges.html %}` and the same with other themes as the initial prefix?

CTFd/__init__.py
```diff
@@ -100,35 +100,31 @@ def _load_template(self, name, globals):
...
 
 class ThemeLoader(FileSystemLoader):
-    """Custom FileSystemLoader that switches themes based on the configuration value"""
-
-    def __init__(self, searchpath, encoding="utf-8", followlinks=False):
+    """Custom FileSystemLoader that is aware of theme structure and config.
+    """
+
+    DEFAULT_THEMES_PATH = os.path.join(os.path.dirname(__file__), "themes")
+    _ADMIN_THEME_PREFIX = ADMIN_THEME + "/"
```
ColdHeat, line 107: Is there a reason that this variable is "private"?

CTFd/__init__.py
```diff
@@ -100,35 +100,31 @@ def _load_template(self, name, globals):
...
-                pass
-        else:
-            # If we found no matching templates then raise an exception for the
-            # first one we intended to try
-            raise jinja2.exceptions.TemplateNotFound(first_cand)
+        # Refuse to load `admin/*` from a loader not for the admin theme
+        if template.startswith(self._ADMIN_THEME_PREFIX):
+            if self.theme_name != ADMIN_THEME:
+                raise jinja2.TemplateNotFound(template)
+            template = template[len(self._ADMIN_THEME_PREFIX) :]
```
ColdHeat, line 124: This seems to differ from the previous code but I can't tell why it's needed. 

Wouldn't the admin themes only ever be handled by the admin ThemeLoader?

ColdHeat (commented):
Looks overall pretty good, just a few questions so far if you have the time.

CTFd/__init__.py
```diff
@@ -100,35 +100,31 @@ def _load_template(self, name, globals):
...
+        theme_name = self.theme_name or str(utils.get_config("ctf_theme"))
```
ColdHeat, line 125: Can this not be set in the constructor if `theme_name` is `None`?

CTFd/__init__.py
```diff
@@ -100,35 +100,31 @@ def _load_template(self, name, globals):
...
+    def __init__(
+        self,
+        searchpath=DEFAULT_THEMES_PATH,
+        theme_name=None,
+        encoding="utf-8",
+        followlinks=False,
+    ):
         super(ThemeLoader, self).__init__(searchpath, encoding, followlinks)
-        self.overriden_templates = {}
+        self.theme_name = theme_name
```
ColdHeat, line 117:
```suggestion
        self.theme_name = theme_name or str(utils.get_config("ctf_theme"))
```

Would this work?

ColdHeat:
wrt testing, I suspect that maybe the simplest thing to do would be to use [`render_template_string()`](https://flask.palletsprojects.com/en/1.1.x/api/?highlight=render_template_string#flask.render_template_string) to test that all of the extends functionality works?

e.g. maybe something like:

```
from flask import render_template_string

render_template_string("{% extends core/page.html %}", content="test")
```

ColdHeat:
I think this PR also allows us to solve #1194 which would be awesome! Very excited to merge this in!

maybe-sybr:
> Thanks for PR'ing this as well! Looks a little complicated so it'll take some time to get through.
> 
> But just so I'm clear, this allows for things like `{% extends core/challenges.html %}` and the same with other themes as the initial prefix?

Yep, sure does. In fact I'm using on a staging fork of CTFd for exactly this purpose. He's an example of the template we use in our theme (it looks trivial because we have a bunch of react stuff involved which actually makes it do useful things):

```jinja2
{% extends "core/challenges.html" %}

{% block challenge_content %}
<div id="challenges-board" class="justify-content-md-center"></div>
{% endblock %}

{% block entrypoint %}
<div id="csrfNonce" style="visibility: hidden">{{ Session.nonce }}</div>
{% include "scripts.html" %}
{% endblock %}
```

CTFd/__init__.py
maybe-sybr: Mostly to discourage external use since it's relevant only to the `ThemeLoader` and nobody else should really be talking about filesystem paths for themes. Happy to make it public though since it's barely relevant.

CTFd/__init__.py
maybe-sybr: It looks like you must have realised why this is done each time, but just to answer the question, it's to avoid ignoring a change of theme at runtime :) Being able to alter the `theme_name` of the default `ThemeLoader` instance (ie. the one with no `theme_name` set) when we alter the config would be useful but IIUC this is done by some indirect logic so working out when the `ctf_theme` option changes specifically might be annoying to do. If we did manage to do that though, the change you suggested on L117 would be exactly what we'd want to do.

maybe-sybr:
> wrt testing, I suspect that maybe the simplest thing to do would be to use [`render_template_string()`](https://flask.palletsprojects.com/en/1.1.x/api/?highlight=render_template_string#flask.render_template_string) to test that all of the extends functionality works?
> 
> e.g. maybe something like:
> 
> ```
> from flask import render_template_string
> 
> render_template_string("{% extends 'core/page.html' %}", content="test")
> ```

Great idea. I'll look into that over the weekend.

CTFd/__init__.py
maybe-sybr: This is intended as a bit of a mitigation against evil themes. The idea being that if an admin installs a theme which provides a `templates/admin/set_password.html` file, they might get tricked into entering data on a malicious form. It's not particularly strong and perhaps a better approach would be to move the prefix loader (which includes the admin theme) in front of the default loader when we construct the `ChoiceLoader`. That would mean the admin theme would always get first chance at resolving `admin/...` template paths and as long as it doesn't refer to absent template files, the default theme wouldn't be able to add it's own content to the admin panels.

In any case, this could use a comment to explain itself if we choose to leave it as is.

Thoughts?

maybe-sybr: Moving the preflix loader in front of the default loader would have the consequence that collisions between the name of another theme (say `foo`) and a subdirectory of the default theme's templates (say `bar/templates/foo`) could end up rendering a template from `foo` rather than the expected one from `bar`.

maybe-sybr:
Haven't had a chance to add tests for this yet. Perhaps tomorrow.

CTFd/__init__.py
ColdHeat: I see the concern now after testing it out. Kind of stems from the lack of separate template environments in Jinja I think but I think your solution is good.

ColdHeat:
I think this looks good to go with these additional tests. I think the new loader functionality is easy enough to verify manually and I think the test would be more complicated to write than it would be worth so I'll just leave it at that. 

Thanks for another great PR!

maybe-sybr:
Great, thanks @ColdHeat. Much appreciated again :)

PR #1840, opened 2021-03-21
fix: Ensure hidden users see their own solves

### LINKED ISSUE #1839: Hidden users can't see challenges they meet the prerequisites for
**Environment**:

- CTFd Version/Commit: 3.3 dev branch
- Operating System:
- Web Browser and Version:

**What happened?**
Solved a challenge which is a prerequisite for another as the admin user which is hidden by default. The other challenge does not appear in the challenge list from the API.

The challenge with a prerequisite set should be visible regardless of the hidden status of the user in question. The logic for banned users should probably not change.

1. Add a user, make the hidden
2. Add challenge A with some flag
3. Add challenge B and set A as a prerequisite
4. Log in as user and solve A
5. Check challenge list API and observe B is not visible
6. Change user to not hidden and then check challenge list again as user - B should be visible

maybe-sybr:
Fixes #1839

### DIFF
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -84,6 +84,11 @@ challenges_namespace.schema_model(
 
 
 def _build_solves_query(extra_filters=(), admin_view=False):
+    """Returns queries and data that that are used for showing an account's solves.
+    It returns a tuple of
+        - SQLAlchemy query with (challenge_id, solve_count_for_challenge_id)
+        - Current user's solved challenge IDs
+    """
     # This can return None (unauth) if visibility is set to public
     user = get_current_user()
     # We only set a condition for matching user solves if there is a user and
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -4,8 +4,7 @@ from typing import List
 from flask import abort, render_template, request, url_for
 from flask_restx import Namespace, Resource
 from sqlalchemy import func as sa_func
-from sqlalchemy import types as sa_types
-from sqlalchemy.sql import and_, cast, false, or_, true
+from sqlalchemy.sql import and_, false, true
 
 from CTFd.api.v1.helpers.request import validate_args
 from CTFd.api.v1.helpers.schemas import sqlalchemy_to_pydantic
@@ -121,7 +120,7 @@ def _build_solves_query(extra_filters=(), admin_view=False):
     solve_ids = (
         Solves.query.with_entities(Solves.challenge_id).filter(user_solved_cond).all()
     )
-    solve_ids = set([value for value, in solve_ids])
+    solve_ids = {value for value, in solve_ids}
     return solves_q, solve_ids
 
 
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -105,26 +105,20 @@ def _build_solves_query(extra_filters=(), admin_view=False):
     # or banned users/teams, even if we are an admin. This is to match the
     # challenge detail API.
     exclude_solves_cond = and_(
-        AccountModel.banned == false(),
-        AccountModel.hidden == false(),
+        AccountModel.banned == false(), AccountModel.hidden == false(),
     )
     # This query counts the number of solves per challenge, as well as the sum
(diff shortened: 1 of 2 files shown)


maybe-sybr:
This could use a unit test prior to merge. I don't have the time right now but can get to it later today

maybe-sybr:
Okay, this updated diff now has some tests for the various hidden/banned behaviours and also actually works when no user is logged in.

ColdHeat:
Thanks for noticing this!

I took a look here and I think now the `/api/v1/challenges` response for `solves` is including the hidden users as well?

ColdHeat:
I'm actually not sure how to resolve this without maybe having a special edge case for hidden users...

ColdHeat:
So assuming `_build_solves_query` returns `chal_id, solve_count, solved_by_user`.

I would say that:
- `solve_count` should never show a hidden user.
- `solved_by_user` should show things specific to the user regardless of the user's hidden status.

ColdHeat:
Okay I see in your `test_api_challenges_get_solve_count_hidden_user` test that you expect the admin to be able to see their own solve even if they're hidden. 

This behavior is different than what it was before. For example the `/[challenge_id]/solves` doesn't show hidden user solves.

I separated the build solves query function out into two queries:
- The challenge solves building query
- The solves for the current user query (copied from v3.2)

This changes some of the logic but I think this is the simplest way of decoupling challenge solves and user solves.

ColdHeat:
Okay I think that if you agree with these changes, this code is probably good to merge. Thoughts?

maybe-sybr:
Yep, I think you're right that trying to keep these in a single query with that extra requirement would make it too difficult to comprehend in future.

CTFd/api/v1/challenges.py
```diff
@@ -102,24 +102,26 @@ def _build_solves_query(extra_filters=(), admin_view=False):
...
         .filter(*extra_filters, freeze_cond, exclude_solves_cond)
         .group_by(Solves.challenge_id)
     )
-    return solves_q
+    # Also gather the user's solve items which can be different from above query
+    # Even if we are a hidden user, we should see that we have solved the challenge
+    # But as a hidden user we are not included in the count
+    solve_ids = (
+        Solves.query.with_entities(Solves.challenge_id).filter(user_solved_cond).all()
+    )
```
maybe-sybr, line 122: This probably needs to respect `freeze_cond, *extra_filters` as well

maybe-sybr: Perhaps we should create a list like `solve_filters = [*extra_filters, freeze_cond, exclude_solves_cond]` to avoid drift between the two queries

ColdHeat: Off of the top of my head I'm not sure... I believe this is for calculating the user's solves which should always be shown regardless of freeze time. For example, I can solve a challenge during freeze time and I should be able to see that, but other teams shouldn't be able to see that I have solved it. 

But not totally sure, I will confirm the behavior later today.

maybe-sybr: Right, I vaguely thought that might be the case but wasn't sure what the expected behaviour should be. I guess the visible effects of this would be:
 * solve counts will only include solves which can contribute to final scores
 * solves outside of the freeze time will properly mark challenges on the board green so the player (and JS) can know not to re-attempt it
 * since the `solves/` endpoint should respect freeze time, solves outside the freeze time will match the score count in the challenge list and details API endpoints

Sounds alright to me :+1:

ColdHeat: I think there is an issue where a user's profile page (`/user`, `/profile`) doesn't show solves during freeze time. Might be a source of confusion here. 

I think that's incorrect and we should always show the user's solves to them. Don't need to fix it in this PR but I'll probably fix it in the next one.

ColdHeat: I believe that I can confirm that we do not need to consider freeze time in the second query. 

It should always return the user's solves. 

I will check for other edge cases like anonymous users, add a clearer docstring and then merge.

ColdHeat: Actually I'm not even sure if we should show a user's solves in their profile during freeze time.... I'll think about it before releasing 3.3.

PR #1862, opened 2021-04-06
ping database with python instead of mysql client

### LINKED ISSUE #725: Use sqlalchemy Python script to test availability of database.

frankli0324:
* Replaces `mysqladmin ping` with a custom script
* Closes #725

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -7,10 +7,18 @@ ACCESS_LOG=${ACCESS_LOG:--}
 ERROR_LOG=${ERROR_LOG:--}
 WORKER_TEMP_DIR=${WORKER_TEMP_DIR:-/dev/shm}
 
+# Check that a .ctfd_secret_key file or SECRET_KEY envvar is set
+if [ ! -f .ctfd_secret_key ] && [ -z "$SECRET_KEY" ]; then
+    if [ $WORKERS -gt 1 ]; then
+        echo "[ ERROR ] You are configured to use more than 1 worker."
+        echo "[ ERROR ] To do this, you must define the SECRET_KEY environment variable or create a .ctfd_secret_key file."
+        echo "[ ERROR ] Exiting..."
+        exit 1
+    fi
+fi
+
 # Ensures that the database is available
 python ping.py
-# calling config.py ensures that a SECRET_KEY is set, complying with the following priority:
-# config.ini > environment variable > .ctfd_secret_key > auto generated .ctfd_secret_key
 
 # Initialize database
 python manage.py db upgrade
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -6,6 +6,7 @@ WORKER_CLASS=${WORKER_CLASS:-gevent}
 ACCESS_LOG=${ACCESS_LOG:--}
 ERROR_LOG=${ERROR_LOG:--}
 WORKER_TEMP_DIR=${WORKER_TEMP_DIR:-/dev/shm}
+SECRET_KEY=${SECRET_KEY:-}
 
 # Check that a .ctfd_secret_key file or SECRET_KEY envvar is set
 if [ ! -f .ctfd_secret_key ] && [ -z "$SECRET_KEY" ]; then
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -6,35 +6,11 @@ WORKER_CLASS=${WORKER_CLASS:-gevent}
 ACCESS_LOG=${ACCESS_LOG:--}
 ERROR_LOG=${ERROR_LOG:--}
 WORKER_TEMP_DIR=${WORKER_TEMP_DIR:-/dev/shm}
-SECRET_KEY=${SECRET_KEY:-}
-DATABASE_URL=${DATABASE_URL:-}
 
-# Check that a .ctfd_secret_key file or SECRET_KEY envvar is set
-if [ ! -f .ctfd_secret_key ] && [ -z "$SECRET_KEY" ]; then
-    if [ $WORKERS -gt 1 ]; then
-        echo "[ ERROR ] You are configured to use more than 1 worker."
-        echo "[ ERROR ] To do this, you must define the SECRET_KEY environment variable or create a .ctfd_secret_key file."
-        echo "[ ERROR ] Exiting..."
-        exit 1
-    fi
-fi
-
-# Check that the database is available
-if [ -n "$DATABASE_URL" ]
-    then
-    url=`echo $DATABASE_URL | awk -F[@//] '{print $4}'`
-    database=`echo $url | awk -F[:] '{print $1}'`
-    port=`echo $url | awk -F[:] '{print $2}'`
-    echo "Waiting for $database:$port to be ready"
-    while ! mysqladmin ping -h "$database" -P "$port" --silent; do
-        # Show some progress
-        echo -n '.';
-        sleep 1;
-    done
-    echo "$database is ready"
-    # Give it another second.
-    sleep 1;
-fi
+# Ensures that the database is available
+python ping.py
+# calling config.py ensures that a SECRET_KEY is set, complying with the following priority:
+# config.ini > environment variable > .ctfd_secret_key > auto generated .ctfd_secret_key
 
 # Initialize database
 python manage.py db upgrade
--- a/ping.py
+++ b/ping.py
@@ -1,5 +1,6 @@
 import time
 
+from sqlalchemy import create_engine
 from sqlalchemy.engine.url import make_url
 
 from CTFd.config import Config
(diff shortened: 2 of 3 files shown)


frankli0324:
¿¿¿

docker-entrypoint.sh
```diff
@@ -6,35 +6,11 @@ WORKER_CLASS=${WORKER_CLASS:-gevent}
...
 
-# Check that a .ctfd_secret_key file or SECRET_KEY envvar is set
-if [ ! -f .ctfd_secret_key ] && [ -z "$SECRET_KEY" ]; then
-    if [ $WORKERS -gt 1 ]; then
-        echo "[ ERROR ] You are configured to use more than 1 worker."
-        echo "[ ERROR ] To do this, you must define the SECRET_KEY environment variable or create a .ctfd_secret_key file."
-        echo "[ ERROR ] Exiting..."
-        exit 1
-    fi
-fi
```
ColdHeat, line 20: I think it's a bad idea to remove this. You're right that the script doesn't consider if config.ini is hard coded but I think with a Docker container the better solution would be to use environment variables anyway.

ping.py
```diff
@@ -0,0 +1,26 @@
...
+import time
+
+from sqlalchemy import create_engine
+from sqlalchemy.engine.url import make_url
+
+from CTFd.config import Config
+
+url = make_url(Config.DATABASE_URL)
+if not url.drivername.startswith("mysql"):
+    exit(0)
```
ColdHeat, line 11: Shouldn't this script work regardless of database driver? Can this just be removed?

ping.py
```diff
@@ -0,0 +1,26 @@
...
+
+engine = create_engine(url)
+print(f"Waiting for {url.host}:{url.port or 3306} to be ready")
+
+while True:
+    try:
+        engine.raw_connection().ping()
+        print()
```
ColdHeat, line 19: I don't think this is necessary? Does it just mimic the original mysql client's behavior?

ping.py
```diff
@@ -0,0 +1,26 @@
...
+        break
+    except:
```
ColdHeat, line 21: Is there a more specific error that can be caught here?

ColdHeat (changes requested):
I like the idea here! I think this is a good solution, just a few minor things that I think are worth addressing.

ping.py
frankli0324: `engine.raw_connection()` returns a underlying connection from another driver, `pymysql` in this case. `ping` is a mariadb/mysql only operation.

docker-entrypoint.sh
frankli0324: but `config.py` does respects env variables...

ping.py
frankli0324: hmmm... no it doesn't. I agree this is weird, logging in to the database *then* sending a ping packet. does removing the `.ping()` part sound reasonable? by doing so we can also remove the `driver.startswith(mysql)` check

frankli0324: how about

```python
while true try:
    engine.connect()
except pymysql.err.Error:
    ...
```

ColdHeat: I just meant removing the empty `print` statement here actually but I will look into the `ping` method as well.

docker-entrypoint.sh
ColdHeat: It respects the envvars but it won't fail the container if the secret key is being randomly generated. This check is primarily for having different secret keys across different workers.

ping.py
ColdHeat: Does this work though? I'm getting an error if MySQL isn't on. 

```
InternalError: (pymysql.err.InternalError) (1049, "Unknown database 'ctfd'")
(Background on this error at: http://sqlalche.me/e/2j85)
```

ColdHeat: We do have SQLAlchemy-Utils so maybe you can use https://sqlalchemy-utils.readthedocs.io/en/latest/database_helpers.html?highlight=database_exists#sqlalchemy_utils.functions.database_exists?

ColdHeat: It looks like you can do something like this: https://stackoverflow.com/a/17941078

Without a database then you can test for the raw connection.

ColdHeat: Looks like this strategy would also work for postgres. 

So you would just need to take the url and recreate it without the database.

ping.py
ColdHeat: I'm okay with this being a catch all as long as this script is covered under flake8 and you ignore the flake8 lint.

ping.py
ColdHeat: With the strategy below you'd probably just want to exit if the driver is sqlite.

ping.py
frankli0324: I wrote that on my phone, haven't tested, sorry😓
it came to my mind that mariadb image actually created the database for us
image
so.... `database_exists` is fine I think?

ColdHeat: You're right but `mysqladmin ping` only tests for connectivity to the database server. It doesn't check for existence of the database. It might be unnecessary but I would still rather let CTFd be responsible for creating the database if it doesn't exist. 

I think just creating a Python version of `mysqladmin ping` is the best way forward so we don't accidentally break something.

So just https://stackoverflow.com/a/17941078.

frankli0324: I found that it's actually not that complicated

docker-entrypoint.sh
frankli0324: I wrote the comments because my script actually imported `config.py`, so it's impossible for different workers to have different secret keys.

ping.py
frankli0324: the empty print statement is to print('ready') on a new line.

before:

```
Waiting for db: to be ready
..........db is ready
```

after:

```
Waiting for mysql+pymysql://ctfd:ctfd@db to be ready
.........
mysql+pymysql://ctfd:ctfd@db is ready
```

ColdHeat: I don't really think it's necessary to have. Can you remove it?

docker-entrypoint.sh
ColdHeat: I still think there are cases that could cause errors that are hard to debug. It's better to leave this as is. I doubt it causes any issues.

ping.py
```diff
@@ -0,0 +1,50 @@
...
+        if url.drivername.startswith("mysql"):
+            mysql_ping(
+                host=url.host,
+                user=url.username,
+                password=url.password,
+                port=url.port,
+                database=None,
+            )
+        elif url.drivername.startswith("postgres"):
+            postgres_ping()
```
ColdHeat, line 42: This is the wrong strategy to go with. You just need to do a `raw_connection()` on the URL without the database. Separating it out by postgres and mysql is unnecessary. You only really need to special case sqlite. For sqlite you can just `exit(0)`

frankli0324: `raw_connection` for postgres doesn't support `ping`ing

ColdHeat: I don't think you need to `ping()` specifically. `raw_connection()` in my testing worked only when the database was available for a connection. Otherwise it returned errors.

docker-entrypoint.sh
```diff
@@ -6,8 +6,6 @@ WORKER_CLASS=${WORKER_CLASS:-gevent}
 ACCESS_LOG=${ACCESS_LOG:--}
 ERROR_LOG=${ERROR_LOG:--}
 WORKER_TEMP_DIR=${WORKER_TEMP_DIR:-/dev/shm}
-SECRET_KEY=${SECRET_KEY:-}
```
ColdHeat, line 9: This var change will also need to be reverted

ping.py
frankli0324: yes, it also checks if a database is created. this isn't what `mysqladmin ping` does.

ColdHeat: mysqladmin ping just connects to the server. It doesn't test for database existence. 

https://serverfault.com/questions/32459/monitoring-what-does-mysqladmin-ping-report
https://dev.mysql.com/doc/refman/8.0/en/mysqladmin.html

ColdHeat: raw_connection doesn't need a database specified, you just need to remove it from the URL.

frankli0324: > mysqladmin ping just connects to the server. It doesn't test for database existence.
> 
> [serverfault.com/questions/32459/monitoring-what-does-mysqladmin-ping-report](https://serverfault.com/questions/32459/monitoring-what-does-mysqladmin-ping-report)
> [dev.mysql.com/doc/refman/8.0/en/mysqladmin.html](https://dev.mysql.com/doc/refman/8.0/en/mysqladmin.html)

I was trying to indicate the same thing... I wasn't aware that `raw_connection` accepts None as a database name...

frankli0324: do you have some instant messaging app? it's is so inefficient making a discussion here......

frankli0324:
running `make lint` locally doesn't produce any errors... wierd...

ColdHeat:
I can review this locally and take over from here. Thanks for working on this!

ColdHeat (approved):

Issue #1863, opened 2021-04-07
Github Actions Randomly fails on `test_theme_fallback_config`


frankli0324:
see run
https://github.com/CTFd/CTFd/runs/2254835896
https://github.com/CTFd/CTFd/runs/2281827174

the direct cause is that ga failed to `os.rmdir` an empty dir created with `try:os.mkdir();except OSError: pass`, my guess is that some error occurred creating the dir. trying to reproduce.

frankli0324:
oh, I'm on the wrong path.
both tests in `test_themes.py` uses the folder `join(app.root_path, "themes", "foo")`, so it's a race condition.

PR #1864, opened 2021-04-07
use different directories for different tests

frankli0324:
#1863


ColdHeat:
Seems reasonable, thanks for the PR.

Issue #928, opened 2019-03-30
Create test helpers for during ctftime and outside of ctftime 


ColdHeat:
This will help reduce some copy pasted boiler plate code.

ColdHeat:
I see this represented with a bunch of enums representing the times that are used and the appropriate freeze_time context managers.

PR #1866, opened 2021-04-16
Ctftime test context #928

ifeLawal:
A starting PR for issue #928. Want to confirm the structure of the context manager and enums before updating all the tests.

### DIFF
--- a/tests/constants/time.py
+++ b/tests/constants/time.py
@@ -1,7 +1,7 @@
 from CTFd.constants import RawEnum
 
 
-class FreezeTime(str, RawEnum):
+class FreezeTimes(str, RawEnum):
     NOT_STARTED = "2017-10-3"  # Tuesday, October 3, 2017
     STARTED = "2017-10-5"  # Thursday, October 5, 2017
     ENDED = "2017-10-7"  # Saturday, October 7, 2017
--- a/tests/constants/time.py
+++ b/tests/constants/time.py
@@ -1,0 +1,9 @@
+from CTFd.constants import RawEnum
+
+
+class FreezeTime(str, RawEnum):
+    NOT_STARTED = "2017-10-3"  # Tuesday, October 3, 2017
+    STARTED = "2017-10-5"  # Thursday, October 5, 2017
+    ENDED = "2017-10-7"  # Saturday, October 7, 2017
+    START = "1507089600"  # Wednesday, October 4, 2017 12:00:00 AM GMT-04:00 DST
+    END = "1507262400"  # Friday, October 6, 2017 12:00:00 AM GMT-04:00 DST
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -15,6 +15,7 @@ from werkzeug.datastructures import Headers
 from CTFd import create_app
 from CTFd.cache import cache, clear_standings
 from CTFd.config import TestingConfig
+from CTFd.constants import RawEnum
 from CTFd.models import (
     Awards,
     ChallengeComments,
@@ -40,6 +41,7 @@ from CTFd.models import (
     UserComments,
     Users,
 )
+from CTFd.utils import set_config
 
 text_type = str
 binary_type = bytes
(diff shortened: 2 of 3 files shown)


tests/helpers.py
```diff
@@ -61,6 +63,25 @@ def open(self, *args, **kwargs):
...
+
+
+class CTFtime:
+    def __enter__(self):
+        set_config(
+            "start", "1507089600"
+        )  # Wednesday, October 4, 2017 12:00:00 AM GMT-04:00 DST
+        set_config(
+            "end", "1507262400"
+        )  # Friday, October 6, 2017 12:00:00 AM GMT-04:00 DST
```
ColdHeat, line 79: We can move these constants into the above enum.

tests/helpers.py
```diff
@@ -61,6 +63,25 @@ def open(self, *args, **kwargs):
         return super(CTFdTestClient, self).open(*args, **kwargs)
 
 
+class FreezeTime(str, RawEnum):
+    NOT_STARTED = "2017-10-3"  # Tuesday, October 3, 2017
+    STARTED = "2017-10-5"  # Thursday, October 5, 2017
+    ENDED = "2017-10-7"  # Saturday, October 7, 2017
```
ColdHeat, line 69: This should be moved into a constants file/directory in the tests so we don't overcrowd the helpers file.

tests/helpers.py
```diff
@@ -61,6 +63,25 @@ def open(self, *args, **kwargs):
...
+    def __enter__(self):
+        set_config(
+            "start", "1507089600"
+        )  # Wednesday, October 4, 2017 12:00:00 AM GMT-04:00 DST
+        set_config(
+            "end", "1507262400"
+        )  # Friday, October 6, 2017 12:00:00 AM GMT-04:00 DST
+
+    def __exit__(self, exc_type, exec_value, exec_traceback):
+        pass
```
ColdHeat, line 82: In this code block you should probably undo the above set_config calls

tests/utils/test_ctftime.py
```diff
@@ -144,20 +141,14 @@ def test_ctf_ended():
...
 
-        with freeze_time("2017-10-5"):
-            assert ctf_ended() is False
+            with freeze_time(FreezeTime.STARTED):
+                assert ctf_ended() is False
 
-        with freeze_time("2017-10-7"):
-            assert ctf_ended() is True
+            with freeze_time(FreezeTime.ENDED):
+                assert ctf_ended() is True
```
ColdHeat, line 153: I like the idea here but I think we can also keep going and create more context managers or functions so that we can setup more time context managers. 

What about something like:

```
with CTFTime.set():
    ...
    with CTFTime.started():
    ...
    with CTFtime.not_started():
    ...
    with CTFtime.ended():
```

Very rough code above, I didn't test it or think about it very much. Might work out better with a different structure.

ColdHeat (changes requested):

tests/helpers.py
ifeLawal: I'm using `set_config("start", 0)` to achieve this right now, but I also saw a `clear_config()` function https://github.com/CTFd/CTFd/blob/master/CTFd/cache/__init__.py#L21. Is `clear_config()` the better approach to undo the value setting?

tests/constants/time.py
```diff
@@ -0,0 +1,9 @@
+from CTFd.constants import RawEnum
+
+
+class FreezeTime(str, RawEnum):
```
ColdHeat, line 4: I would make this plural i.e. `FreezeTimes`

tests/helpers.py
```diff
@@ -61,6 +65,45 @@ def open(self, *args, **kwargs):
         return super(CTFdTestClient, self).open(*args, **kwargs)
 
 
+class CTFtime:
```
ColdHeat, line 68: I don't know why it bothers me but I think this is maybe better as  `class ctftime` ?

tests/helpers.py
```diff
@@ -61,6 +65,45 @@ def open(self, *args, **kwargs):
...
+    @contextmanager
+    def set():
```
ColdHeat, line 70: Can you leave docstrings on each of these functions detailing what they do? Just a simple explanation is fine. Like "This context manager can be used to setup start and end dates for a test CTFd".

tests/helpers.py
```diff
@@ -61,6 +65,45 @@ def open(self, *args, **kwargs):
...
+        try:
+            set_config("start", FreezeTime.START)
+            set_config("end", FreezeTime.END)
+            yield
+        finally:
+            set_config("start", 0)
+            set_config("end", 0)
```
ColdHeat, line 77: I think a better default is None. I believe there are calls elsewhere in the tests so maybe use whatevers there.

ColdHeat (changes requested):

tests/helpers.py
ColdHeat: You would probably want to do set_config("start", None)

tests/helpers.py
ifeLawal: The function is currently called ctftime https://github.com/CTFd/CTFd/blob/master/CTFd/utils/dates/__init__.py#L7
so I tried to avoid using the same name as to not confuse the two (although just uppercasing may not be a big enough change either)

PR #1867, opened 2021-04-17
Limit total number of teams

frohoff:
Support for admin config that limits total number of teams (only applies to non-banned/non-hidden teams). Also includes unit tests and virtualenv dir entries for git/docker/prettier ignore.

This is useful when using automation to spin up per-team cloud infrastructure to limit costs, especially in the case that information on how to join the CTF ends up on the internet.

image

image


frohoff:
fixed lint issues, cleaned up, and rebased

CTFd/auth.py
```diff
@@ -5,6 +5,7 @@
 from flask import current_app as app
 from flask import redirect, render_template, request, session, url_for
 from itsdangerous.exc import BadSignature, BadTimeSignature, SignatureExpired
+from werkzeug.exceptions import abort
```
ColdHeat, line 8: Why not import from flask?

CTFd/auth.py
```diff
@@ -503,6 +504,18 @@ def oauth_redirect():
 
                 team = Teams.query.filter_by(oauth_id=team_id).first()
                 if team is None:
+                    num_teams_limit = int(get_config("num_teams", default=0))
+                    num_teams = Teams.query.filter_by(
+                        banned=False, hidden=False
+                    ).count()
+                    if 0 < num_teams_limit <= num_teams:
```
ColdHeat, line 511: Why `0 <`? Couldn't this be:

```suggestion
                    if num_teams_limit and num_teams >= num_teams_limit:
```

CTFd/auth.py
```diff
@@ -503,6 +504,18 @@ def oauth_redirect():
...
+                        abort(
+                            403,
+                            description="Reached the maximum number of teams ({}). Please join an existing team.".format(
```
ColdHeat, line 514: Can you simplify this with f-strings?

CTFd/teams.py
```diff
@@ -197,6 +197,16 @@ def new():
             description="Team creation is currently disabled. Please join an existing team.",
         )
 
+    num_teams_limit = int(get_config("num_teams", default=0))
+    num_teams = Teams.query.filter_by(banned=False, hidden=False).count()
+    if 0 < num_teams_limit <= num_teams:
```
ColdHeat, line 202: Same simplification here right? 

```suggestion
    if num_teams_limit and num_teams >= num_teams_limit:
```

CTFd/teams.py
```diff
@@ -197,6 +197,16 @@ def new():
...
+        abort(
+            403,
+            description="Reached the maximum number of teams ({}). Please join an existing team.".format(
```
ColdHeat, line 205: f-strings?

ColdHeat (changes requested):
Looks good, sorry for the delayed merging of this. Overall just some minor things and this can be merged.

CTFd/auth.py
frohoff: I think was just a mistake when choosing which function to import in my IDE. I'll switch it to `flask.abort`.

CTFd/auth.py
frohoff: I think this was copied and modified from another code section that was using `format()` so I kept it the same for consistency but am happy to change it to f-strings if that's the preferred style for new changes.

CTFd/auth.py
frohoff: IIRC I started with something similar to the suggested code but then one of the CI linters wanted me to change it to what's in the commit. I'll try with this suggested code and see if it will accept it.

frohoff:
Made requested changes.

PR #1876, opened 2021-04-29
Confirmed REST API does delete directories in s3 bucket. The local filesystem didn't though and that was updated. Closes #1758

### LINKED ISSUE #1758: The REST API for deleting files does not remove the file's directory and does not update the Media Library list
**Environment**: 

- CTFd Version/Commit: 3.2.1
- Operating System: Docker (`python:3.6-slim-buster`)
- Web Browser and Version: NA

**What happened?**

I am using the REST API for deleting files (e.g. `"DELETE /api/v1/files/41 HTTP/1.1"`) and it seems to work fine. The file is removed. However, two things do go wrong I think (at least to my expectation).

1. The file's directory (which has a hash based name) is not deleted. This means, after a while there will be a lot of empty directories.
1. The list of files used by the Media Library is not updated (i.e. file is removed from the list) which means the list grows constantly. The result is a list with many non-existing files as they are deleted.

The REST API returns a successful `200` code which seems to match with the expected behaviour.

When a file is deleted using the REST API, I expect the directory (the hash-based name) to be deleted and that the list used by the Media Library is updated accordingly.

1. Upload a file (e.g. via the web interface or REST API). 
1. Use the REST API to delete this file.
1. Check the `upload/` folder and the Media Library for the behaviour described above.

None

ifeLawal:
- Refactored the `FileSystemUploader.delete` to delete the directory and everything inside using `shutil.rmtree` This solution is under the assumption that the database location is always a hash_folder/filename though.
- Closes https://github.com/CTFd/CTFd/issues/1758


CTFd/utils/uploads/uploaders.py
```diff
@@ -64,7 +64,8 @@ def download(self, filename):
 
     def delete(self, filename):
         if os.path.exists(os.path.join(self.base_path, filename)):
-            os.unlink(os.path.join(self.base_path, filename))
+            file_path = filename.split("/")[0]
```
ColdHeat, line 67: Shouldn't this be `os.path.split()`?

ColdHeat: Mildly concerned about what happens under Windows since the path seperator isn't `/`

PR #1882, opened 2021-05-09
added support for S3 signature version 4 authentication

nopresearcher:
Ran into an issue while accessing files that were uploaded to S3.  'SignatureDoesNotMatch - The request signature we calculated does not match the signature you provided. Check your key and signing method.'.  From troubleshooting it appears that v4 is being enforced on AWS in at least us-east-1 and us-east-2.  I updated the boto connect code to include the signature version.


ColdHeat:
Thanks for putting this up! I haven't personally seen this but it's certainly possible that it's an issue now. Do you know of any docs that show that v4 auth is the default now or anything like that?

nopresearcher:
I went down a rabbit hole of comments and notes from others. The issue was hit or miss for me amongst deployments, it may have been related to their role out across the data centers. Below is a link that states v4 is the current signing protocol.

https://docs.aws.amazon.com/general/latest/gr/sigv4_changes.html

zebde:
Further documentation showing that v4 is the default in "most" regions.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingAWSSDK.html#:~:text=For%20all%20AWS%20Regions%2C%20AWS,by%20default%20to%20authenticate%20requests.

ColdHeat:
What I don't understand is shouldn't boto3 be providing this default? Do we need to just update the boto3 version instead?

Matir:
It looks like the official policy from botocore is currently to continue to use v2 signatures: https://github.com/boto/boto3/issues/2680

From my testing, it seems like the correct version string is `s3v4`, which also works for GCP signatures, as discussed in #1953.

ColdHeat:
Okay I think this is convincing enough. I'll merge this and I will monitor for any breaks on my side and I also imagine if something breaks someone will say something.

Issue #1686, opened 2020-10-14
ValueError: '' does not appear to be an IPv4 or IPv6 address


slightlyskepticalpotat:
**Environment**:

- CTFd Version/Commit: 3.1.1, commit 4c31dc2
- Operating System: Ubuntu 20.04 LTS
- Web Browser and Version: Chromium 85.0.4183.121

**What happened?**

A "500 - Internal Server Error" page showed up after I clicked a user in /admin/users.

**What did you expect to happen?**

I expected to see a user page similar to [this](https://demo.ctfd.io/admin/users/1) (you need to be logged in as admin on the demo ctfd site to see it).

**How to reproduce your issue**

Install CTFd by cloning this, installing dependencies, and using prepare.sh. Then, go to the admin page to manage users and click on one of them. 

**Any associated stack traces or error logs**

Systemctl logs: 
```Oct 14 00:41:45 system-name gunicorn[37980]:     response = IP_ADDR_LOOKUP.get(addr)
Oct 14 00:41:45 system-name gunicorn[37980]:   File "/home/ctfd/.local/share/virtualenvs/CTFd-rOJbThUf/lib/python3.8/site-packages/maxminddb/reader.py", line>
Oct 14 00:41:45 system-name gunicorn[37980]:     (record, _) = self.get_with_prefix_len(ip_address)
Oct 14 00:41:45 system-name gunicorn[37980]:   File "/home/ctfd/.local/share/virtualenvs/CTFd-rOJbThUf/lib/python3.8/site-packages/maxminddb/reader.py", line>
Oct 14 00:41:45 system-name gunicorn[37980]:     address = compat_ip_address(ip_address)
Oct 14 00:41:45 system-name gunicorn[37980]:   File "/home/ctfd/.local/share/virtualenvs/CTFd-rOJbThUf/lib/python3.8/site-packages/maxminddb/compat.py", line>
Oct 14 00:41:45 system-name gunicorn[37980]:     return ipaddress.ip_address(address)
Oct 14 00:41:45 system-name gunicorn[37980]:   File "/usr/lib/python3.8/ipaddress.py", line 53, in ip_address
Oct 14 00:41:45 system-name gunicorn[37980]:     raise ValueError('%r does not appear to be an IPv4 or IPv6 address' %
Oct 14 00:41:45 system-name gunicorn[37980]: ValueError: '' does not appear to be an IPv4 or IPv6 address```

ColdHeat:
[...]

PR #1896, opened 2021-05-27
GeoIp bugfix

linted:
Added error handling in case `IP_ADDR_LOOKUP.get(addr)` fails. I'm not entirely sure what caused this issue in my installation, but it manifested as an empty string being passed in as addr. the exception that was raised caused admins to not be able to manage any users.


ColdHeat:
Thanks for putting this up. I believe this closes #1686. Although I'm pretty sure this is configuration error.

Issue #1881, opened 2021-05-07
Users in admin scoreboard show user position instead of team position


ColdHeat:
In teams mode on the admin panel, users are shown with their user position on the scoreboard instead of their teams position. We should be showing both.

PR #1903, opened 2021-06-01
1881 show users team position

ifeLawal:
- PR fixes https://github.com/CTFd/CTFd/issues/1881
- Edited the private, public, and admin user.html to show users points and place as well as the teams


ColdHeat:
I think this idea is kind of flawed because it really clutters up the users page, sorry to take up your time. I think the admin scoreboard revealing the user positions is enough for this. I'll handle fixing the user/team score discrepancy.

PR #1914, opened 2021-06-13
Sort challenges by id

ConnorNelson:
When you go over 256 (?) challenges, the sorting of everything gets chaotic. Before that threshold, everything seems to be implicitly secondarily sorted by id (probably just an artifact of the database internals). A nice and simple fix is to just sort everything by id explicitly as a secondary sorting key. For most users, this should not have any effect. For the (probably rare) users that go over this 256 barrier, this will make things function as expected.

### DIFF
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -227,7 +227,9 @@ class ChallengeList(Resource):
                 and_(Challenges.state != "hidden", Challenges.state != "locked")
             )
         chal_q = (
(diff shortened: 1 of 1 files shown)


ColdHeat:
Thanks for the PR! I think this change makes sense. Do you mind updating your branch so this can be merged?

ColdHeat:
Thanks! For the record, this doesn't happen on SQLite but does happen on MySQL. Who knows for what reason though.

ConnorNelson:
Interesting, I would guess that it's specific to the database's internals.

Issue #1889, opened 2021-05-12
Easier to change user mode


ColdHeat:
We need to make it easier to change between user and teams mode. If anything we should simply detect that the user wants to switch and then tell them of the downsides of it. It should be easy to do from Users -> Teams but the opposite would need to be a lossy conversion.

PR #1916, opened 2021-06-16
Use patch config to switch modes, but prompt user before making the switch

ifeLawal:
- Resolves issue #1889 
- Enable user_mode dropdown menu
- Implement JS pop-up to warn users of the ramifications with the switch
- Delete Submissions and Notifications when transitioning from teams to users


CTFd/themes/admin/assets/js/pages/configs.js
```diff
@@ -124,6 +124,44 @@ function updateConfigs(event) {
...
+    body: body,
+    success: function() {
+      const params = {
+        value: new_mode
+      };
+      CTFd.api
+        .patch_config({ configKey: "user_mode" }, params)
+        .then(_response => {
+          let formData = new FormData();
+          formData.append("notifications", true);
```
ColdHeat, line 148: Are you supposed to delete notifications here?

CTFd/themes/admin/assets/js/pages/configs.js
```diff
@@ -124,6 +124,44 @@ function updateConfigs(event) {
...
 
+function switchMode() {
+  let new_mode = $("#user_mode option:selected").val();
+  let body = "";
+  if (new_mode == "teams") {
+    body =
+      "Going from user mode to teams mode will not change anything, but are you sure you want to make the change?";
+  } else {
+    body =
+      "Going from teams mode to user mode <b style='color:red'>WILL REMOVE ALL SUBMISSION DATA</b>. Are you sure you want to make the change?";
```
ColdHeat, line 135: There's no other usage of bold red text. Better to just use `<strong>`. 

I would say add two line breaks between the sentences and use "this change" instead of "the change".

CTFd/themes/admin/assets/js/pages/configs.js
```diff
@@ -124,6 +124,44 @@ function updateConfigs(event) {
   });
 }
 
+function switchMode() {
+  let new_mode = $("#user_mode option:selected").val();
+  let body = "";
+  if (new_mode == "teams") {
+    body =
+      "Going from user mode to teams mode will not change anything, but are you sure you want to make the change?";
```
ColdHeat, line 132: This doesn't really convey why it's worth being cautious. I would reword so that the user understands that it's not necessarily the same when switching back.

CTFd/themes/admin/assets/js/pages/configs.js
```diff
@@ -374,6 +412,7 @@ $(() => {
   insertTimezones($("#freeze-timezone"));
 
   $(".config-section > form:not(.form-upload)").submit(updateConfigs);
+  $("#user-mode").change(switchMode);
```
ColdHeat, line 415: This is the wrong selector.

CTFd/themes/admin/assets/js/pages/configs.js
```diff
@@ -124,6 +124,44 @@ function updateConfigs(event) {
...
+      const params = {
+        value: new_mode
+      };
+      CTFd.api
+        .patch_config({ configKey: "user_mode" }, params)
+        .then(_response => {
+          let formData = new FormData();
+          formData.append("notifications", true);
+          formData.append("submissions", true);
+          formData.append("nonce", CTFd.config.csrfNonce);
```
ColdHeat, line 150: Shouldn't the formData only be needed in the if statement?

CTFd/themes/admin/assets/js/pages/configs.js
ColdHeat: You also have the additional problem here where if the user says no, the change is still loaded in the UI. So if a user starts a change, selects no, and then hits save, the change will still be made. 

I feel like you may want to consider an alternative approach. 

Some options off the top of my head:

- Don't automatically try to reset submissions and just tell the user that they should reset it themselves afterwards. I kind of think this may be the best because it doesn't cause data loss. As long as the user is aware of what's going on, I think it's overall okay.  
- Add another checkbox in the reset flow specifically for user_mode reset
- Handle resetting in the configs API route.

ColdHeat (changes requested):

CTFd/themes/admin/assets/js/pages/configs.js
ifeLawal: what about: "When switching from user mode to teams mode, user submissions will be kept but users can not solve problems unless on a team.<br> <br>They will also have to solve problems again under the team for the points to go towards the teams' score.<br><br>Are you sure you want to make this change?"

CTFd/themes/admin/assets/js/pages/configs.js
ifeLawal: That is true! Would a no function call in the ezQuery function be out of the question? IE the code below in ezq.js
```javascript
$(no).click(function() {
    args.reject();
  });
```
This would allow for setting the UI value back to the previous value if a user says no. This is because there may have to be a specific endpoint for changing from teams_mode to user_mode because there has to be some data loss. Teams data should be deleted and Users should have no team_id at the minimum and currently there is only the accounts deletion in the reset endpoint which deletes users and teams.

ColdHeat:
Closing this in favor of #1956

Issue #1892, opened 2021-05-20
Show length error when Configs provided are too long


ColdHeat:
We should show an error when Configs provided are too long. This is overall part of the whole process of making the API more strict though.

ColdHeat:
Also probably worth defining exactly how long TEXT is in the database. Maybe we could bump it up.

ColdHeat:
Closed by #1920

ColdHeat:
I manually verified that `python -c "print('A' * 65536)"` will generate a string that will fail to insert in MySQL but 65535 does not. This aligns with other references as found online https://stackoverflow.com/a/17851638. 64000 is a tad bit too small because of the conversion.

PR #1920, opened 2021-06-17
Show length error when Configs provided are too long

ifeLawal:
- Fix for issue #1892
- Utilizing the built-in schema validation to confirm config string length and that there exists a key (the empty key may not be needed).
- Flashing the error to a user via JS

The validation should take effect on the POST request as well since it is using the schema to do the validation.


CTFd/themes/admin/assets/js/pages/configs.js
```diff
@@ -119,8 +119,19 @@ function updateConfigs(event) {
...
 
-  CTFd.api.patch_config_list({}, params).then(_response => {
-    window.location.reload();
+  CTFd.api.patch_config_list({}, params).then(function(_response) {
+    if (_response.success) {
+      window.location.reload();
+    } else {
+      let errors =
+        `Error with ${_response.key}: ` +
+        _response.errors.value.reduce((err, accum) => accum + "\n" + err, "");
```
ColdHeat, line 128: Why not just `.join("\n")`?

CTFd/api/v1/config.py
```diff
@@ -109,8 +109,12 @@ def post(self):
     )
     def patch(self):
         req = request.get_json()
+        schema = ConfigSchema()
 
         for key, value in req.items():
+            response = schema.load({"key": key, "value": value})
+            if response.errors:
+                return {"success": False, "errors": response.errors, "key": key}, 400
```
ColdHeat, line 117: I think adding the key is the wrong idea here because it breaks the expectation of success, data, and errors that all the API endpoints are using. We can just add the key into the error.

PR #1927, opened 2021-06-29
Add response headers to `/api/v1/users/me`

ajvpot:
Adds X-CTFd-ID and X-CTFd-Type response headers to the `/api/v1/users/me` endpoint.
This allows you to use CTFd to manage sessions along with nginx's [auth_request](https://nginx.org/en/docs/http/ngx_http_auth_request_module.html) module to make reverse proxied applications aware of the user's identity.
```
location /secretapp/ {
    auth_request /auth;                
    auth_request_set $id $upstream_http_x_ctfd_id;      
    auth_request_set $type $upstream_http_x_ctfd_type;      
    proxy_set_header X-User-ID $id;
    proxy_set_header X-User-Type $type;
    proxy_pass protected_app_backend;
}

location = /auth {
    proxy_pass http://ctfd/api/v1/users/me;
    proxy_pass_request_body off;
    proxy_set_header Content-Length "";
    proxy_set_header X-Original-URI $request_uri;
}
```

### DIFF
--- a/CTFd/api/v1/users.py
+++ b/CTFd/api/v1/users.py
@@ -280,19 +280,25 @@ class UserPrivate(Resource):
     @users_namespace.doc(
         description="Endpoint to get the User object for the current user",
         responses={
-            200: ("Success", "UserDetailedSuccessResponse"),
             400: (
                 "An error occured processing the provided or stored data",
                 "APISimpleErrorResponse",
             ),
         },
     )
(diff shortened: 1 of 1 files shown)


ajvpot:
`flake8` wants it one way while `black` wants it another, please let me know what to do here.

ajvpot:
bump @ColdHeat

ColdHeat:
I think there's value in this but it seems like it should be in a plugin more so than in CTFd master? Whats the rationale for having it be in the upstream?

Issue #1928, opened 2021-06-30
API Error: config is too long - JSON bool values not sent correctly - 400 bad request on PATCH requests


phretor:
**Environment**:

- CTFd Version/Commit: 3.3.0 (`ctfd/ctfd:latest` image)
- Operating System: Docker on AWS ECS
- Web Browser and Version: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:89.0) Gecko/20100101 Firefox/89.0

**What happened?**
Send the following request:

```
curl 'https://host/api/v1/configs' -X PATCH -H 'User-Agent: Mozilla/5.0
(Macintosh; Intel Mac OS X 10.14; rv:89.0) Gecko/20100101 Firefox/89.0' -H
'Accept: application/json' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H
'Referer: https://host/admin/config' -H 'Content-Type: application/json' -H
'CSRF-Token: ****************************************************' -H 'Origin:
https://host' -H 'DNT: 1' -H 'Connection: keep-alive' -H 'Cookie:
session=******-******-**************' -H 'Sec-GPC: 1' -H 'TE: Trailers'
--data-raw $'{"successful_registration_email_subject":"Successfully registered
for {ctf_name}","successful_registration_email_body":"You\'ve successfully
    registered for {ctf_name}\041","verification_email_subject":"Confirm your
    account for {ctf_name}","verification_email_body":"Welcome to
    {ctf_name}\041\\r\\n\\r\\nClick the following link to confirm and activate
    your account:\\r\\n{url}\\r\\n\\r\\nIf the link is not clickable, try
    copying and pasting it into your
    browser.","user_creation_email_subject":"Message from
    {ctf_name}","user_creation_email_body":"A new account has been created for
    you for {ctf_name} at {url}. \\r\\n\\r\\nUsername: {name}\\r\\nPassword:
    {password}","password_reset_subject":"Password Reset Request from
    {ctf_name}","password_reset_body":"Did you initiate a password reset on
    {ctf_name}? If you didn\'t initiate this request you can ignore this email.
    \\r\\n\\r\\nClick the following link to reset your
    password:\\r\\n{url}\\r\\n\\r\\nIf the link is not clickable, try copying
    and pasting it into your
    browser.","password_change_alert_subject":"Password Change Confirmation for
    {ctf_name}","password_change_alert_body":"Your password for {ctf_name} has
[...]

PR #1929, opened 2021-07-01
Converting bool to str in config values

phretor:
Addressing regression caused by #1920 as discussed in #1928


ColdHeat:
Thanks for identifying a fix for this! I think this is a little too simplistic though so I'm going to close it in favor of #1931

PR #1949, opened 2021-07-16
Add time to export filename

nella17:


ColdHeat:
Is there a real benefit or need for this? I also dont see a reference for %T on https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior

nella17:
`%T` same as `%H:%M:%S`.
This change can make it more convenient to export data within a day.

image

export.py
```diff
@@ -19,7 +19,7 @@
             shutil.copyfileobj(backup, target)
     else:
         ctf_name = config.ctf_name()
-        day = datetime.datetime.now().strftime("%Y-%m-%d")
+        day = datetime.datetime.now().strftime("%Y-%m-%d %T")
```
ColdHeat, line 22: Can you add another dash here instead of space and I'll merge this.

PR #1986, opened 2021-09-10
Fix non-clickable checkbox label in user creation form in Admin side

### LINKED ISSUE #1985: Fix non-clicable checkbox label in user creation form in Admin side
The user creation form checkbox label for email user credentials isn't clickable. Small fix.

TheSpeedX:
Fixed #1985 by changing the id


ColdHeat:
This looks good thanks!

Issue #1990, opened 2021-09-19
Session cookies aren't permanent


nickfrost-asana:
**Environment**:

- CTFd Version/Commit: 3.4.0
- Operating System: Ubuntu 20.04
- Web Browser and Version: Chrome 93.0.4577.82

**What happened?**

After logging in, your login session expires as soon as you quit your browser.

**What did you expect to happen?**

Have the session cookie set to an expiration date, so it persists after restarting your browser.

**How to reproduce your issue**

1. Log into CTFd
2. Check your cookies in your dev console
3. Observe that the "session" cookie's lifetime is set to "Session", rather than an expiration date

ColdHeat:
Hmm I'm surprised I never realized this, I guess I so rarely entirely close my browser :)

PR #1991, opened 2021-09-19
Make login sessions permanent

nickfrost-asana:
Fixes https://github.com/CTFd/CTFd/issues/1990.


ColdHeat:
This seems valid, thanks for reporting and fixing!

Issue #1993, opened 2021-09-21
Broken URL generation when behind proxy


lamermoon:
**Environment**:

- CTFd Version/Commit: d03692d002938df0eef43404d38035e53c134fe2
- Operating System: Cent OS 7, Linux Kernel 3.10.0-1160.15.2.el7.x86_64, Docker version 20.10.4, build d3cb89e

**What happened?**
A fresh CTFd fails to load the scoreboard and challenges. Also, changes to the competition name and other configuration options like start and end times do not show any effect and do not get stored. This is caused by a broken URL generation algorithm that prepends the X-FORWARDED-PREFIX twice.

**What did you expect to happen?**
The scoreboard should show no results and the challenges page should display a dummy challenge I configured. The competition name should be changeable and the start and end time of the competition can be set to a different value.

**How to reproduce your issue**
In `docker-compose.yml` I commented out the Nginx section as I use Traefik as reverse proxy. CTFd is supposed to be reachable in the subdirectory `domain.com/matz`. Traefic runs a middleware that removes the `/matz` part from URLs before forwarding requests to CTFd. CTFd is aware of the proxy (REVERSE_PROXY=true in docker-compose.yml)

**Any associated stack traces or error logs**
![2021-09-21-132630_4480x1440_scrot](https://user-images.githubusercontent.com/11388608/134166074-e418fdc2-84cb-45c9-a73c-17f07163c3b7.png)
![2021-09-21-135022_4480x1440_scrot](https://user-images.githubusercontent.com/11388608/134166752-65c727ba-371e-4321-8e4a-15a578f09c0d.png)

ColdHeat:
Have you tried the `APPLICATION_ROOT` setting in config.ini?

lamermoon:
That solved the problem, thanks!
However, in my browser I can see that /matz still gets prepended twice (screenshots would still look the same with a single difference: 404 codes are now 200). Due to the `APPLICATION_ROOT = /matz` setting, CTFd now just knows what to do with the request. Wouldn't it make sense to prevent the double-prepending in the first place?

ColdHeat:
Okay after very carefully reading this and looking at your screenshot more carefully, I think you are looking for the more specific setting of REVERSE_PROXY: https://github.com/CTFd/CTFd/blob/master/CTFd/config.ini#L139-L143. You can control how Flask's ProxyFix functionality handles the X-Forwarded headers. Perhaps `1,1,1,1,0` instead of just `true`. 

[...]

PR #1999, opened 2021-10-02
fix regression in REVERSE_PROXY to allow comma seperated integers

lecafard:
The reverse proxy variable used to accept numbers `REVERSE_PROXY: "2,0,0,0,0"` to configure the proxy headers, but now only supports `True` or `False`. This PR removes the `process_boolean_str` function from wrapping it.

Also appears to be part of #1993 .


ColdHeat:
Sorry for the delay with this. I was concerned that the type loosening would break something but it looks like it was all handled before. Thanks for the PR! Will merge as soon as tests pass on the merge.

PR #2000, opened 2021-10-03
Fix typo for dynamic challenge description

iBotPeaches:
I keep re-reading this and I can't tell if the sentence is talking about the folks who have answered the question and have a sort of scale of decreasing values or talking about the unanswered folks that have a point value chance on completion less than/equal to all of those who have answered it.

### DIFF
--- a/CTFd/plugins/dynamic_challenges/assets/create.html
+++ b/CTFd/plugins/dynamic_challenges/assets/create.html
@@ -3,7 +3,7 @@
 {% block header %}
 <div class="alert alert-secondary" role="alert">
 	Dynamic value challenges decrease in value as they receive solves. The more solves a dynamic challenge has,
-	the lower its value is to everyone who has solved it.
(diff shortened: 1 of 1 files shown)


ColdHeat:
Hmm I don't know if the current text is difficult to read but your suggested change doesn't make sense for the functionality. 

Here is some other information quoted from the docs:

> A dynamic value challenge is a challenge whose point value decreases after each solve. By reducing the value of the challenge on each solve, all users who have previously solved the challenge will have lowered scores. Thus an easier and more solved challenge will naturally have a lower point value than a harder and less solved challenge.

Basically the more solves a challenge receives, the lower the value of the challenge to everyone. That ***includes*** the players who have already solved the challenge.

iBotPeaches:
Ah I see my mistake. I thought scores were immutable once scored on a challenge. My change makes no sense then. Removing.

Thanks.

Issue #2003, opened 2021-10-07
All numeric registration code does not work


ColdHeat:
Registration codes that are all numeric don't work b/c we automatically cast it to an int. 

This code needs a call to str so that we get a string always. 
https://github.com/CTFd/CTFd/blob/514ab2c8bd3b0687615307c19fa9618b09e3998a/CTFd/auth.py#L214-L219

ColdHeat:
Sure, you don’t need to ask permission.

PR #2004, opened 2021-10-09
#2003 - All numeric registration codes

iBotPeaches:
fixes: #2003


ColdHeat:
Looks reasonable, thanks for including the test!

PR #2016, opened 2021-11-03
Change logging to csv

kraft87:
Changed logging to csv format and added more details in order to be easier to parse in ELK and logstash.


kraft87:
@ColdHeat if there is another format you would like the logs to be in let me know, I am mostly trying to add additional information to the logs

ColdHeat:
@kraft87 sorry I missed this pull request! 

I think formatting this as CSV makes all the strings look kind of disorganized for not a lot of gain. Is CSV an accepted/common format in the logging world? What about logging to JSON?

kraft87:
@ColdHeat We changed the logging from CSV to JSON like you suggested. Are there any issues with the changes?

PR #2033, opened 2021-12-03
Remove double printing of log

PinkDraconian:
Spent way to long trying to figure out why some of my logs came out twice. Here's the culprit!


ColdHeat:
Thanks for filing this! Where were you seeing double logging? 

I think this print call is when using the debugging server like `serve.py`.  I'm happy to remove it but just need to check whether it affects anything.

PinkDraconian:
@ColdHeat I was using a plugin which used the `log` function.

I'm deploying using docker-compose and this caused every statement being logged using that function to appear twice in the logs.

PR #2073, opened 2022-03-22
Show admins the email server error message when email sending fails

ColdHeat:
* Show admins the email server error message when email sending fails


ColdHeat:
* Show admins the email server error message when email sending fails

PR #2085, opened 2022-04-09
string error Update

Estheryu991:
Since strings are not functions. In line 288, I added string_error because we are calling function () to the end of a function name. This error commonly occurs when you assign a variable called “str”, to use the str() function. Python interprets “str” as a string and you cannot use the str() function in your program


ColdHeat:
Not sure this is a valid PR? If it's for some purpose follow up and I can re-open.

Issue #2086, opened 2022-04-11
Outdated Nginx version in docker-compose.yml


IsaacMe:
**Environment**:

- CTFd Version/Commit: [afc55bf](https://github.com/CTFd/CTFd/commit/afc55bff75481d149eb8416ebf6ec40044ee1416) (latest)
- Operating System: n/a
- Web Browser and Version: n/a

**What happened?**
In the current [docker-compose.yml](https://github.com/CTFd/CTFd/blob/master/docker-compose.yml) file the version of Nginx is pinned on version 1.17. However, this version is not supported anymore and does not recieve any security patches anymore.

**What did you expect to happen?**
I think the version should be updated to at least 1.20 or the docker-compose file should use `stable` or `latest`. I could make a pull request for this, but I was curious (and could not directly find it) if there was a reason to pin the Nginx version to 1.17?

**How to reproduce your issue**
n/a

**Any associated stack traces or error logs**
n/a

PR #2088, opened 2022-04-12
Docker-compose now always uses stable nginx

IsaacMe:
As discussed in #2086, this PR changes the docker-compose file to automatically uses the latest version of Nginx instead of a pinned version.


ColdHeat:
Thanks for filing this!

PR #2100, opened 2022-04-23
Add a notice for China user to use local mirrors in Dockefile

wbsdty331:
use local mirrors in Dockerfile to improve apt download speed in china.


ColdHeat:
This seems more appropriate for a fork or seperate Dockerfile. And even then, I really can't imagine that the speeds would be that significantly hampered? Also the end user can pull one of the built Docker images as well.

PR #2108, opened 2022-05-02
Fix the order of the solves of the user pages to the chronological ordering (latest first).

Cryptanalyse:
**Problem:**
When one looks at a user page, the list of the solves is in a seemingly random order.

**Solution suggested:**
This small PR fixes that ordering by sorting the solves in reverse chronological order (latest solve first).


ColdHeat:
I feel like it was written to be ordered in date descending but it is possible that that functionality was lost somewhere.

Cryptanalyse:
Possible yes. I have a fork of CTFd, and without the ordering fixed (asc or desc, as you prefer), the order appears somewhat random. I feel like even if it's already sorted in this original repo, there is no harm in re-sorting just in case some plugins modify it.

PR #2113, opened 2022-05-06
[Admin] Add a link on the IP modal to search for a given IP address accross all users.

Cryptanalyse:
Add the possibility to click on a user IP on the admin to search for this IP, as if if were entered in the search field.

### DIFF
--- a/CTFd/themes/admin/templates/modals/users/addresses.html
+++ b/CTFd/themes/admin/templates/modals/users/addresses.html
@@ -12,7 +12,11 @@
 			<tbody>
 			{% for addr in addrs %}
 				<tr>
-					<td class="text-center">{{ addr.ip }}</td>
+					<td class="text-center">
+						<a href="{{ url_for('admin.view') }}/users?field=ip&q={{ addr.ip }}">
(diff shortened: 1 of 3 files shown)


CTFd/themes/admin/templates/modals/users/addresses.html
```diff
@@ -12,7 +12,11 @@
 			<tbody>
 			{% for addr in addrs %}
 				<tr>
-					<td class="text-center">{{ addr.ip }}</td>
+					<td class="text-center">
+						<a href="{{ url_for('admin.view') }}/users?field=ip&q={{ addr.ip }}">
```
ColdHeat, line 16: You should be able to do this entirely with a url_for call. Something like `url_for('admin.users', field='ip', q=addr.ip)`.

ColdHeat (changes requested):
If you can make this change I will merge

Cryptanalyse:
Done! Indeed, this is cleaner, thanks!

ColdHeat:
Made a few other changes but thanks for adding this I think it is a good feature.

Issue #2133, opened 2022-05-30
More user information should be included in user mode scoreboard export


ColdHeat:
User mode scoreboard CSV exports are missing the user IDs as well as user emails. 

https://github.com/CTFd/CTFd/blob/2c32791c241cd9ee6bb3183ed2221154e70a927f/CTFd/utils/csv/__init__.py#L117-L126

joelebutler:
I will work on this later tonight if nobody else has gotten to it yet.

ColdHeat:
Closed by #2136

PR #2136, opened 2022-06-05
Add info to user scoreboard csv

joelebutler:
Added user id and user email fields to the user mode scoreboard csv export as per issue #2133.


ColdHeat:
Thanks!

Issue #2131, opened 2022-05-25
Disable Imports on SQLite database provider


ColdHeat:
I don't really want to implement support for migrations on SQLite. Self hosted users should use docker-compose instead of just the Docker image. So I'd rather simply disallow imports if you're using SQLite.

ColdHeat:
Related Issues: #2121, #2130

huornlmj:
@ColdHeat who maintains the current public Docker image?  Is it this group or an independent 3rd party?  if it's this group then I would suggest removing SQlite from the Docker image if a more robust RDMS is required for upgrade capability.

Also by any chance is this related to https://github.com/CTFd/CTFd/issues/2130 or https://github.com/CTFd/CTFd/issues/2121 ?

PR #2139, opened 2022-06-05
Disable Imports on SQLite database provider

joelebutler:
As per issue #2131, the page will detect when a SQLite database is being used. Instead of displaying a file upload and submit button, it will then display an error saying "Importing is disallowed when using SQLite databases"


ColdHeat:
Thanks for making this! Instead of disabling imports at the web UI level I think a better idea would be to disable them at the import function level. That way you would be able to disable the CLI importer as well as the web UI.

joelebutler:
Is this what you meant? It disables it at the CLI importer but also leaves the web UI showing that you can't.

Issue #2128, opened 2022-05-25
Move attempt after CTF configuration to somewhere else


ColdHeat:
<img width="687" alt="image" src="https://user-images.githubusercontent.com/166333/170346949-42d6326f-db1d-409e-9ca6-81cbe689993d.png">

This configuration has like no discoverability. This should be moved somewhere else.

PR #2140, opened 2022-06-06
Moved attempt after CTF setting to settings tab

joelebutler:
Moved "Allow challenges to be viewed (no submissions are recorded) after the CTF End Time." checkbox setting from the time config panel to the settings config panel, to improve discoverability as per issue #2128.


ColdHeat:
I haven't looked at it in an actual CTFd instance yet but I think instead of a checkbox it should be a dropdown/select box.

ColdHeat:
I think you could copy over the behavior from something like the "Verify Emails" setting which is either Disabled or Enabled. It might require changes in Python as well but I'm not sure. I didn't really completely flesh out the idea, I just found this option really hard to find.

Issue #2144, opened 2022-06-07
A team created by an admin in the user facing interface should be hidden by default


ColdHeat:
Admins have a lot of power in CTFd and they can see things before an event opens. In Teams Mode when an Admin creates their own team the team is not automatically marked as hidden. This can accidentally reveal information. 

Insted when an Admin creates a team through the normal Team Creation flow that team should be defaulted to Hidden.

PR #2150, opened 2022-06-29
Hidden admin team by default

Aides2593:
When an Admin creates a team through the normal Team Creation flow that team should be defaulted to Hidden
Fix issue #2144


ColdHeat:
Nice! This looks appropriate but do you think you can add a test for this behavior?

PR #2155, opened 2022-07-31
validate value type is string before using strip

hakkasuru:
When a checkbox field that is required and editable is created after user has created account. Existing users are unable to update their profile as an internal server error is produced.

The internal server error is due to the required field being a boolean and .strip() is invoked on a boolean with causes an error.

The fix is to validate the type of value to be string before checking if the field is empty.

### DIFF
--- a/CTFd/schemas/teams.py
+++ b/CTFd/schemas/teams.py
@@ -264,10 +264,13 @@ class TeamSchema(ma.ModelSchema):
                     field_id=field.id, team_id=current_team.id
                 ).first()
 
-                if field.required is True and value.strip() == "":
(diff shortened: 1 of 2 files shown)


ColdHeat:
This looks reasonable, thanks for reporting the issue. I think perhaps there needs to be an extra case for if the value isn't a string? I haven't looked at it very in-depth yet though.

hakkasuru:
@ColdHeat I tried with latest version, if the checkbox field is required and it is unchecked it will show the note and not allow the user to submit. For textfields, if the textfield is empty it will also show the note and not allow the user to submit.

Issue #1370, opened 2020-04-30
Plugins 'Config.json' parser only set up for one URL.


offsecginger:
**Environment**:

  - CTFd Version/Commit: 2.3.3
  - Operating System: Windows/Linux
  - Web Browser and Version: Chrome

**What happened?**
The config.json file is used for adding URLs to the Admin section of the CTFd instance based on plugin requirements.
The current code that parses it:
```
def get_configurable_plugins():
    Plugin = namedtuple("Plugin", ["name", "route"])

    plugins_path = os.path.join(app.root_path, "plugins")
    plugin_directories = os.listdir(plugins_path)

    plugins = []

    for dir in plugin_directories:
        if os.path.isfile(os.path.join(plugins_path, dir, "config.json")):
            path = os.path.join(plugins_path, dir, "config.json")
            with open(path) as f:
                plugin_json_data = json.loads(f.read())
                p = Plugin(
                    name=plugin_json_data.get("name"),
                    route=plugin_json_data.get("route"),
                )
                plugins.append(p)
        elif os.path.isfile(os.path.join(plugins_path, dir, "config.html")):
            p = Plugin(name=dir, route="/admin/plugins/{}".format(dir))
            plugins.append(p)

    return plugins
```
is not configured to handle multiple entries.

offsecginger:
Solution:
```
def get_configurable_plugins():
    Plugin = namedtuple("Plugin", ["name", "route"])

    plugins_path = os.path.join(app.root_path, "plugins")
    plugin_directories = os.listdir(plugins_path)

    plugins = []

    for dir in plugin_directories:
        if os.path.isfile(os.path.join(plugins_path, dir, "config.json")):
            path = os.path.join(plugins_path, dir, "config.json")
            with open(path) as f:
                plugin_json_data = json.loads(f.read())
                for d in plugin_json_data:
                    p = Plugin(
                        name=d.get("name"),
                        route=d.get("route"),
                    )
                    plugins.append(p)
        elif os.path.isfile(os.path.join(plugins_path, dir, "config.html")):
            p = Plugin(name=dir, route="/admin/plugins/{}".format(dir))
            plugins.append(p)

    return plugins
```
This requires the config.json file to always be a list. So a Sample Config.json file would look like this:

```
[
    {
        "name" : "Config",
        "route" : "/path/to/config"
    }
]
```

ColdHeat:
[...]

PR #2161, opened 2022-08-06
 Made get_configurable_plugins work with multiple plugin entries as a list

dustinwloring1988:
Made get_configurable_plugins work nicely with config.jsons that include multiple plugin entries as a list. 
I do not take credit for this as John Hammond did this and I just did not want to do this every time I clone the project.


ColdHeat:
Related to #1370 but better because it doesn't break backwards-compat.

PR #2171, opened 2022-08-24
Fix missing port when redirecting

vinhjaxt:


ColdHeat:
This honestly doesnt seem like the right idea. Is the point to support running on non-standard ports? I don't think this change needs to be included in the CTFd repo. You could just use your own nginx config right?

PR #2181, opened 2022-09-02
Adds Supports For Arch Linux

Seew0:
adds prepare-arch.sh for  arch and arch based distros


Seew0:
please look

ColdHeat:
You can always run CTFd through Docker. I don't really see the point of supporting this as we're moving towards Docker image being the primary method of deploying CTFd.

PR #2185, opened 2022-09-15
Redirect users to team creation before event start

janosdebugs:
This change redirects users to the team creation page if they access a during_ctf_time_only page before the CTF starts. This results in a better experience since users can start assembling their team immediately.

### DIFF
--- a/CTFd/utils/decorators/__init__.py
+++ b/CTFd/utils/decorators/__init__.py
@@ -29,8 +29,11 @@ def during_ctf_time_only(f):
                     error = "{} has ended".format(config.ctf_name())
                     abort(403, description=error)
             if ctf_started() is False:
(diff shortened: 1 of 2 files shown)


ColdHeat:
I thought we had already done this but this is a good change!

ColdHeat:
Ah I see this was handled specifically in the challenges endpoint: https://github.com/CTFd/CTFd/blob/4793d95338935686ef005c24ebb6606e1be7403b/CTFd/challenges.py#L30-L31

It should be fine to put this into the decorator but I'm wondering why it didn't already work for you? 

Also could you just fix the lint in your test?

ColdHeat:
I see that this behavior actually only takes place if the user directly browses to /challenges sometime after creating an account. A user that freshly registers is prompted to go through team creation. 

I still think this is valid, just mentioning.

PR #2196, opened 2022-10-09
Add autofocus to text fields on authentication pages

tyrumus:
This change adds the `autofocus` attribute to the first text field on each authentication-related page. This results in a better experience since users won't have to click the field, or cycle through several links on the page with the Tab key before reaching it.


ColdHeat:
Hi there, the premise looks fine, can you fix these lints though?

tyrumus:
Linting issues should be resolved now. Thank you for your time.

Issue #666, opened 2018-08-20
Plugin registration order


ColdHeat:
Plugins can end up conflicting on load order due to glob. This is probably something that we should govern. Perhaps sorting the resulting glob result or something better.

PR #2225, opened 2022-11-12
improve plugin loading mechanism

frankli0324:
this allows plugins to specify their own dependencies, instead of passive load order controlling with alphabetical ordering
the pr also saves plugin modules into the app object for future referencing
ref: https://github.com/CTFd/CTFd/issues/666


frankli0324:
alright I admit. I was working on the i18n, and I thought, plugins should provide their own translations. then I tried exposing their module objects, then I realized that the previous "dependency controlling" was not what it should be like, then I thought, how should I write tests for `PluginDependencyException`, then I started working on a plugin system following the suggestions from pypa [here](https://setuptools.pypa.io/en/latest/userguide/entry_point.html). now I stuck on how I should install the plugin packages before testing them. [`b6c874f` (#2225)](https://github.com/CTFd/CTFd/pull/2225/commits/b6c874fc49c58bd38172c80d8bfc0b4ad050b2f0) this seems obviously wrong.

frankli0324:
I tried my best not to break the existing plugin system. hope I didn't

ColdHeat:
I get the idea here and see how it could be useful but I feel like this complicates a pretty simple solution. I prefer to view the plugins as self contained instead of relying on each other. Perhaps I will encounter this issue while working on translations but I think I will need to look at this first before weighing in on it.

PR #2228, opened 2022-11-17
Add support for Docker multi-stage builds

eduardo010174:
Separate docker build in two stages make advantages for discard several binaries, builder, compiler and dev dependencies not needed for production usage.

For transfer python dependencies from first stage to second stage of project, I'm using python venv.

On second stage, only install runtime dependencies and copy source.
Final image reduce size from 648MB for 398MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
-FROM python:3.9-slim-buster
+FROM python:3.9-slim-buster as build
+
 WORKDIR /opt/CTFd
-RUN mkdir -p /opt/CTFd /var/log/CTFd /var/uploads
 
 # hadolint ignore=DL3008
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


eduardo010174:
It's ready for merge.

eduardo010174:
I reorganize code, commit messages, and squash commits into one.
After you have time, review code and send feedback.

Dockerfile
```diff
@@ -10,31 +10,46 @@ RUN apt-get update \
...
+    && chown -R 1001:1001 /var/log/CTFd /var/uploads \
+    && chmod +x /opt/CTFd/docker-entrypoint.sh
+
+COPY --from=build /opt/venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
 
 USER 1001
 EXPOSE 8000
-ENTRYPOINT ["/opt/CTFd/docker-entrypoint.sh"]
+ENTRYPOINT ["/opt/CTFd/docker-entrypoint.sh"]
```
ColdHeat, line 55: Add new line back?

Dockerfile
```diff
@@ -10,31 +10,46 @@ RUN apt-get update \
...
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        libffi6 \
+        libssl1.1 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY --chown=1001:1001 . /opt/CTFd
+
+RUN useradd \
```
ColdHeat, line 41: What is the advantage of switching to useradd?

Dockerfile
```diff
@@ -10,31 +10,46 @@ RUN apt-get update \
...
-    --gecos "" \
+
+FROM python:3.9-slim-buster as release
+WORKDIR /opt/CTFd
+
+# hadolint ignore=DL3008
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        libffi6 \
+        libssl1.1 \
```
ColdHeat, line 35: Do we specifically need to pin versions here or could we just let the package manager pick the latest?

ColdHeat (changes requested):

Dockerfile
eduardo010174: adduser is a wrapper for useradd binary.
First, useradd, by default, not generate home directory. It's not needed in this application.
Second, adduser not have option to remove log-init files. useradd have option to remove lastlog and faillog files, --no-log-init.
Third, using Dive application size is reduced for this change is 440k to 5.6k.

Dockerfile
eduardo010174: Interesting question.
On apt repository of Debian latest tag only have for Foreign Function Interface library (development files).
I not think justification for this choice from apt repository. On Ubuntu, it's the same.

For libffi example link
https://packages.debian.org/source/buster/libffi
Runtime package is named libffi6 and dev package is libffi-dev.

For libssl exatly same.
https://packages.debian.org/buster/libssl-dev
https://packages.debian.org/buster/libssl1.1

Extra:
On apk repository of Alpine have latest tags for dev and runtime packages.

eduardo010174: But for check what packages have on apt reposity use
```
apt-cache search libffi
apt-cache search libssl
```

For libffi example
```
$ apt-cache search libffi
libffi-dev - Foreign Function Interface library (development files)
libffi6 - Foreign Function Interface library runtime
libffi6-dbg - Foreign Function Interface library runtime (debug symbols)
ffindex-dbg - simple index/database for huge amounts of small files (debug)
libbfio-dbg - Library to provide basic input/output abstraction (debug)
libffindex0 - library for simple index/database for huge amounts of small files
libffindex0-dev - library for simple index/database for huge amounts of small files (development)
libjffi-java - Java Foreign Function Interface
libjffi-jni - Java Foreign Function Interface (JNI library)
```

Dockerfile
ColdHeat: Sounds good to me

Dockerfile
ColdHeat: I think this makes sense. I just need to spend some time to test this out because there may be some edge case that we are missing.

ColdHeat: I have tested this out and it seems to be working properly.

PR #2229, opened 2022-11-18
Bump CTFd dependencies

eduardo010174:
Bump bcrypt, gevent, greenlet, python-geoacumen-city, requests.

There are probably more dependencies that need to be updated.

The update is important because new versions of gevent and greenlet support musl.
In the future I will create another docker image based on alpine.
Alpine uses musl.


ColdHeat:
Thanks! This looks good. I'm not so keen on Alpine though. In my experience Alpine hasn't been the greatest base image. I am perhaps willing to keep it as a seperate tagged image but I'm not really sure I see the advantage.

Issue #2231, opened 2022-11-24
Database URLs not compatible with some deployment methods


SmylerMC:
The way database and Redis settings are passed as a single URL isn't really modular and may cause problems in some cloud deployments.

For example, when deploying on Kubernetes, I might have my database host, port, username and password stored in separate configuration objects shared with other deployments (typically the database deployment). With the current configuration settings, I have to store the same values in different configuration objects because I need to provided them to CTFd as one URL.
It would be simpler if I could map each configuration value to a separate environment variable for CTFd.

Said variables could be
- `DATABASE_PROTOCOL`
- `DATABASE_HOST`
- `DATABASE_PORT`
- `DATABASE_USER`
- `DATABASE_PASSWORD`
- `DATABASE_NAME`

Forward compatibility could be kept by simply falling back to `DATABASE_URL` if the configuration is invalid.

I don't know whether such a change would be acceptable ? If so, I could implement it myself if needed.

ColdHeat:
I believe I have run into this before so it's not unreasonable. I have seen people work around it but I could see us support the idea.

I would accept a PR for this. I would still prioritize the `DATABASE_URL` but if that is omitted it can load  the envvars you are suggesting.

PR #2237, opened 2022-12-11
Add individual DATABASE_* options, as an alternative to DATABASE_URL

sudoBash418:
## Description / Motivation

This PR is meant to resolve issue #2231 by adding the following config settings:
- `DATABASE_PROTOCOL`: SQLAlchemy DB protocol (+ driver, optionally)
- `DATABASE_USER`: Username to access DB server with
- `DATABASE_PASSWORD`: Password to access DB server with
- `DATABASE_HOST`: Hostname of the DB server to access
- `DATABASE_PORT`: Port of the DB server to access
- `DATABASE_NAME`: Name of the database to use

As explained in the linked issue, these new settings are meant as an alternative to the monolithic `DATABASE_URL` setting, to allow each individual component to be controlled separately via environment variables.  
One of the main use cases is in cloud deployments, where it's common to treat the database credentials as opaque secrets, and store the rest of the connection details in plaintext.

## Implementation details

As I've implemented it, SQLite databases aren't really supported (`DATABASE_HOST` is required to use these settings), but I also don't see the use case for someone using these new settings with a SQLite DB, as opposed to `DATABASE_URL`.

If specified, `DATABASE_URL` takes full precedence over any of these new settings, as per ColdHeat's request in #2231.

I just used the setting names from the linked issue, and tried to use reasonable defaults for undefined settings.
I've also used the URL class from SQLAlchemy in order to build the URL - hopefully that's not an issue. I tried writing a  version using string concatenation, but it was both harder to understand and came with some annoying limitations (ex. if someone changed the DB type, the default port wouldn't match up).

One last note: I'm not sure if the intent of the linked issue was to add `REDIS_*` settings as well; if that's the case, I can either add that into this PR, or make a second PR similar to this one for those settings.

### DIFF
--- a/CTFd/config.ini
+++ b/CTFd/config.ini
@@ -26,9 +26,37 @@ SECRET_KEY =
 # The URI that specifies the username, password, hostname, port, and database of the server
 # used to hold the CTFd database.
 #
-# If a database URL is not specified, CTFd will automatically create a SQLite database for you to use
+# If neither this setting nor `DATABASE_HOST` is specified, CTFd will automatically create a SQLite database for you to use
 # e.g. mysql+pymysql://root:<YOUR_PASSWORD_HERE>@localhost/ctfd
 DATABASE_URL =
+
+# DATABASE_HOST
+# The hostname of the database server used to hold the CTFd database.
+# If `DATABASE_URL` is set, this setting will have no effect.
+#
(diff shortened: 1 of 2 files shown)


ColdHeat:
Thanks! I would also accept a change like this for `REDIS_URL` as well as I have seen this be an issue for that as well. 

Closes #2231

PR #2238, opened 2022-12-15
Add the possibility to publish solutions to challenges

adb014:
At the end of competitive phases of CTFs we'd like to be able to publish solution and put the CTF on pause. This allows the participants to be able to continue to try the challenges, with knowledge of the solutions, to continue the educative part of the CTF.

This commit adds the possibility  add solutions to the challenges, hide them during the competitive phase and publish them when the CTF is finished.

### DIFF
--- a/CTFd/api/v1/teams.py
+++ b/CTFd/api/v1/teams.py
@@ -84,6 +84,7 @@ class TeamList(Resource):
                         "country": "country",
                         "bracket": "bracket",
                         "affiliation": "affiliation",
+                        "email": "email",
                     },
                 ),
                 None,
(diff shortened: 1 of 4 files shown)


ColdHeat:
I like the general idea behind this but I'm not sure how complete the implementation has to be. It's simple to just have some text but eventually people will want to add in pictures and such. I will circle back on this.

zevlag:
I personally feel that a simple solution like this is a good start.  If it needs more advanced features, they can be iterative improvements.

adb014:
Frankly I don't understand what is the problem. My proposal uses the same markdown editor as the challenge text itself, and you can upload images, etc to the media library and include them in the solution text. So images, etc are already included in the same manner as elsewhere in CTFd

adb014:
Closed this request and will open a new one updated for 3.6.0

PR #2244, opened 2023-01-18
Fix display error on mobile devices

web1n:
Fix display error on mobile devices

### DIFF
--- a/CTFd/themes/admin/templates/challenges/challenges.html
+++ b/CTFd/themes/admin/templates/challenges/challenges.html
@@ -65,7 +65,7 @@
 	</div>
 
 	<div class="row">
-		<div class="col-md-12">
+		<div class="col-md-12 table-responsive">
 			<div>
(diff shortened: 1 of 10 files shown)


ColdHeat:
Hi there, thanks for filing this. I think this is a sstep in the right direction but I think a better fix might be to set overflow-x on the main "col-md-12" ? Just setting this d-none doesn't seem like it really helps with viewing this on my phone.

web1n:
> Hi there, thanks for filing this. I think this is a sstep in the right direction but I think a better fix might be to set overflow-x on the main "col-md-12" ? Just setting this d-none doesn't seem like it really helps with viewing this on my phone.

On mobile devices, the user's email address is not displayed, but the table header is still displayed. This caused problems with different number of columns in thead and tbody. So I added the d-none class to fix that.

Regarding the responsiveness problem on mobile, I think we can add table-responsive class to solve it. This fix works fine for me.

PR #2258, opened 2023-02-14
Add Traefik support

cao:
Add Docker Compose file to use Traefik as a load-balancer/frontend instead of nginx.

Traefik supports ACME Let's Encrypt natively, making it more easy to acquire trusted TLS certificates, which is currently not supported by the default Docker Compose file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ RUN useradd \
     && chown -R 1001:1001 /var/log/CTFd /var/uploads \
     && chmod +x /opt/CTFd/docker-entrypoint.sh
 
-COPY --from=build /opt/venv /opt/venv
+COPY --chown=1001:1001 --from=build /opt/venv /opt/venv
 ENV PATH="/opt/venv/bin:$PATH"
 
 USER 1001


ColdHeat:
Hello! 

I actually have experience with Traefik and in my experience it's a bad choice. Specifically with regards to TLS, [distributed LetsEncrypt certs is a paid feature from Traefik afaik](https://traefik.io/pricing/). You may want to consider [Caddy](https://caddyserver.com/docs/automatic-https#storage) instead although I do not have any major experience with LetsEncrypt on Caddy. 

For example, if you have 3 Traefik servers acting as your load balancers and you then attempt to get a LE cert dynamically one server will start the process and then the the challenge response could end up going to any of the 3 servers. The other 2 servers won't know what to do so your cert allocation may fail or require multiple attempts. If you're seeing different behavior let me know but I am not a huge fan of Traefik because of this. 

I am not against merging this but I would prefer if instead of this being a copied file if it was something that could inherit from the main docker-compose file. I haven't really dug into how this would work but it does seem like docker-compose supports the idea: https://docs.docker.com/compose/extends/. I would also probably prefer if this was Caddy but that's less important.

PR #2261, opened 2023-02-15
Confirm Old Team Password 

mattcurts:
Was unclear where you had to enter the old password


ColdHeat:
Maybe "Confirm Old Team Password"? Not sure if this is an improvement or not. Maybe there's some example online to copy that has similar UX?

mattcurts:
yeah that would be more clear!

ColdHeat:
Can you update your PR to say "Confirm Old Team Password"?

Issue #2265, opened 2023-02-27
Microsoft Azure Blob Storage integration for the CTFd Upload utility


BlackSud0:
At the moment, CTFd does not support Azure Blob Storage; instead, it only supports S3-based object storage services (AWS S3, Google Cloud Storage, and DigitalOcean Spaces). The goal of this issue is to add a new Azure upload provider to the CTFd uploads utility. To offer flexibility to those who have their own infrastructure in Azure to use Azure Blob Storage instead of AWS or GCP.

ColdHeat:
Hi there, thank you for creating a PR for this. What's unclear to me is shouldn't Azure's blob storage be API compatible with S3? I am pretty sure most object storage offerings are.

BlackSud0:
Azure blob storage does not have API compatibility with S3. When I tried it, the s3 uploader failed to authenticate with Azure blob and I received `botocore.exceptions.ClientError: AuthenticationFailed` . That is why I prefer to use the official Azure client library.
```bash
  File "/home/username/www/CTFd/CTFd/api/v1/files.py", line 96, in post
    obj = uploads.upload_file(file=f, **request.form.to_dict())
  File "/home/username/www/CTFd/CTFd/utils/uploads/__init__.py", line 31, in upload_file
    location = uploader.upload(file_obj=file_obj, filename=file_obj.filename)
  File "/home/username/www/CTFd/CTFd/utils/uploads/uploaders.py", line 122, in upload
    self.s3.upload_fileobj(file_obj, self.bucket, dst)
  File "/home/username/.local/lib/python3.9/site-packages/boto3/s3/inject.py", line 539, in upload_fileobj
    return future.result()
  File "/home/username/.local/lib/python3.9/site-packages/s3transfer/futures.py", line 106, in result
    return self._coordinator.result()
  File "/home/username/.local/lib/python3.9/site-packages/s3transfer/futures.py", line 265, in result
    raise self._exception
  File "/home/username/.local/lib/python3.9/site-packages/s3transfer/tasks.py", line 126, in __call__
    return self._execute_main(kwargs)
  File "/home/username/.local/lib/python3.9/site-packages/s3transfer/tasks.py", line 150, in _execute_main
    return_value = self._main(**kwargs)
  File "/home/username/.local/lib/python3.9/site-packages/s3transfer/upload.py", line 692, in _main
    client.put_object(Bucket=bucket, Key=key, Body=body, **extra_args)
[...]

PR #2266, opened 2023-02-27
Add a new upload provider for Azure Blob Storage

BlackSud0:
|Related issue|
|---|
| #2265|

## Description
Azure Blob is the Microsoft equivalent to Amazon’s S3-based object storage services. Within that, a “blob” is like a bucket as the framework for retention of objects. You can upload, download, list, move and carry out other commands in Azure Blobs using Azure Storage client library which is available for multiple languages that include .Net, Java, Node.js, Python, PHP and Ruby.

**Blob storage offers three types of resources:**

* The storage account
* A container in the storage account
* A blob in the container

The following diagram shows the relationship between these resources:

image

## Configuration options

`AZURE_STORAGE_ACCOUNT_NAME` : The storage account name to authenticate to the Azure container.
`AZURE_STORAGE_ACCESS_KEY` : The account key, also called shared key or access key.
`AZURE_STORAGE_CONTAINER` : The name of the container that the blob is in, blob container is like an S3 bucket.
`AZURE_STORAGE_ENDPOINT` : The URL to the blob storage account.

## Does this introduce a breaking change?

```
[ ] Yes
[x] No
```

## Tests

- Each of the features listed below passed all tests without any errors.
- [x] Upload files.
- [x] Download files.
- [x] Delete files.
- [x] Store files (CTF import).
- [x] Sync the blob storage with local storage (CTF export).

### DIFF
--- a/CTFd/config.ini
+++ b/CTFd/config.ini
@@ -170,7 +170,7 @@ MAIL_PROVIDER =
 [uploads]
 # UPLOAD_PROVIDER
 # Specifies the service that CTFd should use to store files.
-# Can be set to filesystem or s3
+# Can be set to filesystem or s3 or azure
 UPLOAD_PROVIDER =
 
 # UPLOAD_FOLDER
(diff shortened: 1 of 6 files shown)


ColdHeat:
I thought for a bit about this PR and I won't be accepting it into the main codebase at this time. In the future if I see a need for this PR I will merge it in. 

The rationale is that s3 has cemented itself as the defacto cloud provider object storage interface. Azure should accept this reality and add a compatibility layer for Azure Blob Storage. Every other object storage provider has an S3-like API so why should we special case Azure.

In addition, there are online solutions for adding S3 compatibility via applications like https://github.com/gaul/s3proxy. 

In the meantime, this functionality could be added as a CTFd plugin if you wish.

gui28347:
Considering this small change and benefit of all CTFd + Azure users can benefit this; I think it would be really nice if you reconsider it.

Just to remember AWS still #1 with 31% market share and Azure 25% market share according with multiple different reports as of June/2024 such as: https://www.statista.com/chart/18819/worldwide-market-share-of-leading-cloud-infrastructure-service-providers/

So getting numbers rather than perspective, I think covering the top 3 providers AWS, Azure , Google its a good thing for all products, that way you are not forcing users to use one or another, they will most likely be using one of the 3 top providers.

I don't think is practical to Azure or Google implement S3 layer because Amazon has 31%. of market share What happen if in 2030 Azure becomes #1 , is AWS going to rewrite their API spec to be compatible with Azure ?! 

Different cloud services, different services, different specs, this is the world we have today, users should have freedom to select which one they think is best for their business considering features and cost.

PR #2268, opened 2023-03-09
Cache problem fix

GramThanos:
The caching of the API that returns the scoreboard does not take into that the given parameters may have changed.

### DIFF
--- a/CTFd/cache/__init__.py
+++ b/CTFd/cache/__init__.py
@@ -51,6 +51,7 @@ def make_cache_key(path=None, key_prefix="view/%s"):
     cache_key = key_prefix % path
     return cache_key
 
+
 def make_cache_key_with_args(path=None, key_prefix="view/%s", view_args=None):
     """
     This function ensures that view arguments are taken into account when generating a cache key


ColdHeat:
I actually have a need for the function you created so thanks for filing this! I do wonder if there's a way we could make use of the original query_string parameter but I admit that I haven't dug too much into the PR yet.

GramThanos:
Let me describe some things:

Regarding the `query_string`, the problem is that when you define it will change the flow of the https://github.com/pallets-eco/flask-caching/blob/ed4c2556f42a610854ab4bcc784dca5534000f38/src/flask_caching/__init__.py#L477 function, and thus the custom function for generating the cache key will not be called.

This is the reason why I decided to script a new `make_cache_key` function instead. Furthermore, since the `query_string` takes into consideration all the parameters given, using only the parameters that the backend uses it is better.

But my changes also introduce a new problem. Now the cache key includes a user defined input thus it should be used with caution and it is not always predictable. Examples:

| URL | Old cache key | New cache key |
| :----- | --------------- | ---------------- |
| `/api/v1/scoreboard/top/1` | `view/api.scoreboard_scoreboard_detail` | `view/api.scoreboard_scoreboard_detail_count_1` |
| `/api/v1/scoreboard/top/10` | `view/api.scoreboard_scoreboard_detail` | `view/api.scoreboard_scoreboard_detail_count_10` |
| `/api/v1/scoreboard/top/100` | `view/api.scoreboard_scoreboard_detail` | `view/api.scoreboard_scoreboard_detail_count_100` |
```

In this particular case, the user input is always `int` thus I don't see any way that it could result in a problem. Though, right now there isn't a way to clear all these caches in a wildcard way, thus, I set the code to clean only the `view/api.scoreboard_scoreboard_detail_count_10` that is the default used by the app front-end. External apps using other counter will have to wait the other cache keys to expire.

PR #2273, opened 2023-03-28
Prevent race conditions on /healthcheck

SmylerMC:
In a high availability deployment scenario, two clients may make a request on /healthcheck at the exact same time, which can lead to check_config returning False if the second requests changes the 'healthcheck' cache key before the first one has had time to fetch the value it had set.

A solution to counter this is to ensure different keys are used for each healthcheck.

### DIFF
--- a/CTFd/utils/health/__init__.py
+++ b/CTFd/utils/health/__init__.py
@@ -1,4 +1,4 @@
-from time import monotonic_ns
+from time import time
 
 from flask import current_app
 from sqlalchemy_utils import database_exists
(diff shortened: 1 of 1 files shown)


ColdHeat:
Thanks for filing this. I like the general idea but I'd rather not have extra cache keys lying around. I think the choice to use monotonic was likely a mistake because the precision is too high. Perhaps if you could use time.time() instead and maybe round to the nearest 5 seconds?

SmylerMC:
The latest commit implements your suggestion, using the same key for all requests but rounding the value to the nearest 5.

PR #2276, opened 2023-04-02
Fixes double loading of plugins through serve.py

### LINKED ISSUE #361: Flask reloader causes CTFd to load plugins twice with serve.py
Flask will cause CTFd to load plugins twice if the reloader is on. This doesn't really affect a production instance but for a serve.py it can be a little gotcha. 

Perhaps we can implement a fix from here: 
https://stackoverflow.com/questions/9449101/how-to-stop-flask-from-initialising-twice-in-debug-mode

SouvikMP:
This change resolves #361 by disabling the reloader, loading the plugins twice when running the app through serve.py.
Signed-off-by : Souvik Mondal <tukaiemondal1@gmail.com>

### DIFF
--- a/serve.py
+++ b/serve.py
@@ -40,4 +40,4 @@ if args.profile:
     toolbar.init_app(app)
     print(" * Flask profiling running at http://127.0.0.1:4000/flask-profiler/")
 
-app.run(debug=True, threaded=True, host="127.0.0.1", port=args.port)
+app.run(debug=True, threaded=True, host="127.0.0.1", port=args.port, use_reloader=False)


ColdHeat:
This disables the reloader correct? So reloading the application when the code changes wouldn't work right? This is kind of the purpose of the `serve.py` server so I'd rather keep that in. If you figure out some way to fix this without disabling the reloader that would be the right solution.

SouvikMP:
True. 
From what I understand from here https://flask.palletsprojects.com/en/2.2.x/debugging/ and going through a multiple stackoverflow posts. The reloader has a different instance that checks for file changes and reloads when it detects a change. In this case it seems there is a unwanted change in the files while loading it. I tried to track it using the terminal and file event monitor but to no progress. May me I am missing something here but this is what I found
fileLogs

Issue #2263, opened 2023-02-16
Replace flake8 with ruff


ColdHeat:
https://github.com/charliermarsh/ruff 

This is so much better than flake8. We should be able to replace all our flake8 integrations with ruff for speed improvements and fewer dependencies.

alperb:
I think I can work on this, I'll be working on it if it's okay.

ColdHeat:
Closed in #2278

PR #2278, opened 2023-04-03
use ruff instead of flake8

alperb:
I've removed `flake8` from dependencies and put `ruff` instead. Also replaced the command in `lint` part in the makefile.

In ruff, I wasn't able to find the equivalent rules for `W503` and `E203`. So I've removed these ignores from ruff as well. I've avoided using a `.ruff.toml` file for the configuration.

I haven't removed `flake8-isort` from `development.txt` in case it's used by `isort`. Unless it's used it can be removed too.

This PR resolves issue #2263

### DIFF
--- a/CTFd/__init__.py
+++ b/CTFd/__init__.py
@@ -185,15 +185,15 @@ def create_app(config="CTFd.config.Config"):
         app.jinja_loader = jinja2.ChoiceLoader(loaders)
 
         from CTFd.models import (  # noqa: F401
-            db,
-            Teams,
-            Solves,
             Challenges,
             Fails,
-            Flags,
-            Tags,
             Files,
+            Flags,
+            Solves,
+            Tags,
+            Teams,
             Tracking,
+            db,
         )
 
         url = create_database()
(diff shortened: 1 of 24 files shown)


ColdHeat:
Thanks for filing this. `flake8-isort` isn't used by isort. If you're removing `flake8` you can remove all of the other `flake8` plugins as well.

ColdHeat:
I updated the PR to have the switches that I believe should cover the flake8 plugins and removed the extra flake8 plugins. If you could go through the new lints and either fix them if they're simple or simply ignore them.

alperb:
3 rules were causing linting problems, `I001`, `B905` and `B904`. 

`I001` is "Import block is un-sorted or un-formatted".

`B904` is "Within an except clause, raise exceptions with raise ... from err or raise ... from None to distinguish them from errors in exception handling".

`B905` is "zip() without an explicit strict= parameter". 

I've ignored all of them, `I001` and `B904` were trivial. However, fixing `B905` may cause backward incompatible behaviors so I simply ignored it too, if you want it fixed I can, no problem.

ColdHeat:
I think we can ignore B904 and B905 but I think for I001 can you ignore all the specific lines instead of ignoring it globally?

alperb:
I've ignored `I001` on the lines which caused linting errors and enabled it globally on `make lint`. There were 32 lines :) So I've changed a lot of files. I think it's okay now.

PR #2282, opened 2023-04-04
Feature Request - Allow Authors to retrieve their own challenges via API

sxmimran:
Feature Request:
That would be of great value if API allow authors to retrieve all their own challenges.


ColdHeat:
Was this code generated by something? Challenges don't have an author_id column right now but I admit it is a good idea. Secondly, any admin can pull down a challenge via the API. This PR doesn't seem to make much sense?

PR #2287, opened 2023-04-18
add: freeze ribbon to admin panel

### LINKED ISSUE #2283: Indicate in Admin Scoreboard whether or not there is a freeze in effect
We should indicate in Admin Scoreboard whether or not there is a freeze in effect and perhaps show the difference in ordering.

alperb:
This PR partially resolves #2283 . 

I've added a ribbon similar to the "A new CTFd version available" ribbon to the base template of the admin panel. I've aimed to show this information every time they view a page instead of just showing it only when admins view the scoreboard.

### DIFF
--- a/CTFd/themes/admin/templates/base.html
+++ b/CTFd/themes/admin/templates/base.html
@@ -114,6 +114,15 @@
 	</nav>
 
 	<main role="main">
+		{% if get_config('freeze') %}
+		<div class="container-fluid bg-info text-center py-2">
+			<div class="row">
+				<div class="col-md-12 text-light">
+					Scoreboard freeze in effect
+				</div>
+			</div>
+		</div>
+		{% endif %}
 		{% if get_config('version_latest') %}
 		<div class="container-fluid bg-warning text-center py-3">
 			<div class="row">


alperb:
In addition to this solution, I am planning to add an indicator for the ordering difference between the scoreboard now and the scoreboard at freeze time. 

I am planning to use a cache key to store the standings at the time of the freeze effect. Then every time scoreboard standings are requested with the `get_standings()` function, an extra field can be added to get the standing difference since the freeze time. This solution requires adding a new extra parameter to the `get_standings()` function.

I'll continue on top of this branch if the solution is okay.

alperb:
> In addition to this solution, I am planning to add an indicator for the ordering difference between the scoreboard now and the scoreboard at freeze time.
> 
> I am planning to use a cache key to store the standings at the time of the freeze effect. Then every time scoreboard standings are requested with the `get_standings()` function, an extra field can be added to get the standing difference since the freeze time. This solution requires adding a new extra parameter to the `get_standings()` function.
> 
> I'll continue on top of this branch if the solution is okay.

Another solution without polluting the settings update function, would be to get the current standings and freeze time standings and insert a difference column if function is called with `admin=True` parameter.

PR #2294, opened 2023-04-26
bump SQLAlchemy-Utils to version 0.41.0

### LINKED ISSUE #2290: Update SQLAlchemy-Utils
The current version of SQLAlchemy-Utils that we are using was yanked for some reason. We should update this soon.

skandix:
this fixes #2290, to bump the version down to 0.36.5


ColdHeat:
Hi thanks for filing this. Any reason that we couldn't go up in version number instead of down?

PR #2312, opened 2023-05-28
fix: Empty file produces errors

### LINKED ISSUE #2257: Empty core.min.js produces a lot of errors
**Environment**:

- CTFd Version/Commit: CTFd 2.5.1 commit 472010f
- Operating System: Linux (official dockerfile)
- Web Browser and Version: Chrome 109

**What happened?**
With a fresh installation of CTFd, gunicorn emit a lot of errors. It seems that gunicorn doesn't like to serve empty files. This is due to the core.min.js being empty.

This issue happens with any standard CTFD setup since it seems to be caused by the [core.min.js](https://github.com/CTFd/CTFd/blob/master/CTFd/themes/core/static/js/core.min.js) file being empty.

```
[2023-02-12 16:39:34 +0000] [19] [ERROR] Error handling request
Traceback (most recent call last):
  File "/opt/venv/lib/python3.9/site-packages/gunicorn/workers/base_async.py", line 113, in handle_request
    resp.write_file(respiter)
  File "/opt/venv/lib/python3.9/site-packages/gunicorn/http/wsgi.py", line 387, in write_file
    if not self.sendfile(respiter):
  File "/opt/venv/lib/python3.9/site-packages/gunicorn/http/wsgi.py", line 377, in sendfile
    self.sock.sendfile(respiter.filelike, count=nbytes)
  File "/opt/venv/lib/python3.9/site-packages/gevent/_socket3.py", line 495, in sendfile
    return self._sendfile_use_send(file, offset, count)
  File "/opt/venv/lib/python3.9/site-packages/gevent/_socket3.py", line 425, in _sendfile_use_send
    self._check_sendfile_params(file, offset, count)
  File "/opt/venv/lib/python3.9/site-packages/gevent/_socket3.py", line 470, in _check_sendfile_params
    raise ValueError(
ValueError: count must be a positive integer (got 0)
```

**Fix**
Filling the core.min.js file with at least one byte removes the exception.
image

msterhuj:
This PR resolves #2257 .

I juste added a comment  on file `CTFd/themes/core/static/js/core.min.js`

### DIFF
--- a/CTFd/themes/core/static/js/core.min.js
+++ b/CTFd/themes/core/static/js/core.min.js
@@ -1,0 +1,1 @@
+/* core.min.js */


ColdHeat:
This change will be overwritten whenever the file is re-generated. Changes to JS need to be made in their original source file not in the minified JS files. However I really don't understand, does this issue happen on the latest version of CTFd?

msterhuj:
Hello
Yes when the application tries to return this file it generates this error and not if the file contains text
Tested on latest version and on master branch

cha5126568:
This error also happens in the admin panel, so to avoid this issue, you need to fix `CTFd/themes/admin/static/js/core.min.js` as well as `CTFd/themes/core/static/js/core.min.js` in the same way.

ColdHeat:
I will accept a PR that includes the changes in the source files. Changes to just the minified JS files will be overwritten the next time the theme is regenerated.

cha5126568:
After doing some investigation, it seems that this bug is not directly caused by CTFd.

It's an issue under gunicorn, which is fixed in the main branch, but not in the latest release, 20.1.0.

https://github.com/benoitc/gunicorn/issues/2692

Instead of waiting for the new release of gunicorn, to avoid this issue, i think you need to change

```python
import glob
import os

static_js_dirs = [
    "CTFd/themes/core/static/js/**/*.dev.js",
    "CTFd/themes/admin/static/js/**/*.dev.js",
]

for js_dir in static_js_dirs:
    for path in glob.glob(js_dir, recursive=True):
        if path.endswith(".dev.js"):
            path = path.replace(".dev.js", ".min.js")
            if os.path.isfile(path) is False:
                open(path, "a").close()
```

to 
```
import glob
import os

static_js_dirs = [
    "CTFd/themes/core/static/js/**/*.dev.js",
    "CTFd/themes/admin/static/js/**/*.dev.js",
]

for js_dir in static_js_dirs:
    for path in glob.glob(js_dir, recursive=True):
        if path.endswith(".dev.js"):
            path = path.replace(".dev.js", ".min.js")
            if os.path.isfile(path) is False:
                open(path, "a").write("/**/").close()
```

in https://github.com/CTFd/CTFd/blob/23cdf85bf2a9755460c98d8028b017f53831c13e/webpack.config.js#L183

and recompile(?) webpack.

I have no experience with webpack, so I'll let someone else do the actual PR.

sxmimran:
Prior to the fix introduced in gunicorn version > 20.1.0, Added a webpack tweak to address the issue. The tweak involves checking if the file is empty and adding a comment, as suggested by @msterhuj. This ensures that we don't lose the comment when regenerating the theme.

Added Fix for empty core.min.js file error: #2365 
Thanks

PR #2315, opened 2023-06-05
Allow users to see their previous submissions

carsso:
new feature to allow users to see their previous submissions

Capture d’écran 2023-06-05 à 06 24 35
Capture d’écran 2023-06-05 à 06 23 47

A possible improvement could be to implement this as an option in the config ?


ColdHeat:
Thank you for the intial work on this. I will take over on this as I would like to see this feature in 3.8.

ColdHeat:
I think my initial reaction is that the endpoint should be under /api/v1/users/me and/or /api/v1/teams/me. 

I think putting it all under challenges doesn't make sense as it would imply that these are all submissions when it's scoped to the user/team. 

The submissions endpoint is admins only so it's not an ideal choice in my opinion.

carsso:
> Thank you for the intial work on this. I will take over on this as I would like to see this feature in 3.8.

Sure. It was a long time ago. Feel free to take my code as a base, or write something different.
Happy to know that idea will be implemented in an future release :)

ColdHeat:
The original reason this wasn't implemented is that CTFd was originally used in fairly public scenarios. So shoulder surfing would be a concern. That is less a concern for events held over the internet but I am unsure if whether it makes sense to have this be configurable or if it should be default.

ColdHeat:
In addition, this makes things like account compromise more of an issue.

ColdHeat:
I think this should be:

* Disabled by default and enabled by config
* There should be shoulder surfing prevention

ColdHeat:
Thanks for your work on this. I have now merged https://github.com/CTFd/CTFd/pull/2776 which implements the server side of this behavior.

Issue #2320, opened 2023-06-07
Score of particpants continues to lower after freeze time


Typhlos:
**Environment**:

- CTFd Version/Commit: 3.5.1
- Operating System: Kubernetes v1.25
- Web Browser and Version: Firefox 114

**What happened?**

The scoreboard was frozen at the end of a CTF but users were still allowed to make submissions that does not appear on the scoreboard as intended. However, the challenges are configured on dynamic and the points continues to lower with new submissions and participants see their score lower on the scoreboard even if it is frozen.

**What did you expect to happen?**

The score of dynamic challenge should not continue to lower with new submissions once the CTF has been frozen.

**How to reproduce your issue**

- Create a dynamic challenge
- Solve it
- Freeze the CTF
- Solve that challenge with another user
- See the score of the challenge lower
- See the scores of people having solved that challenge lower on the frozen scoreboard

**Any associated stack traces or error logs**

ColdHeat:
Thanks for reporting this. I'm not sure what the fix would be for this but I will take a look at some point. If you can submit a PR that would be helpful as well.

PR #2322, opened 2023-06-09
Use freeze time to calculate value of dynamic challenge

Typhlos:
Solves #2320 

I used the code used in [the scoreboard](https://github.com/CTFd/CTFd/blob/master/CTFd/api/v1/scoreboard.py#L93) to edit the function [_calculate\_value_](https://github.com/CTFd/CTFd/blob/master/CTFd/plugins/dynamic_challenges/__init__.py#L53) and filter the solves using freeze time and only keep the solves before freeze time.

Let me know if there is anything missing.

### DIFF
--- a/CTFd/plugins/dynamic_challenges/__init__.py
+++ b/CTFd/plugins/dynamic_challenges/__init__.py
@@ -8,6 +8,8 @@ from CTFd.models import Challenges, Solves, db
 from CTFd.plugins import register_plugin_assets_directory
 from CTFd.plugins.challenges import CHALLENGE_CLASSES, BaseChallenge
 from CTFd.plugins.migrations import upgrade
+from CTFd.utils import get_config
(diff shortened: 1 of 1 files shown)


ColdHeat:
I like this leads to the inverse of the original problem. For example if we freeze time and then the event continues to run, if we then unfreeze the challenge values will be the frozen values until their next recalculation.

Typhlos:
I'll try to see if it's not possible to recalculate all values on unfreeze

Typhlos:
@ColdHeat I modified the code so that on any modification of freeze time, all values are recalculated

Typhlos:
@ColdHeat please tell me if there's anything missing in the PR

ColdHeat:
Instead of having a special case on freeze, I think a better strategy would be to have a list of plugin functions that will be called when config is updated. This way the dynamic value plugin (or other plugins) can register a function to be called after config is updated.

PR #2325, opened 2023-06-12
Enable gzip for nginx

intrigus-lgtm:
This enables gzip for nginx.
Without gzip, our scoreboard takes 5 mb and with gzip, ~100kb.


ColdHeat:
Good change, thank you!

PR #2326, opened 2023-06-13
Additional configurations for boto3 client under the S3 uploader

beryxz:
During our last CTF, we used DigitialOcean Spaces that are S3-Compatible, but some additional configurations on the boto3 client where required to make use of their CDN feature. See https://docs.digitalocean.com/products/spaces/how-to/enable-cdn/. It requires the addressing style to be **virtual** and for the **final url** to be replaced with another one that  inclues `.cdn.`.   

This PR adds 2 new config variables to allow for this:

- `AWS_S3_ADDRESSING_STYLE`: Support for selecting the S3 addressing style. It defaults to "auto" as when it's not set, but can also be set to **virtual** and **path**. See https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html#access-bucket-console-ex.
- `AWS_S3_CUSTOM_DOMAIN`: Domain that replaces the default one in presigned download URLs.

This patch has been tested with version 3.5.2 (3fbfd816) of CTFd and the following S3-related configs:

- UPLOAD_PROVIDER=s3
- AWS_ACCESS_KEY_ID=<KEY_ID>
- AWS_SECRET_ACCESS_KEY=<ACCESS_KEY>
- AWS_S3_BUCKET=testbucket
- AWS_S3_ENDPOINT_URL=https://ams3.digitaloceanspaces.com
- AWS_S3_ADDRESSING_STYLE=virtual
- AWS_S3_CUSTOM_DOMAIN=testbucket.ams3.cdn.digitaloceanspaces.com
- AWS_S3_REGION=ams3


ColdHeat:
I like this PR but I think `AWS_S3_REPLACEMENT_URL` should probably just be `AWS_S3_REPLACEMENT_HOSTNAME`. 

I am also not sure I agree with the strategy of replacing the hostname like this. Is there any other s3 provider that makes use of this strategy?

ColdHeat:
Ah I do see that even AWS uses this concept a bit: https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#VirtualHostingCustomURLs

beryxz:
Right, `AWS_S3_REPLACEMENT_HOSTNAME` would be preferred considering that the expected value is an hostname.

I only know of DigitialOcean CDN and Virtual Hosts of AWS S3 buckets as you mentioned.

I'm open to alternative for replacing the url. I looked at the boto3 api but couldn't find an intended way for doing it

ColdHeat:
I spent some time investigating various S3 providers and this strategy of hostname replacement is unique to DigitalOcean. I don't think any other S3 provider uses it. 

The article I linked for S3 only really works for a specific use case. This setup as described also only works over HTTP which is a bad idea. The general recommendation for that is to use Cloudfront which would imply that you should use a Cloudfront signed URL instead of an S3 signed URL.

Cloudflare uses a different domain stratey for their Public Buckets but that ends up making the entire bucket public: https://developers.cloudflare.com/r2/buckets/public-buckets/ which doesn't really fit the use case. 

I do see that django-storages has a setting called `AWS_S3_CUSTOM_DOMAIN` which implements this behavior however

https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#cloudfront
https://github.com/jschneier/django-storages/blob/a87dea670dc7a6a27cc2cb83f4c90b0129e9998b/storages/backends/s3boto3.py#L312
https://github.com/jschneier/django-storages/blob/a87dea670dc7a6a27cc2cb83f4c90b0129e9998b/storages/backends/s3boto3.py#L542-L561

I was initially going to reject the idea here since only DO seems to use it but since django-storages uses it I am reconsidering...

ColdHeat:
Okay I think this is fine. It will need some changes regarding:

* HOSTNAME vs URL. I think this should just be called HOSTNAME and it should just replace the hostname instead of having to parse twice. 
* Currently it looks like config.ini is now required to have these files. This can be an issue with people who have existing config.ini files. Instead it should fallback to some safe value if the config value doesn't exist.

PR #2329, opened 2023-06-13
Bump dependencies

### LINKED ISSUE #2300: Bump depenencies to more recent versions
**Environment**: localhost

- CTFd Version/Commit: latest
- Operating System: Linux / Mac
- Web Browser and Version:

**What happened?**
I want to install a plugin that uses flask-dance. The latest flask dance depends on flask 2.0.3 while CTFd requires flask 1.1.2.
So I get a conflict when installing the plugin.

Upgrade ctfd to a newer flask >= 2.0.3

ColdHeat:
* Bump dependencies
* Closes #2300

### DIFF
--- a/CTFd/auth.py
+++ b/CTFd/auth.py
@@ -189,6 +189,14 @@ def register():
     if current_user.authed():
         return redirect(url_for("challenges.listing"))
 
+    num_users_limit = int(get_config("num_users", default=0))
+    num_users = Users.query.filter_by(banned=False, hidden=False).count()
+    if num_users_limit and num_users >= num_users_limit:
+        abort(
+            403,
+            description=f"Reached the maximum number of users ({num_users_limit}).",
+        )
+
     if request.method == "POST":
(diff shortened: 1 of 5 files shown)


ColdHeat:
I suspect it is a better idea to not go to the latest version of everything here.

khiemdoan:
I think we should update only one or some libraries at one time, step by step to avoid breaking.
For example:
- Update `flask-sqlalchemy` 2.5  because it starts to support `sqlalchemy` 1.4.
- Then update `sqlalchemy` to 1.4 and `flask`.
- Finally, update `flask-sqlalchemy` to latest.
With `pydantic`, we can update it independently because it doesn't change the interface too much.

ColdHeat:
@khiemdoan I don't know what happened with this PR but I tried to force push to remove my changes and it auto-closed. I've created a new PR at #2332 and I'm going to be going more slowly with the dependency changes.

PR #2340, opened 2023-06-22
fix challenge preview 

MilyMilo:
Fixes challenge preview by providing a default preview template in the admin theme, and pre-rendering it on the challenge page (instead of dynamically rendering the template provided by the theme).

This disables any theming for the preview window and is only a temporary workaround.


ColdHeat:
From some rough testing I do not think this works. If I have my theme as core-beta the preview window is still not operational.

ColdHeat:
I was playing around with this trying to reuse the existing challenge.html and I think I came up with a better strategy. Take a look at https://github.com/CTFd/CTFd/pull/2344.

ColdHeat:
I'm closing this in favor of #2344 since that supports custom themes and challenge plugins. This was the impetus to doing the investigation for that PR though so thank you for filing this one anyway!

PR #2351, opened 2023-07-03
fix challenge display with empty category

MilyMilo:
Having a challenge with empty category name causes this category to include all available challenges, instead of only challenges with an empty category.


ColdHeat:
Please move to the core-beta repo directly. 

I think perhaps the category check should be kept but maybe as `challenges !== null` ?

MilyMilo:
Moved here: https://github.com/CTFd/core-beta/pull/50

PR #2355, opened 2023-07-04
Improve GET /<team_id>/members

reteps:
- Make get `TeamMembers` not admins_only
- Return each member's points


reteps:
> assert r.status_code == 403
Will need to update test

reteps:
Closing as this would need to be rewritten after #2358

Issue #1504, opened 2020-06-21
Flask-restful/Flask-Restplus/Flask-Restx alternative investigation


ColdHeat:
This library is kind of in a sad state and the Python world has moved a lot since then. While the library works, it has a very complicated design and it's been very painful trying to marry it with other libraries. 

https://github.com/python-restx/flask-restx/issues/159

- Does not seem to support OpenAPI 3.0
- Very complicated design 
- reqpare library has been considered deprecated for like years

In the end, the only thing that would need to be cloned would be the api function naming in view_functions so that plugins that manipulate API endpoints refer to the same ones. Or that can be foregone in a major version. 

I've looked at alternatives like [apispec](https://github.com/marshmallow-code/apispec) but I'm not fond of the marshmallow libraries because of their tendency to introduce odd breaking changes. 

Depending on how much functionality needs to be ported, I think a worthwhile approach might be to just make my own api library. These libraries are all based on wrapping the Flask MethodView...

frankli0324:
would break a lot of plugins though

PR #2358, opened 2023-07-08
POC documentation generation

reteps:
As per #1504, you were interested in exploring an alternative documentation generation method. I have a POC up for converting a single challenges endpoint using [APIFlask](https://apiflask.com/api-docs/).

CleanShot 2023-07-08 at 19 53 12@2x

I upgraded Flask + dependencies while I was doing this.

The new syntax for decorating an endpoint is:

```python3
    @admins_only
    @challenges_blueprint.doc(description="Endpoint to create a Challenge object")
    @challenges_blueprint.input(ChallengeSchema, location="json_or_form")
    @challenges_blueprint.output(ChallengeDetailedSuccessResponse, 200, description="Success")
    @challenges_blueprint.output(APISimpleErrorResponse, 400, description="An error occured processing the provided or stored data")
    def post(self, data):
         # data is of type ChallengeSchema already!
  ```
  
  I think this has a few benefits:

-   It should be quick to reuse the existing schema validators for endpoints
-  Don't need to do Schema validation inside of each method first, `APIFlask` does it for you, and emits a 400 error.
-  Don't need to describe schemas twice (once for docs based on pydantic, once for marshmallow), just describe once in marshmallow.

### DIFF
--- a/CTFd/__init__.py
+++ b/CTFd/__init__.py
@@ -3,10 +3,11 @@ import os
 import sys
 import weakref
 from distutils.version import StrictVersion
+from apiflask import APIFlask
 
 import jinja2
-from flask import Flask, Request
-from flask.helpers import safe_join
+from flask import Request
+from werkzeug.security import safe_join
 from flask_babel import Babel
 from flask_migrate import upgrade
 from jinja2 import FileSystemLoader
@@ -45,7 +46,7 @@ class CTFdRequest(Request):
         self.path = self.script_root + self.path
 
 
-class CTFdFlask(Flask):
+class CTFdFlask(APIFlask):
     def __init__(self, *args, **kwargs):
         """Overriden Jinja constructor setting a custom jinja_environment"""
         self.jinja_environment = SandboxedBaseEnvironment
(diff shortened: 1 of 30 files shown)


reteps:
Requesting thoughts from @ColdHeat ! Excited to see if this could be a viable path I can continue developing. The end goal here is a complete openapi.json file that can be used to generated CTFd clients.

reteps:
Was thinking about this some more:

- The amount of work required to convert all usages of `request.data or request.json()` to just using the serialized input may be too high.
- If this amount of work is too high, it seems reasonable to use this as a better documentation generation layer, and then slowly transition to using the serialized input instead of request.

ColdHeat:
I like the core idea but I think it will be a very difficult PR to land. 

I would consider it for maybe 3.7 or 4.0 release but you would need to consider that you likely need to adjust APIFlask to have the same naming scheme for views as Flask-RestX or define a custom one so that we can break compatability in 4.0 and specify a specific naming scheme that we adhere to. 

In addition you cannot break any existing integration with things like CTFd.js. Or at least make the migration easy.

I personally don't really like Marshmallow and it was just something that worked at the time. These days I prefer the ideas in FastAPI and Pydantic. But I am okay with continuing to use Marshmallow if it makes sense. I had bounced around in my head that CTFd 4.0 might have both Flask and FastAPI simultaneously but I think APIFlask makes more sense.

If this is something you want to work on it's likely going to need to be broken out into a few PRs. An initial layer to start the API migration and then slowly converting every controller and model/schema into the updated format while making sure everything still works.

If this is something that you're willing to take on I support it and the POC. If this is something that you'd be interested in being paid to do (along with design reviews, deadlines, and status check-ins and such), we can [discuss something](https://ctfd.io/contact/).

PR #2372, opened 2023-07-21
Fix remove db from connection string

senacor-swr:
Problem: when starting CTFd with an empty database the ping check will fail, complaining that the database does not exist. Calling `set` with parameter value `None` does not actually remove the value from the connection string. 

Solution: Rewrite as explained in the `set` method documentation.

Excerpt from the documentation:

```
Values are used if they are non-None.  To set a value to ``None``
        explicitly, use the :meth:`_engine.URL._replace` method adapted
        from ``namedtuple``.
```


ColdHeat:
Good catch thank you!

PR #2375, opened 2023-07-26
Add email whitelist wildcard

MilyMilo:
This allows to specify domains with a wildcard like `*.edu` for the domain whitelist.

Note that this is not a regex match, it doesn't evaluate the expression to support inner wildcards, e.g. it's not possible do do something like `cs.*.edu` - however that's intentional to keep this simple.

### DIFF
--- a/CTFd/utils/email/__init__.py
+++ b/CTFd/utils/email/__init__.py
@@ -139,6 +139,14 @@ def check_email_is_whitelisted(email_address):
     domain_whitelist = get_config("domain_whitelist")
     if domain_whitelist:
         domain_whitelist = [d.strip() for d in domain_whitelist.split(",")]
-        if domain not in domain_whitelist:
-            return False
-    return True
+
+        for allowed_domain in domain_whitelist:
+            if allowed_domain.startswith("*"):
+                # Handle wildcard domain case
+                suffix = allowed_domain[1:]  # Remove the "*" prefix
+                if domain.endswith(suffix):
+                    return True
+            elif domain == allowed_domain:
+                return True
+
+    return False
--- a/CTFd/utils/email/__init__.py
+++ b/CTFd/utils/email/__init__.py
@@ -137,15 +137,21 @@ def user_created_notification(addr, name, password):
 def check_email_is_whitelisted(email_address):
     local_id, _, domain = email_address.partition("@")
     domain_whitelist = get_config("domain_whitelist")
+
     if domain_whitelist:
         domain_whitelist = [d.strip() for d in domain_whitelist.split(",")]
 
         for allowed_domain in domain_whitelist:
-            if allowed_domain.startswith("*"):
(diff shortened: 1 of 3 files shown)


tests/utils/test_email.py
```diff
@@ -239,3 +240,32 @@ def test_successful_registration_email(mock_smtp):
             email_msg
         )
     destroy_ctfd(app)
+
+
+def test_email_whitelist():
+    app = create_ctfd()
+    with app.app_context():
+        set_config("domain_whitelist", "example.com, uni.acme.com, *.edu, *.edu.de")
```
ColdHeat, line 248: Can you add `*.example.com` and update the test to make sure that we test those cases?

CTFd/utils/email/__init__.py
```diff
@@ -139,6 +139,18 @@ def check_email_is_whitelisted(email_address):
     domain_whitelist = get_config("domain_whitelist")
     if domain_whitelist:
         domain_whitelist = [d.strip() for d in domain_whitelist.split(",")]
-        if domain not in domain_whitelist:
-            return False
+
```
ColdHeat, line 142: Originally I thought that this code didn't work for `*.example.com` but I believe it does. The user would just need to remember that `*.example.com` does not also include `example.com` so they would need to add both. 

Also this logic does allow things like `*ample.com` which I think is okay.

ColdHeat (changes requested):
Update the test and I think we are good here

CTFd/utils/email/__init__.py
MilyMilo: It's all in the handling of the suffix - we can either strip out just the asterisk:

`suffix = allowed_domain[1:]` which requires the user to set both `example.com, *.example.com`

or alternatively, also strip the dot `*.`:

`suffix = allowed_domain[2:]` which then allows things like `john.doe@foobarexample.com`

I think from these options I actually prefer the first case.

MilyMilo:
I think this is as good as it's going to get. 
I don't think it's possible to avoid `.example.com` with this design, but I filtered out `*.example.com`.

tests/utils/test_email.py
MilyMilo: I've added some more isolated test cases

CTFd/utils/email/__init__.py
ColdHeat: We should allow `*.example.com` but not something like `*ample.com`. We don't want someone to rely on the `*ample.com` behavior for some reason and then we need to change it. 

So the overall logic here is right. Now my question is what to do about the `* in domain` check.

ColdHeat: I get the sense that the * in domain check is fine as it will prevent sending out invalid confirmation emails since * isn't valid in a domain. I am good with this PR now.

ColdHeat (approved):

PR #2379, opened 2023-07-31
fix: Accept multipart/form-data with token auth

ntoofu:
## Problem

Uploading files by web UI invokes `post_files_list` API with "multipart/form-data" content-type.
But that content-type is not accepted when users are authenticated with token. As a result, scripting such operations is hard.

## Changes

* In validation of token authentication, accept "multipart/form-data" as well as "application/json"
* Use `flask.Request.mimetype` instead of `flask.Request.content_type` in that validation because "multipart/form-data" usually has "boundary" parameter which does not affect to the validation


ColdHeat:
This feels like it introduces something incorrect and even possibly a security issue if it's applied to every endpoint. 

What is wrong with the existing code in ctfcli? 
https://github.com/CTFd/ctfcli/blob/7b4a09af8414eb1f5f6da9a8422fb53b5e9cbc15/ctfcli/utils/challenge.py#L278

ntoofu:
Thank you for your review.

> What is wrong with the existing code in ctfcli?

I tried ctfcli and it succeeded to upload files, but by capturing requests I found cookies are sent to API endpoints. I think ctfcli is authenticated not by tokens but by session key in cookies, when it POST to `/api/v1/files` .

Is that an intended? IMHO, most API is authenticated only by tokens and session cookie is not required.

ntoofu:
BTW, [this code](https://github.com/CTFd/CTFd/blob/6d74c1c4267b4dc395949a65331c369e0737d16f/CTFd/utils/initialization/__init__.py#L293) might be problematic.

In my experiment with curl, accessing to API endpoint with session cookie has failed because it lacked CSRF-Token. But ctfcli seems to be authenticated without CSRF-Token. I think requests with session cookie and API token can bypass CSRF token validation, regardless of the API token's validity.

ColdHeat:
ctfcli never receives a cookie in its configuration. You only specify an API key. Can you provide some evidence that ctfcli is communicating using a cookie?

I don't believe via CSRF you can set an arbitrary header directly. To pull the attack off you would need to leak an API key which no attacker would do since the API key grants all the access they need anyway. 

With XSS you can set the Authorization header but at that point you could just leverage the existing session cookie. Leaking the API key would just give the victim more access. 

I'm not really understanding your attack vectors so you would need to provide a POC.

ntoofu:
> Can you provide some evidence that ctfcli is communicating using a cookie?

Here is the packet capture. I executed `ctf challenge install` command while running tcpdump.

[attachment.zip](https://github.com/CTFd/CTFd/files/12232124/attachment.zip)

> I'm not really understanding your attack vectors so you would need to provide a POC.

I just suggested the possibility and I can't come up with attack vectors. I will report a PoC if I find it.

ColdHeat:
Thank you for the packet capture, it's very useful! I put it up at https://apackets.com/pcaps?pcap=5b6b7989976c6f7a627317f3eafcd619.pcap&view=charts. 

I see that the session is generated which is the source of the issue. I see what you are saying. I will need to investigate this issue and see what the implication is of the change.

ColdHeat:
Thanks for providing that. However I dont know how a CSRF exploit would set the Authorization header?

ntoofu:
Currently, two things prevent an exploit using the bypass of CSRF token validation.

* `SameSite` attribute on session cookie is set to `Lax`
* CTFd does not support CORS

If an attacker provide javascript calling `fetch()` or `XMLHttpRequest()` via malicious site, potentially any request can be crafted and sent from victims. But a request containing Authorization header is not "Simple Request", so most browsers send preflight request and CTFd does not send CORS related headers in response. Additionally, if the preflight check passed in any way, session cookie with SameSite=Lax is not sent and the request fails to be authenticated.

So I think it is not exploitable immediately, but if [the code](https://github.com/CTFd/CTFd/blob/6d74c1c4267b4dc395949a65331c369e0737d16f/CTFd/utils/initialization/__init__.py#L293) is intended to check users are authenticated by API tokens, that intention is not implemented properly.

ColdHeat:
@MilyMilo FYI the changes being discussed here will affect ctfcli

MilyMilo:
I've been working on ctfcli recently and I can confirm this. CTFCLI's API Session object is not configured to reject cookies, and CTFd will issue sessions even to requests originally authenticated with the Token - making ctfcli save them, which led to this going unnoticed as everything was working. Making a `multipart/form-data` request to `/api/v1/files` is in fact rejected without the session.

I think the best temporary solution to this would be to check the path inside the tokens() function and only allow `multipart/form-data` on the `/api/v1/files` route like this:

```python3
@app.before_request
def tokens():
    token = request.headers.get("Authorization")
    is_multipart_files_upload = request.path == "/api/v1/files"
    if token and (request.mimetype == "application/json" or
                  request.mimetype == "multipart/form-data" and is_multipart_files_upload):

    # (...)
```

@ColdHeat

ColdHeat:
I think it's better to:

* Check for request.path == "/api/v1/files". Just need to make sure that this doesn't break on subdirectory deployments.
* Only do this check if we have the authorization token and not before we know if it's a token request since this behavior only matters for token requests.

I think with that change we can accept this.

ntoofu:
@ColdHeat Thanks for the comment.

Is there any concern in the code suggested at https://github.com/CTFd/CTFd/pull/2379#issuecomment-1685934823 ?
I think `request.endpoint` is better in the aspect of sub-directory deployments, because the value of `request.endpoint` is invariable even when CTFd is running in a sub-directory.

ColdHeat:
I think it's a toss up. I am okay with `request.endpoint`.

The API path is very unlikely to change if ever. I am concerned that one day the API naming convention will change but we can let that slide for the sake of this PR.

ntoofu:
@ColdHeat 
I have updated the code. Since `request.endpoint` is consistent even in a case of sub-directory deployment, I adopted it. And I avoided to use the endpoint name (`files_files_list`) which is more likely to change, by looking up `flask.url_map` with rule string such as `/api/v1/files`.

ColdHeat:
@ntoofu thank you for that. I will take over and sheperd this into the codebase.

ntoofu:
Thank you! I tested on current `master` branch (3e6dfc8cb02f7c5d61fcba80d234ec7fb6317223) and comfirmed this issue is fixed.

PR #2396, opened 2023-09-01
Add the possibility to include solution write ups, and to hide them

adb014:
This is an updated version of the pull request #2238 that also includes 

- The possibility to hide solutions individually
- The possibility to bulk edit the visibilty of the solutions
- Updated  for 3.6.0

ColdHeat stated

```
I like the general idea behind this but I'm not sure how complete the implementation has to be. It's simple to just have some text but eventually people will want to add in pictures and such. I will circle back on this.
```

to the previous pull request. Though this code uses the same editor as the challenge description that already includes the possiblity to include images. See 

image

For an example (sorry in french) of what I already have. So I'm not sure I understand the issue. Please note that this pull request modifies the javascript, but doesn't include the changes to the "*.dev.js" and "*.min.js" files. So an `npm run build` will be needed to actually use this code. I felt it was better to not include the webpacked files to make it easier to read.

### DIFF
--- a/CTFd/admin/challenges.py
+++ b/CTFd/admin/challenges.py
@@ -7,7 +7,7 @@ from CTFd.schemas.tags import TagSchema
 from CTFd.utils.decorators import admins_only
 from CTFd.utils.security.signing import serialize
 from CTFd.utils.user import get_current_team, get_current_user
-
+from CTFd.utils.config.visibility import solutions_visible
 
 @admin.route("/admin/challenges")
 @admins_only
@@ -15,7 +15,8 @@ def challenges_listing():
     q = request.args.get("q")
     field = request.args.get("field")
     filters = []
-
+    solution_visible = solutions_visible()
+    
     if q:
         # The field exists as an exposed column
         if Challenges.__mapper__.has_property(field):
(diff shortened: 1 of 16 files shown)


ColdHeat:
Hi there, I am not against this idea but it will take time to find the time to scope out how this should look and merge it in.

adb014:
I've rewritten this submission as a plugin available at https://github.com/adb014/CTFd-solutions

ColdHeat:
I am unsure if I can continue on the underlying base of this PR. It accomplishes the goal but I think more work is needed to achieve what Im looking for. 

* Writeups should be a new top level api endpoint `/api/v1/writeups` essentially. It could be called solutions. 
* Writeups should not be a new column within challenges but it should instead be a seperate table.
* Writeup creation should be a tab on update. Similar to hints or topics. I think the top level Challenges is already very busy so we shouldn't add to it. 
* I see now that solution_state is for individual control of a writeup for a challenge. I think this is a reasonable idea. I wonder if writeup visibility should be stored on the challenge or on the writeup.

adb014:
Funny you say that ´cause in the end I converted this PR to a plugin that does indeed store the writeups in a seperate table as mentioned above.

ColdHeat:
This has been implemented via #2778. Thank you for the initial prototype and plugin.

PR #2400, opened 2023-09-16
Fix of html markup of team private template

dbucki:
### Issue
Not valid HTML and token for localization

I wanted to improve translations and just after run of `make translations-extract`, and then `make translations-update` I noticed that there is one kind of typo inside html markup of `private template` for team.

This PR is about hotfix of that issue.

Without that we have not valid HTML markup and strange localization token like: 
image

### DIFF
--- a/CTFd/themes/core-beta/templates/teams/private.html
+++ b/CTFd/themes/core-beta/templates/teams/private.html
@@ -203,7 +203,7 @@
 
         </div>
         <div class="modal-footer">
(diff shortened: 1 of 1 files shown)


ColdHeat:
Hi again! Good catch!

Can you actually put this PR against https://github.com/CTFd/core-beta? 

Currently we are keeping the core-beta theme seperate and merging changes manually into CTFd.

dbucki:
@ColdHeat Sure, PR has been created here -> https://github.com/CTFd/core-beta/pull/53

I didn't notice that you have separate repo for that theme 👀

PR #2401, opened 2023-09-21
Resolve email configuration issues

Konstantinos-Papanagnou:
This branch will fix some issues with the email configuration + will add some verbose output when something fails with the configuration.

### DIFF
--- a/CTFd/auth.py
+++ b/CTFd/auth.py
@@ -77,15 +77,19 @@ def confirm(data=None):
     if data is None:
         if request.method == "POST":
             # User wants to resend their confirmation email
-            email.verify_email_address(user.email)
-            log(
-                "registrations",
-                format="[{date}] {ip} - {name} initiated a confirmation email resend",
-                name=user.name,
-            )
-            return render_template(
-                "confirm.html", infos=[f"Confirmation email sent to {user.email}!"]
-            )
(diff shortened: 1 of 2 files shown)


ColdHeat:
Hi you don't need to comment out the code, you can just remove it. We will see the changes in git history. 

However can you cite to me how you are determining the correct means of "configuring TLS communications for SMTP"?

Konstantinos-Papanagnou:
Hello,

Ah right, forgot to remove the code from my tests. I'll issue another commit removing those on this branch.

Basically what the issue is here is that when using the SMTP_SSL function the operation fails with an os.SSL error every time, as incorrect SSL cipher. The fix I found was to use the plain SMTP function while retaining the starttls command later. 

I was using Microsoft's O365 smtp endpoint on port 587 which actually requires TLS and StartTLS enabled. 

The patch I've issued here fixes it (which was trial and error)

Konstantinos-Papanagnou:
Had another look today, Microsoft says SSL and TLS are required while configuring but it's actually not based on this:
https://stackoverflow.com/questions/69541840/ssl-error-while-sending-email-using-smtp-with-python

So if we disable SSL it works like before. No need to amend any changes in the smtp.py function as it may break ctfd for other people. I reverted that change. However, feel free to amend the verbose update on what's wrong to auth.py file.

PR #2405, opened 2023-09-29
Fix bug that overwrite the configs in "extra" section

ktpss95112:
I forked CTFd and customized it to my need and I added some extra configurations both in `config.ini` and in `config.py`. The reason I needed to add it in `ServerConfig` is that I needed to do some checking and rewriting on the value of the config, but I found that my rewriting does not take place. It turns out that my rewriting is overwritten by the code after the instantiation of `ServerConfig`, since all the config in "extra" section is added to `Config` there. I think it might not be considered a bug to some people, but it would be nice if we add some hints and comments to make it clearer.


ColdHeat:
I debated this one in my head for awhile and I think it is okay. 

I think if you had not used the extra section to store keys that you specified in config.py it would have worked. 

I think your change specifically enforces that the extra field should not be used to set keys that are already defined in config.py. 

If it is already defined, do not overwrite. if it doesn't exist it can be added. 

Adding existing keys into the extra field will not work. For example if you put SECRET_KEY it will not work. However do we want this to work? I would argue no so I am okay with this.

PR #2420, opened 2023-11-03
Add unit tests for plugins

pve:
Situation: there are many unit test cases, but none for plugins.
Solution in this pull request:
Describe the workflow, tests can live with the plugins.
Show how to let ChatGPT help
Sample tests for flags unit.

Added a dev container description for more convenient developer workflow.


ColdHeat:
Hi there, why does this PR have a number of files unrelated to running test cases? I don't think these Dockerfile changes are necessary. I think `devcontainer.json` is probably better in a seperate PR as well. 

I'm also not a big fan of encouraging ChatGPT code in CTFd. I think it's okay for people to use for themselves but not something we should explicitly endorse for writing test cases. 

I would expect this PR to contain some added plugin tests (which you've already done) as well as whatever changes are necessary to our Makefile to make plugin test running work.

pve:
thanks, appreciate the comments. We will update this.

pve:
Reverted the less relevant changes.

pve:
looks like I managed to do that.

pve:
not sure what you exactly mean. Do you have an example in one of the main tests, so I can copy/adapt??

ColdHeat:
This is just because of one of the test steps. See https://github.com/CTFd/CTFd/actions/runs/7077554078/job/19266261539?pr=2420#step:5:222. 

You want to add `# nosec` to your assert lines as they discuss here: https://bandit.readthedocs.io/en/latest/config.html#exclusions.

ColdHeat:
The `# nosec` is specific to your change because these tests live outside of the tests folder. Please update your PR so that it passes the lints properly and I will merge.

pve:
Should be done now., though I have trouble reproducing your exact toolchain and pipeline. Is there a doc somewhere for onboarding developers? If not, shall I have my next student team work on that?

Issue #910, opened 2019-03-20
Taiwai is part of Chine 


hubu-wangpei:
Taiwai is part of Chine ,Can you make it right?

ColdHeat:
I did not personally generate the list of countries used in CTFd. It comes from other online sources. 

https://en.wikipedia.org/wiki/ISO_3166-1
https://github.com/umpirsky/country-list/blob/master/data/en_US/country.csv#L223

Also related: https://github.com/umpirsky/country-list/issues/89

PR #2424, opened 2023-11-20
Fix some issues that may cause trouble for Chinese users

f00001111:
Fix some issues that may cause trouble for Chinese users.
Taiwan is a part of China.
Taiwan is a province in China. So it can't be displayed side by side with China.

### DIFF
--- a/CTFd/constants/languages.py
+++ b/CTFd/constants/languages.py
@@ -7,8 +7,8 @@ class Languages(str, RawEnum):
     POLISH = "pl"
     SPANISH = "es"
     ARABIC = "ar"
-    CHINESE = "zh_CN"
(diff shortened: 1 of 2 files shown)


ColdHeat:
The change of language title seems within reason for clarity but the change of the country name I likely will not accept. See the rationale at https://github.com/CTFd/CTFd/issues/910. Should that dataset change we can regenerate our list here. 

I think the right solution here is to copy what Wikipedia does and use the characters listed here: https://en.wikipedia.org/wiki/Chinese_Wikipedia#Automatic_conversion_between_traditional_and_simplified_Chinese_characters.

f00001111:
Sorry. I didn't see that issue before.

PR #2426, opened 2023-11-24
add prefix option to S3 uploader

raihalea:
This proposed modification enables the addition of a custom prefix when saving files to AWS S3, which can enhance integration with CloudFront through path pattern matching. If implemented, this approach would allow CloudFront to effectively cache content by using S3 as its origin.

The AWS_S3_CUSTOM_PREFIX is dynamically added during file uploads and removed during downloads, preserving compatibility with existing setups.

For CloudFront Origin settings, the following configurations are expected:
- AWS_S3_CUSTOM_DOMAIN = "CloudFront Domain"
- AWS_S3_CUSTOM_PREFIX = "s3/sub/" (as an example)

Implementation Guide:

In the CloudFront distribution, configure the S3 bucket to use a custom prefix like  `s3/sub/` (this is just an example, and can be adjusted as needed).
Set up behaviors in CloudFront for the path pattern `s3/sub/*` (or your custom path).

- Before:
  - `s3://<bucket name>/<md5 hash>/image.png`
- After:
  - `s3://<bucket name>/<AWS_S3_CUSTOM_PREFIX>/<md5 hash>/image.png` 

By implementing these settings, accessing a URL like `https://<AWS_S3_CUSTOM_DOMAIN>/s3/sub/<md5 hash>/image.png` (or with your custom path) will redirect to the URL issued by CTFd, allowing CloudFront to efficiently serve the S3 file.


ColdHeat:
Thank you for filing this. I like the idea but I would like to spend some time testing this out.

ColdHeat:
Overall I see this change as allowing CTFd to stores files within a folder within a bucket instead of having to use the bucket top level. This might help Cloudfront but it doesn't seem like the main purpose.

ColdHeat:
This appears to be missing the implementation for deletion?

PR #2427, opened 2023-11-25
Support for optional and configurable OAuth

cvandeplas:
CTFd already supports OAuth in the backend.
The only thing that is missing are a few configurable variables to allow full integration with an OAuth server such as KeyCloak.

This commit allows the admin to configure an external OAuth server in `CTFd/config.ini`,  while not breaking existing Major League Cyber preferences of the project.

image


ColdHeat:
Hello, thank you for submitting this. This concept will be considered for a 4.0 release.

cvandeplas:
Hi there, I was honestly hoping to see it merged in the current version :-) 
It's a fairly small change  and does not change the default behaviour of CTFd in regards to authentication.

Anyway, great to know you're considering more OAuth for v4.0 !

README.md
```diff
@@ -83,6 +83,19 @@ OAUTH_CLIENT_ID = None
...
+To integrate with any other OAuth provider, such as KeyCloak, edit `CTFd/config.ini` and set:
+```python
+OAUTH_NAME = MyAuth
+OAUTH_CLIENT_ID = None
+OAUTH_CLIENT_SECRET = None
+OAUTH_TOKEN_ENDPOINT = https://oauth.local/realms/<yourrealm>/protocol/openid-connect/token
+OAUTH_API_ENDPOINT = https://oauth.local/realms/<yourrealm>/protocol/openid-connect/userinfo
+OAUTH_API_ENDPOINT = https://oauth.local/realms/<yourrealm>/protocol/openid-connect/userinfo
+```
+You will have to add a mapper for `id` in your client scope, and ensure `openid` is added as client scope.
```
BrandonIngalls, line 97: Any thoughts on how to handle an oauth provider where you do not have a readily available `id` value that is an integer?

cvandeplas: Excellent question @BrandonIngalls .
Looking at the code the this `id` is mapped to a `Users.oauth_id` and potentially to a `Teams.oauth_id`.. The requirement of it being an integer is located in the `Users` and `Teams` classes; that also triggers the DB structure integer requirement. 

Looking at the code I see no place where this `oauth_id` must be an ID. Actually it looks it's mostly used as a boolean like `{% if team.oauth_id %}`.

Summarized, unless I misunderstand the code, I think it could even be dropped from the database and transformed as a local boolean to track if the session is oauth-based or not. Thus dropping the requirement for getting an `id` from the oauth provider.

jonakarl:
Thanks for the PR, based on this I made a version that works with gitlab (tested with on premise installation of gitlab, I can make the code public after  the holidays with that cavate that it works for me but it is in now way finished).  

However,  at least gitlab and maybe other providers need some more modifications. In the case of gitlab we need to supply a redirect_uri and a different scope (read_user or possibly openid).
Gitlab work (with the above modifications)  OOB as a provider for users but will not work as a provider for teams without some hackish workarounds (I therefore disabled that code path).
 
While I am quite new to ctfd (started looking in to it just recently) and I am sure there is a good reason for they way it is implemented, I feel ctf/mlc teams maps a little bad to how I would consider it to work for a general identity provider. For example naturally I would use the same "scope" for both users and teams but if "teams" are configured I would also read the users group membership and "form" teams based on that probably with some whitelist.

PR #2430, opened 2023-11-28
Add ability to generate social share links

ColdHeat:
* Add Pillow to dependencies
* Add ability for CTFd to generate social share links
* Add `/api/v1/shares`
* Add /shares/[type]
* Add /shares/[type]/assets/[path]

WIP items:
* Ability to change banner image in Admin Panel (may require the implementation of the named files idea)
* ~~Ability to set the meta tags in base.html for core-beta~~
* ~~Buttons for social share links~~
* ~~Downloading logo image from S3 to generate the share image~~ (needs testing)
* ~~Disabling share feature entirely~~

SKIP:
* Controlling the Share template.
  * This is being skipped for now because of the possibility that an admin accidentally introduces an XSS vulnerability into their instance
* Make initial banner image set to `ctf_banner` config 
  * It's tough to implement this because if a user updates the banner it will not update the image on the front page

Example:

image

http://localhost:4000/share/assets/solves/1/2cdf0ccce88d12456eddbbf4654cdf305fbce63a.png

http://localhost:4000/share/solves/1/2cdf0ccce88d12456eddbbf4654cdf305fbce63a


ColdHeat:
Some notes on the privacy implications of social shares:

The knowledge that User X solved Challenge Y can be a sensitive piece of information.
If a user extracted all social shares they could reconstruct the scoreboard which in some cases is a private piece of information.
However if User X chooses to share that they solved Challenge Y, this is an okay action. Because sharing your own position does not actually help you determine your scoreboard position. User Z may be higher but hasn't shared their information. 

Therefore a user must be in control of choosing what they share, meaning that an arbitrary user cannot determine that User X has solved Y without User X directly revealing that information. 

Thus we need a way for shares to not be bruteforceable.

ColdHeat:
I think it is worth noting that the social share feature is only really beneficial for public events and it's not beneficial for private events. It could even be detrimental for private events if users are in some kind of sensitive environment and accidentally share some private information. 

I think this feature should be an opt-in feature that is disabled by default.

PR #2435, opened 2023-12-08
fix: update gevent dependency

RiRa12621:
fixes https://github.com/gevent/gevent/issues/1989 which is rated on a solid 9.8


ColdHeat:
This did show up on my radar but there shouldn't be any impact as we use gevent as part of gunicorn which is then fronted by nginx. But this should be safe to merge anyway.

PR #2436, opened 2023-12-08
fix: update EOL base image

RiRa12621:
buster is EOL and doesn't receive security fixes. Current is bookworm. This causes security scanners to fire up for docker base images with issues that should be fixed in the new bases:

Screenshot 2023-12-08 at 08 57 18


RiRa12621:
@ColdHeat same as the other one: probably no impact, but rather safe than sorry and stops image scanners to light up like a Christmas tree

ColdHeat:
I am merging this into the dev branch for 3.7.0 just in case there are some edge cases where things break. Thanks again for the PR!

RiRa12621:
👍 feel free to ping me if something breaks. Happy to follow up then.

PR #2447, opened 2023-12-21
Fix Display Error When Website Language is Set to Traditional Chinese (zh_TW)

rex978956:
This fix addresses an issue with Flask-Babel's handling of `zh_CN` and `zh_TW`.
According to the discussion in [Ckan Issue #4413](https://github.com/ckan/ckan/issues/4413)
Flask-Babel fails to correctly identify `zh_CN` or `zh_TW` because it looks for the directories named `zh_Hant_CN` and `zh_Hant_TW`.

Therefore, I have renamed the original `zh_TW` and `zh_CN` directories to `zh_Hant_TW` and `zh_Hant_CN`, respectively, to comply with Flask-Babel's recognition standards.

Additionally, regarding the language setting options, I suggest that instead of just using '中文' or '國語', we should clearly differentiate between '简体中文' (Simplified Chinese, zh_CN) and '繁體中文' (Traditional Chinese, zh_TW). This clarification will help users better understand the language options they are selecting.


ColdHeat:
Thank you for filing this, I am okay with this. I will need to see how this will work with POEditor before merging though. We may be moving away from it anyway.

ColdHeat:
As part of this I am moving translations over to Crowdin which gives us the ability to add custom language codes.

PR #2449, opened 2024-01-02
3.7.0

ColdHeat:
* Update base image to `python:3.11-slim-bookworm`
* Raise exception if a built-in config is defined in the extra config section
* Migrate Admin Panel from webpack to Vite
* Adds Alpine to Admin Panel
* Add ability for users to generate social share links 
* Update dataset to 1.6.2
* Implement scoreboard brackets feature
* Add `uploaders.open()` to open a file from an uploader
* Add ability to control the link target for a page (i.e. open in a new tab)
* Add exmaple of unit tests for plugins


CTFd/utils/uploads/uploaders.py
```diff
@@ -19,21 +19,47 @@
...
+        """
         raise NotImplementedError
 
     def sync(self):
+        """
+        Download all remotely hosted files for the purpose of exporting
+        """
+        raise NotImplementedError
+
+    def get(self, mode="rb"):
```
ColdHeat, line 57: This should be called `open()` no?

CTFd/config.py
```diff
@@ -269,4 +273,8 @@ class TestingConfig(ServerConfig):
 # Actually initialize ServerConfig to allow us to add more attributes on
 Config = ServerConfig()
 for k, v in config_ini.items("extra"):
+    # We should only add the values that are not yet loaded in ServerConfig.
+    if hasattr(Config, k):
+        continue
```
ColdHeat, line 278: I wonder if perhaps this should raise an Exception instead of silently allowing it.

PR #2451, opened 2024-01-07
File upload improvements

### LINKED ISSUE #1595: Upload Path Randomness
Is there a specific security or practical reason that the component before the file name of an upload path is random? The variable names in `uploaders.py` seem to indicate that at some point this was a hash sum of the file (`md5hash`).

I ask because I'm trying to implement an automated CTF management tool which can synchronize a CTFd instance with a local configuration (defined in YAML files). The challenge definitions define file names/paths to upload, but there's no way for me to quickly check whether a file has changed, which makes it difficult to implement the sync logic.

Currently my only options are:

1. Assume the file is dirty and always delete/re-upload
2. Assume the file is clean if it exists, and require a separate option to force re-uploading.

In case one, this makes synchronizing the entire platform always take a long time (due to deleting/re-uploading every challenge attachment). Option two requires the user to remember if/when files are changed. In a distributed development environment, it would be great to allow developers to push changes to the CTFd configuration and an admin to simply synchronize the configuration.

I'm open to running a local fork of CTFd for my use-case, but I wanted to make sure there wasn't a serious reason it was implemented in this way before I just decided to change it. Thanks!

ColdHeat:
* Calculate a files sha1sum on upload for future local change detection purposes
* Allow clients to control the location of an uploaded file
* Adds the optional path field to the Uploaders class to control where files get uploaded to
* Closes #1595

### DIFF
--- a/CTFd/api/v1/files.py
+++ b/CTFd/api/v1/files.py
@@ -90,10 +90,25 @@ class FilesList(Resource):
         # challenge_id
         # page_id
 
+        # Handle situation where users attempt to upload multiple files with a single location
+        if len(files) > 1 and request.form.get("location"):
+            return {
+                "success": False,
+                "errors": {
+                    "location": ["Location cannot be specified with multiple files"]
+                },
+            }, 400
+
         objs = []
         for f in files:
             # uploads.upload_file(file=f, chalid=req.get('challenge'))
-            obj = uploads.upload_file(file=f, **request.form.to_dict())
+            try:
+                obj = uploads.upload_file(file=f, **request.form.to_dict())
+            except ValueError as e:
+                return {
+                    "success": False,
+                    "errors": {"location": [str(e)]},
+                }, 400
             objs.append(obj)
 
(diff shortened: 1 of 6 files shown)


ColdHeat:
Reverted the unique constraint b/c uniqueness on a text column (or long varchar) in MySQL can only guarantee uniqueness of a certain length. This length seems to be either 191 chars or 767 chars. I did not validate the limit but in either situation given the existing column type it's not enough. 

I would be okay with limiting to 767 chars but we would need to make sure any existing paths don't go over that limit.

I think the true solution is to store location as a _location field as well as a _hash field and have _hash be a unique hash of the _location string. That involves more work and I don't think it's particularly necessary here. Even if there's a duplicate path CTFd will just get the first result when querying the DB.

ColdHeat:
@MilyMilo We can use the changes here to help ctfcli to determine if a file upload needs to be done or not. Also with uploading files that we want to have a guaranteed name.

PR #2462, opened 2024-01-29
Using same base image for both build and release stages

khiemdoan:


ColdHeat:
Good catch, thank you!

PR #2487, opened 2024-03-06
use pycountry to replace self-generated country list

mudongliang:


mudongliang:
This is a simple comparison script - https://gist.github.com/mudongliang/e4dbd960678502ce981dfd8f79e19e67

There are many differences between self-generated country list and pycountry iso standard database.

mudongliang:
BTW, I don't quite understand this error message. Shall I migrate this PR to other branches?

ColdHeat:
The most important one is just the linting errors. You just need to run `isort` and `black`. See the lint command in the Makefile. 

But I can clean up the PR once it's decided this is the approach to take. 

If you look at the output from your script many of the discrepancies are minor (just an ampersand). In other situations, the difference is a matter of verbosity. For example `"Congo - Kinshasa" is different from "Congo, The Democratic Republic of the"`. I think the first one is better than the second. 

Is there a cleaner list in pycountry that isn't so verbose?

mudongliang:
https://github.com/pycountry/pycountry/blob/main/src/pycountry/databases/iso3166-1.json

This is the country list json file in pycountry.

mudongliang:
> The most important one is just the linting errors. You just need to run `isort` and `black`. See the lint command in the Makefile.

I just check this with pylint in my local workspace. Will check this later

PR #2488, opened 2024-03-06
Fix json flag hint import in Challenge CSV

ColdHeat:
* Fix issues where types which we did not want to accept as JSON were being processed as JSON in Challenge CSVs


ColdHeat:
@bkimminich this change now forces all JSON flags/hints to be a list of dictionaries. Unsure if you ever got the chance to implement this into the juice shop ctf cli tool but you may need to update it if so.

PR #2491, opened 2024-03-09
Assets.js: add optional defer option

erdnaxe:
When creating custom templates, it can be useful to include some JavaScript code without `defer`. For example, in the case of a Bootstrap template this is needed to load a dark mode selector before showing the page (to prevent white/dark flickering).

Currently the only solution to remove defer is:
```diff
-  {{ Assets.js("assets/js/color_mode_switcher.js") }}
+  <script src="{{ Assets.file("assets/js/color_mode_switcher.js") }}"></script>
```
This pull request proposes to add an optional parameter "defer" to `Assets.js` to be able to write:
```diff
-  {{ Assets.js("assets/js/color_mode_switcher.js") }}
+  {{ Assets.js("assets/js/color_mode_switcher.js", defer=False) }}
```

### DIFF
--- a/CTFd/constants/assets.py
+++ b/CTFd/constants/assets.py
@@ -16,20 +16,21 @@ class _AssetsWrapper:
         )
         return get_asset_json(path=manifest)
 
-    def js(self, asset_key, theme=None):
+    def js(self, asset_key, theme=None, defer=True):
         if theme is None:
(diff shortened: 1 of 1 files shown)


ColdHeat:
I like this change, thanks for the PR!

Issue #2493, opened 2024-03-14
Improve scoreboard API endpoints for scoreboard matrix and top per challenges category


erdnaxe:
**Current situation**:

In current CTFd API, there are 2 endpoints for scoreboards:

- `/api/v1/scoreboard/`: return the whole standings, with team members. In core-beta theme, this is used to generate the scoreboard table.
- `/api/v1/scoreboard/top/10`: return the detail of the top 10 standings (each challenge with solve date). In core-beta theme, this is used to generate the scoreboard table.

These API endpoints seems to have been designed with `core-beta` theme in mind, and are not really generic.

**What I am trying to achieve**:
I am trying to add a scoreboard matrix page, and a top 3 scoreboard per category (crypto, reverse...).

Here is a render of such matrix (done with Jinja2 templates, and a lot of dirty patching of CTFd code):
![image](https://github.com/CTFd/CTFd/assets/2663216/b79735c1-67f3-44b5-8229-c463013b7a9d)

Here is a render of the top 3 page:
![image](https://github.com/CTFd/CTFd/assets/2663216/06274f0f-eee1-49aa-8a09-d88e7efff415)

**Limitations observed in current API**:
To generate the matrix, I could use `/api/v1/scoreboard/top/100`, but there is no pagination support in the API, and I need some details that only `/api/v1/scoreboard/` provides.
First/second/third blood information is also missing from API.

To generate the top 3 per challenge, I would need `/api/v1/scoreboard/top/3`, but with an extra parameter to specify a challenge category.

**API changes proposal**:

- Add a `category` optional parameter to `ScoreboardDetail`, to easily get the top players/teams in one category.
- Add a `bracket_id` optional parameter to `ScoreboardDetail`, to easily get the top players/teams in one bracket: https://github.com/CTFd/CTFd/pull/2495
- Make the `count` parameter in `ScoreboardDetail` optional, or add pagination in the API (`count` + `offset` parameters ?).
- Make `ScoreboardDetail` returns information `ScoreboardList` is already returning: https://github.com/CTFd/CTFd/pull/2494
- Add first/2nd/3rd blood information in the `solves_mapper` in `ScoreboardDetail`. This could be done with a query to get  `account_id` and `challenge_id` from `Solves`, ordered by `Solves.date.asc()`, and limited to 3 per `challenge_id`.

**Environment**:

- CTFd Version/Commit: fde98ce6f7e3e0a5f0a750374e8263d8f7f4054a
[...]

PR #2494, opened 2024-03-15
Return more fields in ScoreboardDetail

erdnaxe:
Linked to issue https://github.com/CTFd/CTFd/issues/2493

When using CTFd API, I would expect `ScoreboardDetail` to return more information than `ScoreboardList`. JavaScript codes need to query twice the API then merge information.

This patch proposes to expose more fields for each standing in `ScoreboardDetail`.

*Note: renaming `id` to `account_id` would be nice, but it would break backward compatibility, so I'm not proposing that.*

### DIFF
--- a/CTFd/api/v1/scoreboard.py
+++ b/CTFd/api/v1/scoreboard.py
@@ -142,10 +142,14 @@ class ScoreboardDetail(Resource):
                 solves_mapper[team_id], key=lambda k: k["date"]
             )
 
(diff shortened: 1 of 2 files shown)


ColdHeat:
Thanks for catching this. I think this endpoint just got missed during Brackets implementation.

ColdHeat:
You could possibly consider adding the account_id field instead of replacing the id field. Perhaps in a seperate PR.

PR #2495, opened 2024-03-15
Add bracker filter in ScoreboardDetail

erdnaxe:
Linked to issue https://github.com/CTFd/CTFd/issues/2493

This patch proposes to add an extra GET parameter in ScoreboardDetail to filter by bracket.
This is useful to create custom scoreboards in downstream CTFd themes.

**I need help with `@cache.cached(timeout=60, key_prefix=make_cache_key, query_string=True)`.** It seems that CTFd custom cache keys does not support `query_string` option.

### DIFF
--- a/CTFd/api/v1/scoreboard.py
+++ b/CTFd/api/v1/scoreboard.py
@@ -4,7 +4,7 @@ from flask import request
 from flask_restx import Namespace, Resource
 from sqlalchemy import select
 
-from CTFd.cache import cache, make_cache_key
+from CTFd.cache import cache, make_cache_key, make_cache_key_with_query_string
 from CTFd.models import Awards, Brackets, Solves, Users, db
 from CTFd.utils import get_config
 from CTFd.utils.dates import isoformat, unix_time_to_utc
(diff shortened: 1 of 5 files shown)


ColdHeat:
Yes I believe query params don't work for some reason. I think the function was mostly copied out of Flask-Caching though. I will take a look.

ColdHeat:
@erdnaxe I have a rough strategy for query string caching. I have made it so that we only cache based on specific parameters. How does this work for you?

This was tricker than I expected but likely this will be beneficial for other uses as well as it shows how to cache based on a specific item and then clear it as well meaning we can cache individual pages more easily.

ColdHeat:
FYI it looks like Flask-Caching does not allow controlling the query string logic if you specify the query_string argument which is why I think it's better to side step the behavior entirely.

https://github.com/pallets-eco/flask-caching/blob/a367b9f13109d35c1e3df069f5da4c29110a689c/src/flask_caching/__init__.py#L473-L477

ColdHeat:
Seperately I have no idea why this endpoint returns a dictionary instead of a list. Might be an old holdover from when Flask could not return a list at the top level of a JSON response. 

Could be worth changing in 4.0.

PR #2507, opened 2024-03-27
support complex data structure for submission

ericdum:
Since I try to add a plugin that needs more than one submission data I need to submit to /attempt by adding a JSON object to the submission parament. 

In this situation, everything works as expected except the logging part.

The previous code attends to log submission data as a string which not good for JSON object. So I change the code to determine if it is not a string then stringify it as a JSON object.

### DIFF
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -506,7 +506,7 @@ class ChallengeAttempt(Resource):
 
         challenge_id = request_data.get("challenge_id")
         submission = request_data.get("submission", "")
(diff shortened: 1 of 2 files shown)


ColdHeat:
I believe in your plugin you can just treat the entered data as JSON and unserialize it. I don't think this PR is necessary.

ColdHeat:
See https://github.com/CTFd/CTFd/blob/c918b0ada8aeeb3c9821600698ec2ea744b0eb42/CTFd/plugins/challenges/__init__.py#L110-L129

Issue #1389, opened 2020-05-07
Use AWS S3 credentials provided by AWS Resources (Instance Role, Task Role, IRSA)


hugoprudente:
The S3Uploader class today requires the explicitly credentials not working with the temporary provided by AWS as we can confirm on the below.

```python
    def _get_s3_connection(self):
        access_key = get_app_config("AWS_ACCESS_KEY_ID")
        secret_key = get_app_config("AWS_SECRET_ACCESS_KEY")
        endpoint = get_app_config("AWS_S3_ENDPOINT_URL")
        client = boto3.client(
            "s3",
            aws_access_key_id=access_key,
            aws_secret_access_key=secret_key,
            endpoint_url=endpoint,
        )
       return client
```

As this is set in the code directly the AWS credentials evaluation provided by the [Boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html) are never used.

It's recommended by AWS security team  to use the AWS temporary provided credentials when using AWS solutions.

As the Environment Variables are already supported by boto3 if you sent `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` it will already read it and use all the evaluation.

For the `AWS_S3_ENDPOINT_URL` it can be replaced by the `AWS_REGION` environment variable that will generate the endpoint automatically for you.  

```python
    def _get_s3_connection(self):
        client = boto3.client("s3")
       return client
```

I would like to know if you have any thoughts on this before I submit a PR so we can drill down all use cases and cover all ground to let it flexible and secure as AWS recommends.

ColdHeat:
Hmm someone has probably run into this before so I suppose this is worth fixing. I don't think removing the direct configuration values makes sense though because we can't deprecate config values whilly nilly. 

Also I dont think `AWS_REGION` is as flexible as `AWS_S3_ENDPOINT_URL` b/c `AWS_S3_ENDPOINT_URL` lets you point to any S3 implementation URL. Correct me if I'm wrong though.

hugoprudente:
Hi ColdHeat, 

I think we can work on the following way.

If the `AWS_ACCESS_KEY_ID` is set on the `config.py` is used, but if not try to use `boto3` evaluation system. 

Same for the `AWS_S3_ENDPOINT_URL` if you don't use it `AWS_REGION` will be the one setting the S3 endpoint destination, and if you need an specific endpoint you manually config on `config.py` 

[...]

PR #2508, opened 2024-03-27
Enhance S3 Connection with Environment-Based session_token

raihalea:
Recently, I've been addressing the following issue: https://github.com/CTFd/CTFd/issues/1389.

Modified _get_s3_connection() to directly fetch session_token from environment variables. 
 `session_token` values often change, and it seemed impractical to retrieve them through `config.ini` or `config.py`.

I added a test called `test_s3_uploader_sts()`. 
However, I encountered difficulties simulating how CTFd handles environment variables during test execution, rendering this test potentially meaningless. To avoid confusion, I am considering removing it.

After running `make lint` and `make format`, I noticed a significant increase in the number of files altered. I opted to leave these changes as they are due to uncertainty about the best course of action regarding these modifications.


ColdHeat:
Regarding the linter/formatter errors are you using the same versions that are specified here? https://github.com/CTFd/CTFd/blob/master/linting.txt

raihalea:
I realized I had indeed used the incorrect version. 
After reverting those changes and running make format, I have now pushed the corrected updates once again.

ColdHeat:
I'm not particularly familiar with STS tokens, how do they differ from the typical access key and secret key flow? If STS is being changed as an envvar, CTFd/the container would still need to be restarted so I'm unsure how this is better than changing the access/secret keys?

raihalea:
In response to your comment, I had some concerns about the handling of environment variables, so I've slightly altered our approach.

raihalea:
STS tokens, usually tied to IAM roles, can be auto-assigned by AWS services.
if access_key or secret_key were set, there wasn't a way to incorporate a session_token. I believe boto3 should manage these credentials.

Honestly, I'm not fully convinced we need this change, though it could offer some advantages. 
I've encountered this issue in specific situations, but it hasn't affected EC2, as it uses IMDS instead of environment variables for STS tokens.

PR #2509, opened 2024-03-27
Allow plugins to add an entry in the config page

miyoyo:
CTFd Plugins feel disconnected from the rest of the program, and requiring a separate page for configuration is something that separates them from how everything else does things.
Giving them an option to appear in the `/admin/config` page makes them just a little bit more integrated.

Here are the changes:
- Added a new supported field, `config`, in the config.json file. this field points to a template that will be rendered when the config page is shown.
- Added a new function to jinja, `get_menubar_plugins`, which filters out plugins that don't provide a `route`, that way, plugins can have a config file, and not show up in there. (This is backwards compatible, as the existing `get_configurable_plugins` only added one field to the namedtuple)
- Add a new tab and page to the config page for every configurable plugin that defines a `config` field.

This change should be perfectly backwards compatible.

Changes to the documentation should follow, documenting
- that `config` exists
- that `config` should point to a template, likely `plugins/plugin_name/templates/config.html`
- an example of how to update the config, for example, by having a form with a custom action, or using JS


miyoyo:
Not _quite_ sure how or if to add a test for that function.

ColdHeat:
I will review this shortly when I am able to. I like the ideas you have outlined in the PR.

CTFd/themes/admin/templates/config.html
```diff
@@ -105,6 +116,16 @@ <h1>Configuration</h1>
 				{% include "admin/configs/backup.html" %}
 
 				{% include "admin/configs/usermode.html" %}
+
+				{% if plugins %}
+					{% for plugin in plugins %}
+						{% if plugin.config %}
+							<div role="tabpanel" class="tab-pane config-section" id="{{plugin.name  | replace(' ', '')}}">
+								{% include plugin.config %}
```
ColdHeat, line 124: This flow implies that the plugins themselves need to register a template to the jinja loader. How are you proposing that this should be done? 

You can do it as part of a blueprint registration but is it intended that the plugin developer needs to register this template?

Also there's only one namespace so I believe you can't have multiple config files with the same name (i.e. `config_plugin.html` instead of just `config.html`)

miyoyo: Plug-in templates are always available as `plugins/plugin_name/path_to_plugin.html`

So a template in the plug-in folder for `mysuperplugin` named `mytemplate.html` would be `plugins/mysuperplugin/mytemplate.html`

ColdHeat: Ah so it is, I forgot about that loader. Okay that was my major concern with this PR.

PR #2515, opened 2024-04-04
Add decay function to `DynamicValueChallenge.read`

### LINKED ISSUE #2514: Getting dynamic challenges by ID does not return decay function
When getting a dynamic challenge from `GET /api/v1/challenges/<challenge-id>`, the challenge does not return its decay function.

This seems to be caused by [this](https://github.com/CTFd/CTFd/blob/master/CTFd/plugins/dynamic_challenges/__init__.py#L60-L90) function which does not include `challenge.function`.

This seems be have been missed out when implementing different decay functions for dynamic challenges. This does not have any impact on normal usage of CTFd, but I see no reason not to include the function in the response.

Jus-Codin:
Fixes #2514


ColdHeat:
I think this is fine, thanks!

PR #2519, opened 2024-04-09
French translation fixes

erdnaxe:
"Trophés" does not exist in French, it's "Trophées".

"Tableau des scores" and "Panneau d'Administration" feel unnatural, let's use "Classement" and "Administration" which are also shorter (less chance of having a newline in the navbar).


ColdHeat:
I think these changes seem fine. 

I believe Crowdin will take this back into itself properly. If it doesn't I will manually copy over these changes.

ColdHeat:
Actually I forgot that this needs to compile the translations. 

I don't know the exact command to compile a specific po file but you can take a look at `pybabel compile -f -d CTFd/translations`. I will take a look at compiling them for you later if you don't get to it.

PR #2531, opened 2024-04-27
fix: 500 error no score

numanturle:
In the Core theme, if team mode is active and the user has not yet joined a team, a 500 server error occurs. I fixed it in this commit.


ColdHeat:
Technically the core theme is end of life (or pretty much there) but I think we can take bugfixes like this.

PR #2534, opened 2024-05-01
Changes to admin settings page for navigability

cgoecknerwald:
- Nebulous "Settings" becomes "Visibility" (we should probably rehome the 'pause' but I defer that to later work restructuring this page for navigability)
- Nebulous "Pages" becomes "Robots.txt" and moves near the bottom of the left-side navigation column
- Minor wording changes

image

image


ColdHeat:
I am okay with the visibility changes but I think robots.txt should be kept under the Pages section.

cgoecknerwald:
Current 'Pages' setting section doesn't show anything but Robots.txt for me, even after I create a new page, are there cases where it does?

ColdHeat:
No but I don't see why robots.txt should be top level instead of Pages which could have more things grouped under it. You can move more things into the Pages section but I don't think you should move robots.txt to the top level.

cgoecknerwald:
I wanted to take another shot at this, by basing it off Github repo's admin settings and improving grouping of concepts. I feel that the current organization of this page leaves room for improvement.
- Push integrations down to bottom of page
- Cohabitate related concepts (such as Accounts, or Appearance)
- Separate distinct concepts (separate out "Logo" from "Theme" settings; separate out "Pause" from "Visibility" settings)
- Add FA icons to speed comprehension

If you are on board, I would also:
- Actually create the pages I've proposed here
- Clarify wording on various settings (e.g. "Whether CTFd will attempt to sanitize HTML content from content." could be improved)
- Increase the number in-text references to the CTFd documentation
- Update the documentation to reflect these changes

## Github
image
## Proposal
image
image
image

## Current
image

ColdHeat:
Looks pretty good, I'm pretty happy with your proposed changes. Let me look a little more and I will clean up/merge.

PR #2541, opened 2024-05-21
Allow admins to create user API tokens

theopolis:
Hi folks, I am curious what you think about this feature. It adds functionality to the the `POST /api/v1/tokens` endpoint to optionally accept a `user_id` request parameter. If the caller is an admin, then this input `user_id` is the target for the API token creation.

The use case here is that we have automation that auto-registers users and provides them an API-key-only for interacting with CTFd (no UI). Thus we need a way to create tokens on their behalf.

Heads up that I want to test this thoroughly, and right now I am opening a PR to get feedback on direction/approach/applicability.

### DIFF
--- a/CTFd/api/v1/tokens.py
+++ b/CTFd/api/v1/tokens.py
@@ -90,6 +90,13 @@ class TokenList(Resource):
             expiration = datetime.datetime.strptime(expiration, "%Y-%m-%d")
 
         user = get_current_user()
+
+        # Allow admins to create API tokens for users
+        if is_admin():
+            user_id = req.get("user_id")
+            if user_id is not None:
+                user = Users.query.filter_by(id=user_id).first_or_404()
+
         token = generate_user_token(
             user, expiration=expiration, description=description
         )


ColdHeat:
Hi Ted!

If you have command line access you can do this via the shell similarly to how `manage.py` works. You import CTFd, create the app and use `with app.app_context()` and you can import the Users & Tokens models and create the tokens that you need. 

Something sort of like the following:

```python
from CTFd import create_app
from CTFd.models import db, Users, Tokens

app = create_app()
with app.app_context():
    t = Tokens()  # You will need to add the arguments here
    db.session.add(t)
    db.session.commit()
```

You could also create a plugin if you need it exposed via HTTP. 

I see the benefit of this capability but codifying it in the API means that we are blessing that admins have the ability to become users. And other admins as well. 

It's not that this isn't currently possible (resetting the user's password, using XSS) but I think there's some kind of better implementation. Maybe low privilege token, maybe service accounts, maybe this is behind a configuration but I don't know for sure.

Does creating the secondary script (perhaps it could be a simple HTTP service) or a plugin work for you?

theopolis:
Hey Kevin! Thanks for the great response, that's exactly what I wanted to know. I definitely understand wanting to preserve the expectations of admin control over users.

I will go with the plugin route for my use-case.

PR #2544, opened 2024-05-28
Add none-check to time filter funcs

cgoecknerwald:
Add additional layer of None-checks to the time filter functions, to prevent 500s. Although we already covered the bug, it's possible we missed some edge cases, so I am fixing it closer to the source of the error

### DIFF
--- a/tests/utils/test_ctftime.py
+++ b/tests/utils/test_ctftime.py
@@ -1,4 +1,5 @@
-from datetime import datetime, timezone
+from datetime import datetime as DateTime
+from datetime import timezone as TimeZone
 
 import pytest
 
@@ -190,8 +191,8 @@ def test_unix_time():
     """
     Tests that the unix_time function returns the correct value and fails gracefully for strange inputs
     """
-    assert unix_time(datetime(2017, 1, 1)) == 1483228800
-    assert type(unix_time(datetime(2017, 1, 1))) == int
+    assert unix_time(DateTime(2017, 1, 1)) == 1483228800
+    assert type(unix_time(DateTime(2017, 1, 1))) == int
     assert unix_time(None) is None
     assert unix_time("test") is None
     assert unix_time(1) is None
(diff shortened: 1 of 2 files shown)


ColdHeat:
I think overtime we should switch to using these Pascal Case aliases. I think I was wrong to use `import datetime` but I think the ambiguity of having `from datetime import datetime` is going to get confusing. `import datetime as dt` is going to lead to issues because we used the `dt` variable name. Thus I think using the cased aliases makes the most sense. The imports are broken up because on our isort version that's what it wants and I think it's easier to fix this later than to add the isort switch to fix this across the entire codebase. 

https://discuss.python.org/t/renaming-datetime-datetime-to-datetime-datetime/26279
https://github.com/PyCQA/isort/issues/1841

tests/utils/test_ctftime.py
```diff
@@ -173,3 +184,53 @@ def test_ctf_ended():
...
+    assert unix_time_millis("test") is None
+    assert unix_time_millis(1) is None
+
+
+def test_unix_time_to_utc():
+    """
+    Tests that the unix_time function returns the correct value and fails gracefully for strange inputs
+    """
+    assert unix_time_to_utc(0) == datetime(1970, 1, 1, tzinfo=timezone.utc)
+    assert unix_time_to_utc(1483228800) == datetime(2017, 1, 1, tzinfo=timezone.utc)
```
ColdHeat, line 217: Note this fails ont he previous commit because there did not used to be a timezone and there is one now.

ColdHeat: These times are different

```
>>> dt1 = datetime.utcfromtimestamp(1483228800)
>>> dt2 = datetime(2017, 1, 1, tzinfo=timezone.utc)
>>> dt1
datetime.datetime(2017, 1, 1, 0, 0)
>>> dt1.timestamp()
1483246800.0
>>> dt2.timestamp()
1483228800.0
```

PR #2553, opened 2024-06-05
Safer handling of None from get_current_user_attrs

cgoecknerwald:
`get_current_user_attrs` can and will return `None` and needs to be handled appropriately by the functions that rely on it. Some already had `None` checkers, some did not.

### DIFF
--- a/CTFd/utils/user/__init__.py
+++ b/CTFd/utils/user/__init__.py
@@ -123,7 +123,8 @@ def get_team_attrs(team_id):
 def get_current_user_type(fallback=None):
     if authed():
         user = get_current_user_attrs()
-        return user.type
+        if user and user.type:
+            return user.type
     else:
         return fallback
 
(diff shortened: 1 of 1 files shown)


CTFd/utils/user/__init__.py
```diff
@@ -123,7 +123,8 @@ def get_team_attrs(team_id):
 def get_current_user_type(fallback=None):
     if authed():
         user = get_current_user_attrs()
-        return user.type
+        if user and user.type:
```
cgoecknerwald, line 126: This is a marginal functionality change. If `user.type == None`, the prior functionality returns `None`. Now, we return `fallback`.

Although fallback is defaulted to `None`, calling `get_current_user_type(fallback='XYZ')` for a user with `user.type==None` now returns `'XYZ'` instead of `None`. This will be the case in [3 spots in the code](https://github.com/search?q=repo%3ACTFd%2FCTFd%20get_current_user_type&type=code) where we call `get_current_user_type(fallback="user")`.

If this change is concerning to you, I can change this line to:

`if user and hasattr(user, "type"):`

PR #2564, opened 2024-06-25
Update challenges.py

einar-lanfranco:
Hi, i add this to solve a problem i detected  when POST to  /api/v1/challenges/attempt  content-type include charset -> 'content-type: application/json; charset=utf-8'.

I hope it helps!


ColdHeat:
Perhaps the better way to check is this is to use https://tedboy.github.io/flask/generated/generated/flask.Request.is_json.html ?

PR #2567, opened 2024-07-05
Update smtp.py to add  List-Unsubscribe header.

whoo:
Add List-Unsubscribe header:
- rfc2369
- to avoid being classified as spam.
- to be compliant with CAN-SPAM.

use the sender's address to avoid creating a dedicated configuration field.


whoo:
Error with variable and email part.

PR #2568, opened 2024-07-06
Add List-Unsubscribe header on email

whoo:
Add List-Unsubscribe header:

- rfc2369
- to avoid being classified as spam.
- to be compliant with CAN-SPAM.
- use the sender's address get_config("mailfrom_addr") or get_app_config("MAILFROM_ADDR") without formatting  to avoid creating a dedicated configuration field.


ColdHeat:
I see the benefit here but these emails are transactional and not marketing emails. I do not believe that transactional emails need to support unsubscribing?

PR #2573, opened 2024-07-18
Fix display problem when Website Language is Set to Chinese(Both Traditional and Simplified)

eran0926:
According to [# 2447](https://github.com/CTFd/CTFd/pull/2447) we need to renamed the original `zh_TW` and `zh_CN` translation directories to `zh_Hant_TW` and `zh_Hant_CN` so Flask-Babel can identify it correctly. But after tests ,I think that we actually need to change `zh_CN` to `zh_Hans_CN` instead of  `zh_Hant_CN`. So I changed the dictionary name to `zh_Hans_CN`.(After research ,`zh_Hant_CN` probably means Traditional Chinese used in China)

Additionally, still according to [# 2447](https://github.com/CTFd/CTFd/pull/2447). I change the language setting options from  `中文` ,`國語` to `简体中文` (Simplified Chinese, zh_CN) and `繁體中文` (Traditional Chinese, zh_TW) by editing `CTFd\constants\languages.py`.
I think this will make it easier for users to distinguish them.(As a native user of Traditional Chinese, I like this)


eran0926:
I am sorry ,I forgot to delete zh folder , I'll close this PR and open a new one.

PR #2579, opened 2024-08-08
Remove warning

khiemdoan:
```
WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 28)
```


ColdHeat:
Thanks, if you have some kind of linting tool for these kinds of errors we could consider adding it to the linting Github flow.

PR #2580, opened 2024-08-11
Add email logging system

ColdHeat:
* Add email logger
* Add the `logextra` parameter to the sendmail function for email sending functions to sanitize the email before logging


pandatix:
Hello, quick question as there is additional logging of PII (email) in this PR, is there a list of all those PII stored in CTFd (either in DB or logs) ?

ColdHeat:
Can you elaborate on what you are looking for?

PR #2581, opened 2024-08-12
Update pip-tools 7.4.1

khiemdoan:


ColdHeat:
Hi there, what's the rationale/benefit of this change?

ColdHeat:
Seems like the main difference is pip's default resolver changing. I think that's worth the update.

PR #2588, opened 2024-08-28
assets.py: JS as module implies defer

erdnaxe:
See https://github.com/CTFd/core-beta/pull/88 for discussion.

My bad, I didn't know that `type="module"` on a `<script>` tag implies defer mode.


ColdHeat:
@erdnaxe Any concerns about this change? This lets us change the type of the script as well as add extra attributes if necessary.

PR #2589, opened 2024-08-29
perf: move from pip to astral-sh/uv package manager

GuacOn:
[uv](https://github.com/astral-sh/uv) is a drop-in replacement for pip that reduces the initial package installation time from ~40s to 5s.


ColdHeat:
I'm not against this but I am personally more in favor of pip. I feel like it is the blessed Python package manager and I'm concerned about the VC funded aspect of uv. 

I will [give it a shot](https://lucumr.pocoo.org/2024/8/21/harvest-season/) locally but I am hesitant.

ColdHeat:
I think related information: https://pythonspeed.com/articles/faster-pip-installs/

PR #2595, opened 2024-09-11
Add challenge byline

ColdHeat:
* Add the attribution field to Challenge model
* Add byline section to challenge viewing and updating


pandatix:
Good idea :tada:

Did not have time to test it, but does the current changes return the new Challenge attributes as part of the API ? Like a call to `/api/v1/challenges/<id>` returns the attribution data ? If not, it should be :smile: we need it for the Terraform Provider, elseway we cannot generate the deltas thus be idempotent.

Issue #2600, opened 2024-09-19
Initial database migration fails on Windows: sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (2013, 'Lost connection to MySQL server during query')


Jus-Codin:
**Environment**:

- CTFd Version/Commit: 3.7.3
- Operating System: Windows 11, Docker Desktop (WSL2)

**What happened?**

When starting up CTFd using `docker compose up` for the first time, the `ctfd-1` service errors out during the database migration. The specific migration causing the issue is [add_brackets_table.py](https://github.com/CTFd/CTFd/blob/edbc1e1ecb26576da1ce6349094a73e688ea9354/migrations/versions/9889b8c53673_add_brackets_table.py#L32). The `brackets` table is created before this error occurs which results in the error shown in #2556 when the service restarts. It must be noted that if the fix in https://github.com/CTFd/CTFd/issues/2556#issuecomment-2293569821 is done, the exact same query will work on the second run and allow the server to start successfully.

This issue only seems to occur on Windows, as it worked perfectly fine on Ubuntu when I tested it.

Based on the error messages, my guess is somehow MariaDB is not a fan of running on WSL2, but I was unable to pinpoint the exact issue causing this error. Some potentially related issues I could find online include:
- https://stackoverflow.com/questions/70784437/innodb-preallocating-147456-bytes-for-file-xxx-ibd-failed-with-error-2
- https://jira.mariadb.org/browse/MDEV-26381

Various online sources also suggest increasing max packet size and timeout of the database, but considering the simplicity and speed of the query I highly doubt that it is the correct fix here.

A potential workaround for this issue is mentioned in https://github.com/CTFd/CTFd/issues/2556#issuecomment-2293569821, where we just check if the `brackets` table and `bracket_id` column is added to the database. Since the services restarts on errors, the migrations will run again and work the second time round.

<details>
<summary>ctfd-1 Logs:</summary>

```
2024-09-19 20:04:38 Waiting for db to be ready
2024-09-19 20:04:44 ......db is ready
2024-09-19 20:04:48 INFO  [alembic.runtime.migration] Context impl MySQLImpl.
2024-09-19 20:04:48 INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
2024-09-19 20:04:49 INFO  [alembic.runtime.migration] Running upgrade  -> 8369118943a1, Initial Revision
[...]

Issue #2598, opened 2024-09-13
CTFD Docker install: Docker-entrypoint not found, no such file or directory


GrainedPanther:
**Environment**: Docker

- CTFd Version/Commit: 3.7.3
- Operating System: Windows 
- Web Browser and Version: Firefox 130.0

**What happened?**
I ran the docker installation proces, as described in the documentation on (https://docs.ctfd.io/docs/deployment/installation/). Copying the repository and changing the compose file went fine, when i composed the docker with docker compose i got an error with the installation of Ctfd-1:

"ctfd-1   | exec /opt/CTFd/docker-entrypoint.sh: no such file or directory"

This halts the docker setup proces in docker desktop and prevents the setup from completing. 

**What did you expect to happen?**

Docker file would download, docker container created and web application would open.

**How to reproduce your issue**

1. git copy the repository 
2. open the repository location with powershell, command prompt or other windows command line.
3. run docker-compose up
4. error shows up in command line

**Any associated stack traces or error logs**

mendoitarou:
I had the same problem for the same reason, but the following website provides a solution.

[Windows で Docker 環境構築時のエラー対処法【no such file or directory】](https://kakiblo.com/docker-windows/)

An English translation is provided below.

The reason for this is that the line endings in Git for Windows are CRLF, which means that they cannot be read in the Linux environment on Docker.

When installing Git for Windows, the ‘Configuring the line ending conversions’ setting defaults to this:’ If you select ‘Checkout as-is, commit Unix-style line endings’ (‘core.autocrlf’ is set to ‘input ‘), and you should be good to go.

If you run the following command to change the configuration, clone again and then recreate the container, it starts without any problems.

```
git config --global core.autocrlf input
```

**Translated with DeepL.com (free version)**

Issue #2556, opened 2024-06-11
v3.7.1  pymysql.err.OperationalError: (1050, "Table 'brackets' already exists")


Pd1r:
ctfd_1   | Waiting for db to be ready
ctfd_1   | db is ready
ctfd_1   | INFO  [alembic.runtime.migration] Context impl MySQLImpl.
ctfd_1   | INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
ctfd_1   | INFO  [alembic.runtime.migration] Running upgrade 5c4996aeb2cb -> 9889b8c53673, Add Brackets table
ctfd_1   | Traceback (most recent call last):
ctfd_1   |   File "/opt/venv/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1905, in _execute_context
ctfd_1   |     self.dialect.do_execute(
ctfd_1   |   File "/opt/venv/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
ctfd_1   |     cursor.execute(statement, parameters)
ctfd_1   |   File "/opt/venv/lib/python3.11/site-packages/pymysql/cursors.py", line 148, in execute
ctfd_1   |     result = self._query(query)
ctfd_1   |              ^^^^^^^^^^^^^^^^^^
ctfd_1   |   File "/opt/venv/lib/python3.11/site-packages/pymysql/cursors.py", line 310, in _query
ctfd_1   |     conn.query(q)
ctfd_1   |   File "/opt/venv/lib/python3.11/site-packages/pymysql/connections.py", line 548, in query
ctfd_1   |     self._affected_rows = self._read_query_result(unbuffered=unbuffered)
ctfd_1   |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ctfd_1   |   File "/opt/venv/lib/python3.11/site-packages/pymysql/connections.py", line 775, in _read_query_result
ctfd_1   |     result.read()
ctfd_1   |   File "/opt/venv/lib/python3.11/site-packages/pymysql/connections.py", line 1156, in read
ctfd_1   |     first_packet = self.connection._read_packet()
ctfd_1   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ctfd_1   |   File "/opt/venv/lib/python3.11/site-packages/pymysql/connections.py", line 725, in _read_packet
ctfd_1   |     packet.raise_for_error()
[...]

PR #2601, opened 2024-09-19
Add .gitattributes to keep .sh files with LF endings for Windows

Jus-Codin:
When trying to run CTFd on Windows using Docker Desktop, there is 2 issues that result in it being unable to run.

Some users might have Git configured to change line endings from LF to CRLF on checkout, resulting in Docker Desktop which runs the Docker Engine in WSL2 to be unable to interpret the line endings.

While it is possible to configure git to not auto-convert line endings, it would be better to specify to not change line endings for `.sh` and `.conf` files using `.gitattributes`, which would CTFd to run on Windows out of the box. Cheers to [this](https://github.com/CTFd/CTFd/issues/2598#issuecomment-2348927586) comment for identifying what was causing the issue.
  
Additionally, this PR also implements the [workaround](https://github.com/CTFd/CTFd/issues/2556#issuecomment-2293569821) for #2600. If this workaround is not ideal or a proper fix is identified, feel free to remove the changes made.


ColdHeat:
I am against the migration change as I discussed in the related issue. If the migration needs to run, it should probably run and we should see what the underlying issue is. I feel that the issue is more related to MariaDB or WSL so I will likely remove the migration change. 

As for the line ending changes once I have the chance to verify this on Windows I will merge.

Jus-Codin:
I found another fix for the migration issue in https://github.com/CTFd/CTFd/issues/2600#issuecomment-2365065037, lmk if it that fix is fine and I'll revert and add the fix accordingly

Jus-Codin:
I've just reverted the migration issue workaround and swapped to the fix mentioned above, feel free to remove it if it is undesirable.

ColdHeat:
On a fresh git clone (and also a zip download) I did not need to make any changes besides the volume change. I did not need to change the line endings. I was able to reproduce the bug though after manually changing the line endings. 

It looks like this is specific to users who have `git config --global core.autocrlf true` set. 

From my testing only the `.sh` file needs to be LF. And to get things working in Windows with WSL2 you do need to use one of the listed fixes. However, I really don't like the volume pattern personally.

ColdHeat:
This issue also still appears to happen on more recent versions of MariaDB.

ColdHeat:
From digging though this, the issue is MariaDB or WSL2's fault. This doesn't happen with `mysql:5.7`. 

CTFd shouldn't have to work around the lack of a proper `ALTER TABLE` functionality in the underlying database. 

I will accept the .gitattributes change but the volume change is something that needs to be worked out elsewhere. I will leave comments on the relevant issues to state that you need to change the volume to a dedicated volume.

PR #2606, opened 2024-09-26
Switch pybluemonday for nh3

ColdHeat:
Experimental branch for testing out switching from pybluemonday to nh3. Comments welcome on whether this is beneficial or detrimental.

### DIFF
--- a/CTFd/utils/security/sanitize.py
+++ b/CTFd/utils/security/sanitize.py
@@ -1,7 +1,12 @@
-from pybluemonday import UGCPolicy
+import nh3
+from copy import deepcopy
+
+HTML_ALLOWED_ATTRIBUTES = deepcopy(nh3.ALLOWED_ATTRIBUTES)
+
 
 # Copied from lxml:
 # https://github.com/lxml/lxml/blob/e986a9cb5d54827c59aefa8803bc90954d67221e/src/lxml/html/defs.py#L38
+# We specifically remove rel to support the link_rel parameter
 # fmt: off
 SAFE_ATTRS = (
     'abbr', 'accept', 'accept-charset', 'accesskey', 'action', 'align',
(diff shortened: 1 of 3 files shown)


ColdHeat:
I think this change makes sense but we need a lot of tests to make sure that we don't lose some functionality when switching from bluemonday to nh3. 

I do prefer bluemonday but it's a struggle to maintain the wheels.

ColdHeat:
I think we can use https://github.com/messense/nh3/releases/tag/v0.2.20 as it resolves https://github.com/messense/nh3/issues/49

eduardo010174:
@ColdHeat nh3 is rust with python.
Probably will finally build alpine conteiner with this change xD
Rust don't have problems with musl at all.

It's possible do speed comparation with nh3 vs pybluemonday?

ColdHeat:
@eduardo010174 Yes I'm aware that nh3 is just rust bindings. But the benefit is more about using something better maintained. I could invest some time to figuring out how to use something like [gopy](https://github.com/go-python/gopy) maybe but nh3 seems fairly good. 

I can add some speed tests to the pybluemonday repo.

ColdHeat:
I don't know why but in my testing here locally nh3 is slower compared to pybluemonday. 

```
❯ python speed_nh3.py 
26.094523000065237

❯ python speed_bluemonday.py 
8.120098500046879
```

This is for 300 runs of maybe 500 HTML test cases.

If I had to guess pybluemonday can initialize only one sanitizer and then re-use it.

ColdHeat:
It looks like this craziness is no longer considered viable considering a bunch of issues, comments, and archived repos. 

https://github.com/go-python/gopy/issues/370
https://github.com/golang/go/issues/65050#issue-2074509727
https://github.com/asottile-archive/setuptools-golang

Looks like nh3 is likely the only viable choice here longer term which I am reasonably okay with.

PR #2626, opened 2024-10-09
#2623 Fix saving bracket bug.

### LINKED ISSUE #2623: Saving new bracket few times creates new instances of the same bracket
**Environment**:

- CTFd Version/Commit: latest
- Operating System: Linux Debian 12
- Web Browser and Version: Firefox 130.0b6 (64-bit)

**What happened?**
Creating new bracket and clicking save several times adds new bracket each time instead of patching the existing one. 

It should save one bracket and then update it.

Steps to reproduce:

1. Go to admin->config
2. Go to Scoreboard brackets
3. Click Add New Bracket
4. Fill the form
5. Click Save several  times - each time you click the bracket is saved as new. 

There is a bug in code right here:

https://github.com/CTFd/CTFd/blob/0d12a13d6a39d5b9dc21c5efd4e29c4ca05e6128/CTFd/themes/admin/assets/js/components/configs/brackets/Bracket.vue#L115

Seems like a copy-paste bug. It should be this.bracket = response.data; 

I have created a fix and will send a PR soon.

br0xpl:
Finally fixes #2623


br0xpl:
@ColdHeat  - This time it has the actual bugfix! I even managed to recompile just the configs js.

CTFd/themes/admin/package.json
```diff
@@ -11,7 +11,7 @@
     "verify": "vite build; git diff --quiet --exit-code"
   },
   "devDependencies": {
-    "@vitejs/plugin-vue": "^3.0.1",
+    "@vitejs/plugin-vue": "^5.1.4",
```
ColdHeat, line 14: Why does this PR need to bump this dependency?

br0xpl: @ColdHeat  Without bumping this npm throws an peer dep
package.json wants:
@vitejs/plugin-vue ^3.0.1 -> requires vite ^3.0.0
vite ^5.1.2

vite ^5.1.2 != vite ^3.0.0 and thus error. 

I can run install with --legacy-peer-deps but this seems like using it wrong. Also when using --legacy-peer-deps I cannot regenerate just a single file I need to regenerate all static js-es.

If you preffer --legacy-peer-deps I can change this. Let me know.

The real fix is obiously just in Bracket.vue.


Here is the conflict:
```
npm error While resolving: @vitejs/plugin-vue@3.2.0
npm error Found: vite@5.1.2
npm error node_modules/vite
npm error   dev vite@"^5.1.2" from the root project
npm error
npm error Could not resolve dependency:
npm error peer vite@"^3.0.0" from @vitejs/plugin-vue@3.2.0
npm error node_modules/@vitejs/plugin-vue
npm error   dev @vitejs/plugin-vue@"^3.0.1" from the root project
npm error
npm error Conflicting peer dependency: vite@3.2.11
npm error node_modules/vite
npm error   peer vite@"^3.0.0" from @vitejs/plugin-vue@3.2.0
npm error   node_modules/@vitejs/plugin-vue
npm error     dev @vitejs/plugin-vue@"^3.0.1" from the root project```

Comment (ColdHeat) [reply CTFd/themes/admin/package.json:14]: We are using yarn so what happens if you use yarn?

Comment (br0xpl) [reply CTFd/themes/admin/package.json:14]: yarn warns about same problem:
```
warning " > bootstrap@4.3.1" has unmet peer dependency "popper.js@^1.14.7".
warning " > @vitejs/plugin-vue@3.2.0" has incorrect peer dependency "vite@^3.0.0".
```

but yarn lock seems to have this forced (like --legacy-peer-deps does) and thus it is just a warning. 

Looking at the code line 38 of the package vitejs/plugin-vue at versions 3.X.X it indeed requires version 3 vite. 
https://www.npmjs.com/package/@vitejs/plugin-vue/v/3.0.1?activeTab=code

If you want I can keep the wrong combination vite - vite/plugin-vue but I don't understand where it comes from - this seems like an error. I think PR https://github.com/CTFd/CTFd/pull/2473 was supposed to bump vite and should have also bumped the plugin-vue at the same time.

ColdHeat:
This PR shouldn’t need to bump any dependencies.

CTFd/themes/admin/package.json
br0xpl: Okay, I left the version conflict as it was although imho it is not correct.

New PR without bumping plugin-vue as you wanted.

ColdHeat: Okay I see the point of this. You're right this is the better approach since we need to build anyway.

PR #2627, opened 2024-10-10
Fix attribution for dynamic challenges

### LINKED ISSUE #2625: Missing attribution in Challenge API (GET)
**Environment**:

- CTFd Version/Commit: 3.7.4
- Operating System: NA
- Web Browser and Version: NA

**What happened?**

The API call (GET) to `/api/v1/challenges/<id>` does not return the `attribution` field, as stated in #2595.

Return the `attribution` field.

```
curl "http://my.ctf/api/v1/challenges/1"
```

NA

pandatix:
Fixes #2625


ColdHeat:
A better strategy I think would be to call the standard challenge read
function and then add the new custom attributes

pandatix:
Ok I understand that you want better future-proof code to avoid facing this issue next times :smile:

I suggest the following where updates to the inherited object will propagate, and the new attributes of the specific object will only require to be set here.

Issue #2632, opened 2024-10-12
Various issues with generated Swagger docs


Jus-Codin:
**Environment**:

- CTFd Version/Commit: https://github.com/CTFd/CTFd/commit/13efb55e73f378d4a3b88f07ed983848b4ab2458

## TL;DR
There are several issues with the swagger.json file generated by CTFd, and most of them is due to the [`validate_args`](https://github.com/CTFd/CTFd/blob/13efb55e73f378d4a3b88f07ed983848b4ab2458/CTFd/api/v1/helpers/request.py#L15) decorator.

Issues are as follows:
1. Missing `$ref` type definitions (as mentioned in #2574)
2. Incorrect specification of parameters for POST JSON body endpoints
3. Duplicate `get_flag_types` operationId for `/flag/types` and `/flag/types/<type_name>` endpoint

I'm interested in fixing these issues, but would like some feedback to see if the fixes are going in the right direction.

## Missing `$ref` type definitions (refer to above issue for details)
This issue is caused by the `RawEnum` definitions for the `field` query parameters. Pydantic converts these to references in the model schema to reduce repeated enum definitions, but `validate_args` [only includes `properties` from the schema](https://github.com/CTFd/CTFd/blob/13efb55e73f378d4a3b88f07ed983848b4ab2458/CTFd/api/v1/helpers/request.py#L38), resulting in the missing definition.

<details>
<summary>Sample affected schema (AwardFields)</summary>

```json
{
    "title": "",
    "type": "object",
    "properties": {
        "user_id": {
            "title": "User Id",
            "type": "integer"
        },
        ...
        "field": {
            "$ref": "#/definitions/AwardFields"
        }
    },
    "definitions": {
        "AwardFields": {
            "title": "AwardFields",
            "description": "This is a customized enum class which should be used with a mixin.\nThe mixin should define the types of each member.\n\nFor example:\n\nclass Colors(str, RawEnum):\n    RED = \"red\"\n    GREEN = \"green\"\n    BLUE = \"blue\"",
            "enum": [
                "name",
                "description",
                "category",
                "icon"
            ]
        }
    }
}
```
</details>

I have created a fix on my [fork](https://github.com/Jus-Codin/CTFd/blob/testing/CTFd/api/v1/helpers/request.py#L36-L39), which just resolves the reference and adds in the `enum` field.

## Incorrect specification of parameters for POST JSON body endpoints
[...]

PR #2633, opened 2024-10-13
Fix swagger docs

Jus-Codin:
Fixes for most issues in #2632 

I have some ideas of how to fix the incorrect specification for JSON POST bodies, but it would require a lot more changes. If you're fine with that @ColdHeat, I can work on that


Jus-Codin:
I was thinking of making `validate_args` able to specify the mime type for the request and response bodies using `consumes` and `produces` (https://swagger.io/docs/specification/v2_0/mime-types/) and then removing the [`Content-Type` security definition](https://github.com/CTFd/CTFd/blob/13efb55e73f378d4a3b88f07ed983848b4ab2458/CTFd/api/__init__.py#L43-L48) as it seems that the content type is supposed to be set on a per operation basis.

If we go down this route, I think it would be ideal to rename the decorator to properly describe it's actual behaviour since it does a lot more than just validate the request parameters?

ColdHeat:
@Jus-Codin I think a different function should probably specify the request and response bodies. validate_args was just supposed to be a way to simply leverage the existing models to validate requests a little more. I would consider integration with RestX just a happy accident. 

I think long term it would be better to replace Flask RestX with something like [APIFlask](https://apiflask.com/).

Jus-Codin:
Is there any target release for a version of CTFd with APIFlask? I was thinking if it was still a while off that it might still be worth it to put a hacky patch in the current code

ColdHeat:
There wouldn't be a big shift like that until a 4.0 version was going to be developed so you can put in the changes if you think it's worth it.

Jus-Codin:
I also notice that pydantic is used for input validation but marshmallow is used for loading. Would it maybe be preferrable to entirely replace one with the other? Either replacing pydantic by using APIFlask or another library for pydantic?

ColdHeat:
I would prefer to use pydantic entirely but marshmallow is what currently exists and integrates better with SQLAlchemy afaik. Migrating to pydantic will take some time.

Jus-Codin:
I've decided to not create a decorator to specify request and response mimetypes since there is only one endpoint (post_files) that accepts a multipart-form while the rest takes json. I've updated `validate_args` to correctly format the docs for these endpoints instead. If you put the swagger.json into https://editor.swagger.io/ it should not have any issues. I also notice that most endpoints do not seem to have args or models specified for validation, is there a specific reason for this?

ColdHeat:
@Jus-Codin Sorry for the delay, I will review this likely next week as I'm busy for the next few days. 

Can you give me an example? I believe you, I just would like to look at something more concrete. 

If I had to guess, I would say that I simply didn't enjoy working with the Flask-Restplus/Flask-RestX swagger generation and I simply wanted to get the API out the door. 

If you are interested I have offered money for people to sit down and overhaul the API but thus far it hasn't been completed/taken. If you're interested feel free to reach out to the email mentioned in https://github.com/CTFd/CTFd/blob/master/SECURITY.md and I can dig up the details.

ColdHeat:
> I also notice that most endpoints do not seem to have args or models specified for validation, is there a specific reason for this?

Can you give me an example of what you're talking about? I suspect you're saying that we don't use the Marshmallow schemas or Pydantic models to specify what an endpoint is expecting?

ColdHeat:
Yes this is a little bit b/c it was migrated from the previous versions of CTFd which used raw forms and I was an even worse developer way back then.

Challenges is a bit of an edge case b/c it has to ingest the regular challenge data as well as the custom challenge plugin data but we could specify at least a little bit of it. 

I think that it would be a good amt of manual work. It's also possible that your library has done a lot of converting requests to code so perhaps there's a way to integrate that in. 

I do think that likely if you want to specify this, it should be in documentation only (at least at first) because plugins can be used to override pretty much any endpoint and enforcing parameters could possibly break those plugins.

Jus-Codin:
> It's also possible that your library has done a lot of converting requests to code so perhaps there's a way to integrate that in.

This is a great idea! If we are able to fully (or almost fully) complete the swagger docs, it might actually be possible to programmatically generate the API for version 4.0 with APIFlask. However, it seems that APIFlask relies on marshmallow schemas and I can't seem to find any relatively mature flask frameworks which uses pydantic schemas. The other alternative would be to use FastAPI with Flask for frontend routing, but that would be an insane amount of changes.

> I do think that likely if you want to specify this, it should be in documentation only (at least at first) because plugins can be used to override pretty much any endpoint and enforcing parameters could possibly break those plugins.

Good point, I haven't actually really considered backwards compatibility. I'll agree with you that we should probably make it documentation only for now. I'm not too familiar with CTFd's plugin system, so I'm unsure how much access to CTFd internals a plugin could have.

Maybe we could add an additional argument to `validate_args` to specify if the endpoint should be validated or not, and make it `False` by default. This way, any plugins that use `validate_args` will still function correctly. It would be a bit weird to call the decorator `validate_args` if validation would be turned off by default, but renaming it could break some plugins. Maybe we rename it and make a new `validate_args` decorator with `validate=True`.

For the challenges edge case, what could be done is add another argument to `validate_args` to allow extras. (Pydantic ignores extra values by default)

Regarding time commitment, I'm currently relatively free for the next few weeks but I'll most likely be unavailable closer to the end of the year, but I think laying out the groundwork for this should be good enough.

CTFd/api/v1/helpers/request.py
```diff
@@ -23,19 +23,57 @@ def validate_args(spec, location):
...
     props = schema.get("properties", {})
     required = schema.get("required", [])
 
+    # Remove all titles and resolve all $refs in properties
     for k in props:
-        if k in required:
-            props[k]["required"] = True
-        props[k]["in"] = location
+        if "title" in props[k]:
+            del props[k]["title"]
```
ColdHeat, line 34: Is there a reason why this needs to be deleted?

ColdHeat: Ah I see you covered it in your original issue.

ColdHeat:
I think this PR is fine. If you could edit your issue description to just talk about the remaining issues that would be ideal. 

Regarding plugins, CTFd plugins basically can modify almost anything within CTFd but I think as long as there is a way for a plugin developer to bypass/disable the validation it would be okay.

ColdHeat:
One improvement for another PR I think might be if there's a way to test the validity of the generated swagger file.

PR #2637, opened 2024-10-15
Remove the "paused" message when the ctf is actually ended.

AAA3A-AAA3A:
Hello,
This small change is to remove the `<ctf> is paused` message when the ctf is actually finished (both messages are displayed at the same time, currently).
Thank you in advance for approving this PR.
Have a nice day,
AAA3A


ColdHeat:
I think a CTF can be both ended and paused. Not sure I agree with this change.

ColdHeat:
I see these notifications as informational and I think it's okay to have both of these.

PR #2643, opened 2024-10-23
Create new decorator for documenting endpoints

Jus-Codin:
Created a new `expects_args` decorator which is the old `validate_args` decorator but with support for extra unspecified paramters to be passed to it and optional validation.

### DIFF
--- a/CTFd/api/v1/helpers/request.py
+++ b/CTFd/api/v1/helpers/request.py
@@ -34,6 +34,9 @@ def expects_args(spec, location, allow_extras=False, validate=False):
     """
     if isinstance(spec, dict):
         spec = create_model("", **spec)
+    else:
+        # We have to make a copy in order to not overwrite the original config
+        spec = create_model("", __base__=spec)
 
     if allow_extras:
         spec.__config__.extra = Extra.allow
--- a/CTFd/api/v1/helpers/request.py
+++ b/CTFd/api/v1/helpers/request.py
@@ -1,7 +1,7 @@
 from functools import wraps
 
 from flask import request
-from pydantic import ValidationError, create_model
+from pydantic import Extra, ValidationError, create_model
 
 ARG_LOCATIONS = {
     "query": lambda: request.args,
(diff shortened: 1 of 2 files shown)


CTFd/api/v1/helpers/request.py
```diff
@@ -96,3 +119,23 @@ def wrapper(*args, **kwargs):
...
+
+    Example usage:
+
+    .. code-block:: python
+
+        @validate_args({"name": (str, None), "id": (int, None)}, location="query")
+        def my_route(query_args):
+            return {"success": True, "name": query_args["name"], "id": query_args["id"]}
+    """
+    return expects_args(spec, location, allow_extras, validate=True)
```
pandatix, line 141: Shouldn't we put a "deprecated" something here ? To make sure plugin developers have the info of a breaking change in their IDE.
Or do we expect developers to not use the `validate_args` decorator ?

Jus-Codin: Currently the decorator still has the same behavior as by default, just now with a new option to allow extras. We could mark it as depreciated, but I'll leave it up to @ColdHeat.

ColdHeat:
How are you expecting to make use of the `allow_extras` parameter? Is that for plugin purposes?

Jus-Codin:
Yep, it's mainly meant for plugins that add extra data to CTFd models as discussed [here](https://github.com/CTFd/CTFd/pull/2633#issuecomment-2423010789).

CTFd/api/v1/helpers/request.py
```diff
@@ -79,6 +98,10 @@ def decorator(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
             data = ARG_LOCATIONS[location]()
+
+            if not validate:
+                return func(*args, data, **kwargs)
+
```
Jus-Codin, line 104: I also just realised that if we add these decorators to existing endpoints, we'll have to change them to access the parameter data from the function arguments instead of from the request directly. We could maybe just make it not pass the data if validate is set to false. An alternative would be to do some inspection of the function to see if it will accept the argument, but not sure if we want to go down that route. 

```suggestion
            if not validate:
                return func(*args, **kwargs)

            data = ARG_LOCATIONS[location]()

```

Jus-Codin:
@ColdHeat After further digging around, I'm rather doubtful of the possibility of this being useful for documenting existing endpoints, mainly due to the inconsistency of the pydantic models' fields compared to the actual fields accepted by the endpoint. Examples would be the post_users and post_hints endpoints.
image

I suspect that this will be the case for most endpoints, which would mean the only way to get accurate documentation would be to get it from the marshmallow schema. The alternative is to manually document the endpoints using dicts, which is a decently viable possibility, but would make it a nightmare to add new fields to a model.

We could potentially add support for the decorator to accept a marshmallow schema as a spec, and either convert it to a pydantic schema or directly pass its values to the apidocs. My main worry would be marshmallow's usage of fields classes which makes it tricky to convert into types that either pydantic or json allows.

Issue #2641, opened 2024-10-21
Improving release process (SLSA and SBOMs)


pandatix:
Hello Kevin and folks here :grin:

The current CTFd build and release process **works**, but we can -and need- to go further.
Why ? TL;DR: **security**... [SLSA Level 3](https://slsa.dev/get-started#slsa-3), SBOMs, etc.

The current release process can be summarized as follows (to my understanding of course, feel free to correct anything).

![image](https://github.com/user-attachments/assets/1e6a79cb-8f6e-474f-af72-a4999f679a0a)

For a maintainer (i.e. Kevin) the release process requires three steps:
1. Maintain code, add features, improve others, fix bugs, review PRs, discuss issues, etc.
2. Create a tag
3. Create the release

A lot of manual steps... a burden which could and **must be avoided**: more time could be spent on the 1st step i.e. work on CTFd rather than its maintenance.
Moreover, with the current release process, we cannot ensure the integrity of the Docker images: they are not signed, thus could be tampered by anyone with access to the gcr.io or Docker Hub registries (not especially the CTFd one, the whole service could get compromised).

---

I advocate for a better release process that will save time and improve the security of CTFd: the maintainer will only need to create a tag. With what follows, we would be able to guarantee that the CTFd images (e.g. running in a Kubernetes environment as part of a CTF event) were not tampered thus what is supposed to run actually run. Moreover, with a more stable release process, we could generate SBOMs or other artifacts the community would need.

![image](https://github.com/user-attachments/assets/9087264b-3334-42d7-89e6-3a918dad905f)

As a CTF host, I want to be sure that I run CTFd, and that CTFd knows what it uses thus guarantee a minimum security level through dependency management.
For instance, I want to configure my Kubernetes cluster to only admit a Pod if the container has a valid signature through [Kyverno  and SLSA Level 3 compliance](https://kyverno.io/blog/2023/02/01/kyverno-and-slsa-3/). I would use the SBOM to pipe it into a VMS as [DependencyTrack](https://dependencytrack.org/) or other.

---

I'll work on a fork to submit a PR, but this will require multiple changes in the release process, thus would first like to discuss it with @ColdHeat to know if this effort would be acceptable.

I expect most things to be plug and play (e.g. the [SLSA Level 3 generator](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md#provenance-for-python)), but others would be technical choices.
[...]

PR #2644, opened 2024-10-23
Bring SLSA 3 on Docker containers (plug & play)

pandatix:
Hello,

This PR comes as a partial response to #2641: it brings SLSA 3 on the Docker containers for both [Docker Hub](https://hub.docker.com/repository/docker/pandatix/ctfd/tags) and [ghcr.io](https://github.com/pandatix/CTFd/pkgs/container/ctfd/versions).
It will be triggered as before, when a maintainer creates a release.
It still tags them for the specific tag and `latest`, build them for `linux` `amd64` and `arm64`, then pushes them.
Finally, it generates the provenance and publish them as OCI blobs.

For plugins, this does not imply any change : the CTFd Docker image could mount the plugin to a volume thus unchange the digest of the image (and so the attestation).
For end users, verifying SLSA compliance is not required thus unchange the current CTFd usage workflow.

I demonstrated the proper work of this PR with a [random tag](https://github.com/pandatix/CTFd/releases/tag/v1.0.0) which triggered the [workflow](https://github.com/pandatix/CTFd/actions/runs/11480508240). It comes with nearly no cost for the release process (<1 min).

Notice I dropped the usage of `DOCKER_USERNAME` secret as it will then be contained in an output, and for security purposes, GitHub Actions drop the content of the output that is passed through jobs... Frustrating hours of debug :upside_down_face:
Instead I use `github.actor`, but I'm unsure if `ctfdbot` (hardcoded string) fits there better :shrug: 

Dependencies are [pinned as a good practice from OpenSSF](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies).

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -10,37 +10,118 @@ jobs:
     permissions:
       contents: read
       packages: write
+    outputs:
+      dockerhub_image: ${{ steps.image.outputs.dockerhub_image }}
+      ghcrio_image: ${{ steps.image.outputs.ghcrio_image }}
+      digest: ${{ steps.build.outputs.digest }}
     steps:
+      - name: Checkout the repository
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+
       - name: Set repo name lowercase
         id: repo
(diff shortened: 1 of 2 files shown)


pandatix:
Up @ColdHeat

PR #2645, opened 2024-10-23
Add OpenSSF Scoreboard audit

pandatix:
Hello,

This PR comes as a partial response to #2641: it brings the [OpenSSF Scoreboard](https://openssf.org/projects/scorecard/) as a continuous security audit tool. It should help contributors that are non-developers but have a security profile to give them guidelines on how to help CTFd improve its security practices and posture.

My last manually-triggered audit showed **current CTFd scores 4.9/10** which highlights plenty room for improvement, but no big security defect. The expected behavior with Scoreboard is to raise both the score thus the security practices **and** the security knowledge of the community.
Moreover it has a cool badge :smile: 

Dependencies are [pinned as a good practice from OpenSSF](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies).

### DIFF
--- a/.github/workflows/scoreboard.yml
+++ b/.github/workflows/scoreboard.yml
@@ -1,0 +1,35 @@
+name: Scorecard supply-chain security
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: '30 6 * * 6'
+  push:
+    branches: [ "master" ]
+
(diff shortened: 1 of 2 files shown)


pandatix:
Up @ColdHeat :)

PR #2648, opened 2024-10-30
Language chooser in navbar

erdnaxe:
Currently, users are unable to change their language settings before connecting with an account.

This patchset proposes to:
- add a new "language" cookie, used when users are not logged in,
- prepare a form and expose `get_locale` for templates.

There is also a corresponding pull request in core-beta theme to use this new API: https://github.com/CTFd/core-beta/pull/99

I also plan in the future to propose 2 other features:
- add an option in admin to whitelist available languages (if only one language then the menu should be hidden),
- make challenge descriptions multilingual.

Example with patched core-beta


ColdHeat:
The idea is okay but I don't like the approach. Why not just store the language code in a cookie?

erdnaxe:
> Why not just store the language code in a cookie?
Oh! I haven't thought of using cookies directly. I will make a second iteration of this patch with cookies. Thanks!

erdnaxe:
@ColdHeat Now the patch no longer introduces a new API endpoint, and uses a `language` cookie which seems cleaner.

CTFd/forms/__init__.py
```diff
@@ -42,6 +43,7 @@ class _FormsWrapper:
 Forms.submissions = submissions
 Forms.users = users
 Forms.challenges = challenges
+Forms.lang = lang
```
ColdHeat, line 46: Should probably be called Forms.language. I am unsure if this definitely needs a new form.

CTFd/forms/lang.py
```diff
@@ -0,0 +1,10 @@
+from wtforms import RadioField
+
+from CTFd.constants.languages import LANGUAGE_NAMES
+from CTFd.forms import BaseForm
+
+
+class LanguageForm(BaseForm):
+    """Language form used in navbar"""
```
ColdHeat, line 8: This can probably be used elsewhere as well.

ColdHeat (commented):
Just some notes

CTFd/forms/__init__.py
erdnaxe: > I am unsure if this definitely needs a new form.

`forms.self.SettingsForm` cannot be used as it was designed to render data based on the logged in user.
`forms.users.UserEditForm/UserCreateForm` cannot be used as they need to refer to an existing user.

IMHO, we should keep the separate form as a cleaner solution.

CTFd/forms/lang.py
erdnaxe: I changed the comment to say that it is used in core-beta navbar (to help users understand why this form exists).

ColdHeat:
I actually meant the notes were just for me but thank you for working on them. I will try to merge this today.

ColdHeat:
I did try to write some tests for this but Flask refuses to see the cookie that I'm setting. If you have the time to investigate it would be good to get some tests but I am okay to merge this.

ColdHeat:
@erdnaxe I missed that you want to make challenge descirptions multilingual. I think that could be useful but it will be quite hard to implement. I do not know how you would get started with that.

PR #2650, opened 2024-11-06
Add exception handling for incorrect database password

puckk:
I added an exception to catch incorrect passwords, because otherwise it just keeps trying to connect forever (You can try it in docker-compose) without timing out or ending—especially during AWS deployments. To make things worse, it doesn’t provide any feedback to the user, so you’ll never know if it’s a network issue or an incorrect password. It nearly cost me my job, but at least it'll save you all a few precious hours of your lives!


ColdHeat:
Wouldn't the Docker container simply restart the container and it will attempt to reconnect again? 

Perhaps we can do a better job at emitting what kind of exception is generated. Or perhaps we should just go back to mysqladmin

puckk:
The container doesn’t actually attempt to reconnect or exit on its own—it just gets stuck indefinitely without any progress. There’s no timeout or retry logic, so it simply hangs, which makes debugging even harder. Adding the exception allows it to fail fast on incorrect credentials, saving a lot of troubleshooting time

ColdHeat:
Okay I understand what you're going for. I think this is more about making the exceptions visible instead of suppressing them. 

I think it might make more sense to only suppress the connection issues instead of any Exception? How does that sound?

ColdHeat:
I think something simple like this would be good. What do you think? 

```python
print(f"Waiting for {url.host} to be ready")
while True:
    try:
        engine.raw_connection()
        break
    except Exception as e:
        print(e)
        time.sleep(1)
```

PR #2651, opened 2024-11-07
Modify challenge attempt function for already solved challenges

hidehiroanto:
If challenge is already solved, tell user if flag is correct or not.


ColdHeat:
I'm not against this but this is a big change since this behavior has existed since probably the begining of CTFd. Is there a particular reason that you want this behavior?

hidehiroanto:
> I'm not against this but this is a big change since this behavior has existed since probably the begining of CTFd. Is there a particular reason that you want this behavior?

I ran into an issue where the flag was randomly generated but also the output was an obfuscated graphical representation, so I tried a bunch of solutions quickly, but I was unable to verify which one actually worked, other than contacting the CTF admins which is not ideal.

CTFd/api/v1/challenges.py
```diff
@@ -745,11 +745,16 @@ def post(self):
...
+            if message[-1] not in "!().;?[]{}":
+                message = message + "."
             return {
                 "success": True,
                 "data": {
                     "status": "already_solved",
-                    "message": "You already solved this",
+                    "message": "{} You already solved this.".format(
+                        message
+                    ),
```
ColdHeat: Instead of doing this punctuation tricks, I think better to just respond with "Correct but you already solved this." or that alternative "Incorrect".

ColdHeat:
I agree with merging this.

Issue #2197, opened 2022-10-10
Consider removing build-essential and other dev dependencies from Dockerfile


ColdHeat:
I think Python wheels are prevalent enough that we can consider removing the build dependencies so we can have a lighter Docker image.

ColdHeat:
Unsure if this was actually closed.

PR #2652, opened 2024-11-07
Dockerfile: use cffi wheel

erdnaxe:
cffi 1.15.0 does not include cp311-manylinux wheels, which forces us to use a build stage with build-essential.

By upgrading cffi to 1.15.1, we can now use the upstream wheels and simplify the Dockerfile.

Fix https://github.com/CTFd/CTFd/issues/2197

A nicer fix would be to identify which dependency caused cffi to be at version 1.15.0 and upgrade it.


ColdHeat:
Thanks for the PR. I think likely 1.15.0 is just the lowest to match the requirements of the other dependencies. 

However I think #2197 is no longer relevant as using the venv builder approach actually reduced file size the last time I checked but I will check again. 

If you have the time to investigate bumping the other cffi related dependencies that would help. I suspect pybluemonday will be dropped in favor of nh3.

ColdHeat:
Not comfortable with changing the Dockerfile at this time. I think using a seperate venv is the better idea.

PR #2653, opened 2024-11-07
Tracking: Fix "cache miss" bug which causes too many DB queries

adam-lebon:
# Current behavior

Before each request, the initialization process module executes the following process:
1. Query the cache to get the recent IPs of the current user
    - If the cache is missing, select the recent IPs from the database.
    - The comparison is computed using `datetime.datetime.now() - datetime.timedelta()` 
    - The MySQL query isn't timezone aware, the date will be encoded in local time without timezone information
2. If the request IP isn't stored in database OR if the request is not GET then update the database
    - The date column is filled using `datetime.datetime.utcnow()`
    - The cache is invalidated

# Bugs

The first bug is caused by the following condition:
```python
if (ip not in user_ips) or (request.method != "GET"):
```
Browser issue many HEAD requests and trigger this condition. This invalidates cache and leads to two SELECT queries and one UPDATE query before every HEAD request.

The second bug is caused by differences in timezone treatment. The Datetime column type of SQLAlchemy isn't timezone aware. Because of that, the SELECT query is encoded using local time, while the UPDATE query was encoded using UTC time.

If a CTFd server is hosted on a timezone >= UTC+1, then the SELECT query would always return an empty result. This results in cache invalidation + UPDATE query **before each HTTP request** (even GET requests).

# Fixes

The condition `if ... or request.method != "GET"` has been removed. I don't really understand why it has been introduced in the first place. It removes forced updates during HEAD requests.

The usage of `datetime.datetime.now()` has been replaced by `datetime.datetime.now(datetime.UTC)` in order to always use UTC time accordingly to the whole project, which uses the deprecated `datetime.datetime.utcnow()` 

# Further proposal

`datetime.now()` with default timezone should be avoided, since the `Datetime` column type of SQLAlchemy doesn't store timezone information.

Every `datetime.datetime.utcnow()` should be replaced by `datetime.datetime.now(datetime.UTC)` in order to follow the [python deprecation notice](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow)

### DIFF
--- a/CTFd/utils/initialization/__init__.py
+++ b/CTFd/utils/initialization/__init__.py
@@ -235,7 +235,7 @@ def init_request_processors(app):
             ip = get_ip()
 
             track = None
-            if (ip not in user_ips):
+            if ip not in user_ips:
                 track = Tracking.query.filter_by(
                     ip=get_ip(), user_id=session["id"]
                 ).first()
--- a/CTFd/utils/initialization/__init__.py
+++ b/CTFd/utils/initialization/__init__.py
@@ -235,7 +235,7 @@ def init_request_processors(app):
             ip = get_ip()
 
             track = None
-            if (ip not in user_ips) or (request.method != "GET"):
+            if (ip not in user_ips):
                 track = Tracking.query.filter_by(
                     ip=get_ip(), user_id=session["id"]
                 ).first()
--- a/CTFd/utils/initialization/__init__.py
+++ b/CTFd/utils/initialization/__init__.py
@@ -235,7 +235,12 @@ def init_request_processors(app):
             ip = get_ip()
 
             track = None
-            if ip not in user_ips:
+            if ip not in user_ips or request.method in (
+                "POST",
+                "PUT",
+                "PATCH",
+                "DELETE",
+            ):
                 track = Tracking.query.filter_by(
                     ip=get_ip(), user_id=session["id"]
                 ).first()
--- a/CTFd/utils/initialization/__init__.py
+++ b/CTFd/utils/initialization/__init__.py
@@ -237,7 +237,6 @@ def init_request_processors(app):
             track = None
             if ip not in user_ips or request.method in (
                 "POST",
-                "PUT",
                 "PATCH",
                 "DELETE",
             ):
--- a/CTFd/utils/initialization/__init__.py
+++ b/CTFd/utils/initialization/__init__.py
@@ -241,7 +241,7 @@ def init_request_processors(app):
                 ).first()
 
                 if track:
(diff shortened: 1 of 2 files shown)


ColdHeat:
Hello, thanks for the PR and the description. It makes sense but here is some additional background. 

Tracking is not intended to just track IPs, it is also intended to roughly track when state changes happen. So POST requests and such. For example, we should track an IP when a submission is made to the Tracking table. However in your code, this would only happen if a new IP was used. 

CTFd used to at some point really only be GET and POST but that's changed over time. 

So with your change, it would only track the usage of new IPs which loses some granularity. 

Instead I would suggest that we make the request method filter something like `request.method in ("POST", "PATCH", "DELETE")`. 

I don't have any concerns about the datetime change given the upcoming deprecation.

adam-lebon:
> Instead I would suggest that we make the request method filter something like request.method in ("POST", "PATCH", "DELETE").

Hello, the code only updates the date column, I don't really understand the point of forcing the date update when doing POST requests.

With this update, the user IP will always be monitored. It's just that the `date` field will update every 300 seconds.

If you require the date field to be updated during each POST request, I could edit my PR to remove this specific change (even if I don't really understand the point)

ColdHeat:
Ah yes I forgot that it only updates the date column. Still the point is to track roughly when a user uses an IP. It should be updating when the user uses the IP to make a state change. Why would it update every 300 seconds?

adam-lebon:
Sorry I mean that the date field is updated every hour. It should be enough to track user activities.

ColdHeat:
I think it's still better to track IP on any kind of state changing action. It's plausible that a user could do something that would be missed.

adam-lebon:
I reverted my change,
the PR should be ready to merge

ColdHeat:
CTFd doesn't really use PUT so I'm going to remove it. If it's needed it can be added in another PR.

ColdHeat:
It should be noted that switching from `utcnow` to `now` is not really a backwards compatible change. The good thing is that IP tracking is only done for tracking purposes and not for any kind of truly date sensitive action so this should generally be safe (at worst the date is inaccurate, not broken).

ColdHeat:
Some discussion on naive datetimes: https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated

ColdHeat:
Hmm I actually think there is rationale to say that the true issue here is using `datetime.datetime.now()` when we have stored `datetime.datetime.utcnow()` since in 3.11 utcnow isn't deprecated but just not recommended.

PR #2656, opened 2024-11-08
Dockerfile: remove Debian deps and cache CTFd requirements

erdnaxe:
The first commit removes unused apt-get dependencies, that are no longer needed thanks to https://github.com/CTFd/CTFd/pull/2652.

The second commit greatly improves build-time when rebuilding the Docker image during development. By separating requirements.txt installation in a layer, before copying CTFd source, we no longer need to reinstall all pipy deps after small code changes. It should also improve CI/CD speed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,10 +6,14 @@ RUN python -m venv /opt/venv
 
 ENV PATH="/opt/venv/bin:$PATH"
 
-COPY . /opt/CTFd
+# Install CTFd requirements before copying the whole project,
+# this layer will be reused when rebuilding CTFd after a small change
+COPY ./requirements.txt /opt/CTFd
+RUN pip install --no-cache-dir -r requirements.txt
 
-RUN pip install --no-cache-dir -r requirements.txt \
-    && for d in CTFd/plugins/*; do \
+# Install CTFd plugins requirements
+COPY . /opt/CTFd
+RUN for d in CTFd/plugins/*; do \
         if [ -f "$d/requirements.txt" ]; then \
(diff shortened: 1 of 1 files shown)


ColdHeat:
I think this will cause issues with other Python dependencies of plugins as they may not have been updated to provide wheels. I can explore this at some point but I think since plugins cannot specify apt dependencies this might be a non-starter.

Dockerfile
```diff
@@ -2,23 +2,18 @@ FROM python:3.11-slim-bookworm AS build
...
+# Install CTFd requirements before copying the whole project,
+# this layer will be reused when rebuilding CTFd after a small change
+COPY ./requirements.txt /opt/CTFd
+RUN pip install --no-cache-dir -r requirements.txt
 
-RUN pip install --no-cache-dir -r requirements.txt \
-    && for d in CTFd/plugins/*; do \
+# Install CTFd plugins requirements
+COPY . /opt/CTFd
+RUN for d in CTFd/plugins/*; do \
```
ColdHeat, line 16: I think I would accept just this change. 

This first stage is just a build stage so while it is slow, it should not be frequently re-run. 

I don't want to risk breaking some kind of downstream plugin so I would rather keep these apt dependencies.

erdnaxe: I agree with this option. Do you prefer to cherry-pick 01ecab3e78dfc74ed17776dcd6c1b740dc30f729 or should I remove the other commit?

ColdHeat: If you want you can force push just the changes from 01ecab3 but commits get squashed down so it's not a big deal. Whichever is easiest for you.

PR #2663, opened 2024-11-21
Revert datetime changes from #2653

ColdHeat:
@adam-lebon I am reverting the datetime changes here because I think it will be better to put it in an whole application wide migration

Also using utcnow in the recent IPs query to fix the issue you had raised


adam-lebon:
As long as utc and utcnow are not mixed up together I think this is fine

PR #2678, opened 2024-12-11
feat(plugin): add get_plugin_folder helper function

d3vyce:
This function allow a plugin to retrive his folder path to register assets.
This is especially useful when a plugin's git repo doesn't exactly correspond to the names present in the code.

I deliberately didn't touch the `register_plugin_assets_directory()` function, so I converted the path to a string.


ColdHeat:
I don't like the usage of inspect here to be honest. Would a symlink not work for you?

d3vyce:
I agree that `inspect` is not the most elegant solution. I can try to find another solution, but I have the feeling that it will be even more of a "hack".

The aim of this function is above all to simplify life for plugin developers/users.
It's often the case that after git cloning a plugin, CTFd doesn't start anymore because the folder doesn't have the same name as in the plugin code.
Some plugins specify this in their README, but it's rare and impractical for less technical users (as a symlink could be too).

It also unifies the dynamic declaration of the plugins folder to avoid this kind of thing: `__name__.split(“.”)[-2]`.

GramThanos:
I am not sure if your solution will work if the function is called from a script inside a subfolder of the plugin...

d3vyce:
Indeed 😢...
In any case the solution is not obtimal and does not solve the problem of the static declaration of the folder in the `config.json`.

PR #2679, opened 2024-12-15
Convert confirmation & reset password emails to single use

ColdHeat:
* Convert confirmation & reset password emails to single use by storing the email tokens in caching

### DIFF
--- a/CTFd/utils/security/email.py
+++ b/CTFd/utils/security/email.py
@@ -1,0 +1,40 @@
+from CTFd.cache import cache
+from CTFd.exceptions.email import (
+    UserConfirmTokenInvalidException,
+    UserResetPasswordTokenInvalidException,
+)
+from CTFd.utils.security.signing import hmac
+
+
+def generate_email_confirm_token(addr, timeout=1800):
+    nonce = hmac(addr)
+    cache.set(f"confirm_email_{nonce}", addr, timeout=timeout)
+    return nonce
+
+
+def verify_email_confirm_token(nonce):
+    addr = cache.get(f"confirm_email_{nonce}")
+    if addr is None:
+        raise UserConfirmTokenInvalidException
+    return addr
+
+
+def remove_email_confirm_token(nonce):
+    cache.delete(f"confirm_email_{nonce}")
+
+
+def generate_password_reset_token(addr, timeout=1800):
+    nonce = hmac(addr)
+    cache.set(f"reset_password_{nonce}", addr, timeout=1800)
+    return nonce
+
+
+def verify_reset_password_token(nonce):
+    addr = cache.get(f"reset_password_{nonce}")
+    if addr is None:
+        raise UserResetPasswordTokenInvalidException
+    return addr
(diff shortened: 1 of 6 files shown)


CTFd/utils/security/email.py
```diff
@@ -0,0 +1,40 @@
+from CTFd.cache import cache
+from CTFd.exceptions.email import (
+    UserConfirmTokenInvalidException,
+    UserResetPasswordTokenInvalidException,
+)
+from CTFd.utils.security.signing import hmac
+
+
+def generate_email_confirm_token(addr, timeout=1800):
+    nonce = hmac(addr)
```
br0xpl, line 10: @ColdHeat this should not be hmac of address, I would suggest it to be a random number (more than 128bit secure entrophy). 

This is becuase otherwise a single password reset link can be reused infinitely by initiating reset function. 

Or to say it differently: the key is **the same** in every password reset for the same user. Stealing this token once allows to reset the password many times in the future.

br0xpl:
@ColdHeat, one more thing: to be honest I don't know if we need to sign the tokens at all (hmac), two things:
1. If the token is random enough and is stored in redis and assuming noone has access to redis then random number without hmac is enough.
2. Still if we want to sign the token then it should be also verified during token verification.

ColdHeat:
@br0xpl The reason the HMAC is there is to allow revocation of the tokens by rotating the secret key. 

Signing would really only be useful if we wanted to give the user a message that their token expired but I don't think that's needed anymore. With the HMAC there's no way to verify it without having the original random bytes.

br0xpl:
@ColdHeat  sorry I was not clear enough - I did not mean signing, what I meant is that we don't need hmac at all. 

Let me try again, we can have two options:
1. Use random number without hmac and store it in redis. You don't need to hmac it becuase you can revoke tokens by simply removing them from redis. 
2. The other option is to use hmac as you are now but right now you are not validating the hmac -- look at functions verify_reset_password_token and verify_email_confirm_token -- you are just checking if they exist in redis. This is equal to simply having a random number in redis without hmac at the moment. If you want to keep hmac, then you should also at the end of verify_...._token functions perform a hmac check on the email taken from redis and check if it is equal to the provided nonce.

ColdHeat:
@br0xpl 

1. Revocation by redis key is more tedious in comparison to rotating the secret key and letting the redis keys expire. 

2. The hmac is built off of a random value, not the email. So there's no way to verify the hmac without also storing the random value (which seems unnecessary).

br0xpl:
2. Haha of course I forgot, you would have to store the random number as well ;)


1. So how do you want to revoke the token if there is no verification of hmac? You rotate the key but the token is still in Redis and still working.. this is exactly my original comment :-)

I don't see any point of using hmac here at all. Just random secure with enough entropy should be good to go.

ColdHeat:
@br0xpl Ah yes I understand now, thank you for clarifying. I will update the PR to use just the random bytes.

PR #2685, opened 2024-12-24
Autocomplete attribute on login, register and password_reset

janic0:
This PR is a small improvement for the input fields on the login, register and password_reset pages. 
It signals password managers to use
- `current-password` for login
- `new-password` for register and password_reset

See documentation for the `autocomplete` attribute here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#new-password

### DIFF
--- a/CTFd/themes/core/templates/login.html
+++ b/CTFd/themes/core/templates/login.html
@@ -30,7 +30,7 @@
 				</div>
 				<div class="form-group">
 					<b>{{ form.password.label }}</b>
-					{{ form.password(class="form-control", value=password) }}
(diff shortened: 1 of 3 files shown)


ColdHeat:
Hi there, can you put this change on https://github.com/CTFd/core-beta ?

janic0:
Hi @ColdHeat – Thanks for the response.

I've created https://github.com/CTFd/core-beta/pull/100. 
I'm assuming that you don't want to implement this change here then, so I'll close the PR.

PR #2692, opened 2025-01-21
Fixed plugin list vibility on configuration

GramThanos:
The plugins section on the configuration is visible even if the installed plugins do not implement a configuration page.

Example of the minor issue:
image

Changed the listing to first filter the plugins using Jinja `selectattr` filter and `defined` test
https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.selectattr https://jinja.palletsprojects.com/en/stable/templates/#jinja-tests.defined


ColdHeat:
I haven't tested this out but shouldn't `get_configurable_plugins()` return an empty list? Would it be better to improve the `{% if plugins %}` check instead?

GramThanos:
> I haven't tested this out but shouldn't `get_configurable_plugins()` return an empty list? Would it be better to improve the `{% if plugins %}` check instead? 

I had the same thought, but I didn't want to change `get_configurable_plugins()` because there isn't a `get_plugins()` function and thus I didn't want to break anything that is using this as "get all plugins".

ColdHeat:
I see what you're saying. 

What do you think about something like this:

```
				{% set plugins = get_configurable_plugins() %}
				{% if plugins %}
					<small class="section-title">Plugins</small>
					{% for plugin in plugins %}
						{% if plugin.config %}
							<li class="nav-item">
								<a class="nav-link rounded-0 py-1" href="#plugin_{{loop.index}}" role="tab" data-toggle="tab">
									<i class="fa fa-folder action-icon"></i>{{ plugin.name }}
								</a>
							</li>
						{% elif plugin.route %}
							<li class="nav-item">
								<a class="nav-link rounded-0 py-1" href="{{ plugin.route }}">
									<i class="fa fa-folder action-icon"></i>{{ plugin.name }}
								</a>
							</li>
						{% endif %}
					{% endfor %}
				{% endif %}
```

So that the route appears in the configuration panel as well as the navbar. It's not technically configuration but I think is still useful. Maybe change the icon to be a link icon instead of a folder

GramThanos:
Yes, this can work too.
A parameter could also be added on the URL, something like `config`, so that the plugins can handle it if they want to do so.
Maybe like:
```
{{ plugin.route ~ ('&' if '?' in plugin.route else '?') ~ 'config' }}
```

For some of my plugins, this would make more sense than generating a 2nd/duplicate configuration page.

GramThanos:
Hey, @ColdHeat , give me a direction and I will adapt the code.

ColdHeat:
@GramThanos I think this should be implemented as if it has a config setting it should render as a the HTML as it currently does. If it just has a route it should show as just a link using a link icon. The link should point to that route.

ColdHeat:
@GramThanos In my local testing I setup the following:

```
{% set plugins = get_configurable_plugins() %}
{% if plugins %}
	<small class="section-title">Plugins</small>
	{% for plugin in plugins %}
		{% if plugin.config %}
			<li class="nav-item">
				<a class="nav-link rounded-0 py-1" href="#plugin_{{loop.index}}" role="tab" data-toggle="tab">
					<i class="fa fa-folder action-icon"></i>{{ plugin.name }}
				</a>
			</li>
		{% elif plugin.route %}
			<li class="nav-item">
				<a class="nav-link rounded-0 py-1" href="{{ plugin.route }}">
					<i class="fa fa-folder action-icon"></i>{{ plugin.name }}
				</a>
			</li>
		{% endif %}
	{% endfor %}
{% endif %}
```

I think this is a more explicit idea than the jinja filters you have

GramThanos:
@ColdHeat I did some tests. And based on our discussion, here is my updated suggestion:
image

Puzzle icon to indicate a plugin/extension. Added a link icon to indicate that a plug-in does not have a special config template.
I have 1 concern though. Usually, the sidebar items will not reload the page, thus any unsaved changes will not be lost. Here this functionality breaks that. The easiest approach is to add a `target="_blank"` attribute on these links.

Issue #2697, opened 2025-02-06
Optional hint title


d3vyce:
A title can be added to an hint to give an indication of its content before purchase.

I've started preparing the PR, but I'm not too comfortable with updating core-beta theme.
I also need to create a PR on the core-beta repo and then add the commits here?

PR #2699, opened 2025-02-10
Feat/hint title

d3vyce:
Feature to add an optional hint title (see #2697)

Since the core-beta theme is also in another repo, I've split the modifications into 2 commits.


CTFd/themes/admin/assets/js/components/hints/HintEditForm.vue
```diff
@@ -175,11 +191,15 @@ export default {
...
     },
+    getTitle: function () {
+      return this.$refs.title.value;
+    },
     updateHint: function () {
       let params = {
         challenge_id: this.$props.challenge_id,
         content: this.getContent(),
         cost: this.getCost(),
+        title: getTitle(),
```
d3vyce, line 202: `title: this.getTitle(),`
I forgot to add a `this` as a result the hints update no longer works.

ColdHeat: @d3vyce Can you make a new PR with the fixes?

PR #2700, opened 2025-02-10
filter by bracket and set selected challenges decay

N04H2601:
- Added the filter by bracket option in Users and Teams pages.
- Added the possibility to change all decay values of selected challenges at once.
- One must run ``npm install`` then ``npm run build`` in ``CTFd/themes/admin`` then something like ``docker build -t my-ctfd .`` in the root directory to apply all changes then something like ``docker run -p 8000:8000 -it my-ctfd`` to run the instance.


ColdHeat:
I think decay doesn't make sense being in this PR. It also doesn't make sense as a bulk edit because decay doesn't exist on all challenges. 

Can you slim down this PR so it's just the bracket change?

N04H2601:
Did it work?

N04H2601:
image

PR #2707, opened 2025-02-17
Fix CVE-2025-230

parchinski:
This PR adds a middleware in wsgi.py that validates the Host header against configured ALLOWED_HOSTS and ALLOWED_PROXIES. It remains a no-op if these settings are not defined, ensuring backward compatibility with unconfigured instances.


parchinski:
If config.ini can not be read using the array syntax: i, i+1, i+2. Then changes to how it is parsed are necessary to make this functional.

PR #2713, opened 2025-02-26
Make some user-facing strings as translatable and regenerate messages.pot

erdnaxe:
Make the challenges page messages translatable.

Also make the bracket field translatable, as it is the only field during registration that is missing translation.

Finally, regenerate `messages.pot` to let crowdin find the new translations (is this enough to let people translate the new strings?).


CTFd/challenges.py
```diff
@@ -34,15 +35,15 @@ def listing():
     errors = get_errors()
 
     if Configs.challenge_visibility == ChallengeVisibilityTypes.ADMINS:
-        infos.append("Challenge Visibility is set to Admins Only")
+        infos.append(_l("Challenge Visibility is set to Admins Only"))
 
     if ctf_started() is False:
```
erdnaxe, line 40: @ColdHeat This condition seems impossible to trigger here, is this dead code?

ColdHeat: I believe this path is for admins to receive a visual cue that the CTF has not started. You should see it if you view the challenge page as an admin.

Issue #2649, opened 2024-11-01
File upload hashes should be displayed in UI


ColdHeat:
File upload hashes should be displayed in UI. Could be Admin Panel and/or user side

parchinski:
Before:
![Image](https://github.com/user-attachments/assets/21c8014f-f722-4883-b740-cca04c131fd3)

After:
![Image](https://github.com/user-attachments/assets/cf6d58f2-6dac-4c76-9110-3b2e622c94ad)

PR #2717, opened 2025-03-03
Close Issue #2649, Added file sha1sum hashes to admin UI for challenge files.

parchinski:
## Related Issues

Changes made to close: https://github.com/CTFd/CTFd/issues/2649

## Screenshots

Before:

Screenshot 2025-03-02 231214

After:

Screenshot 2025-03-03 000552

## Changes

- added "sha1sum" to object response from the api.
- added table column for sha1sum also added copy to clipboard functionality.

## Potential Additions

Adding a toast style popup when checksum is copied to clipboard successfully.


parchinski:
I don't like how the hash is aligned a bit higher than the file name. I'm gonna fix this in the morning.

parchinski:
Updated screenshot using only bootstrap to not have some jank gross inline css

image

CTFd/api/v1/challenges.py
```diff
@@ -753,7 +753,7 @@ def get(self, challenge_id):
         ).all()
 
         for f in challenge_files:
-            response.append({"id": f.id, "type": f.type, "location": f.location})
+            response.append({"id": f.id, "type": f.type, "location": f.location, "sha1sum": f.sha1sum})
```
ColdHeat, line 756: Yes this makes sense thanks for adding this. It should align with the responses for `/api/v1/files`.

CTFd/themes/admin/assets/js/components/files/ChallengeFilesList.vue
```diff
@@ -4,6 +4,7 @@
       <thead>
         <tr>
           <td class="text-center"><b>File</b></td>
+          <td class="text-center"><b>SHA-1 Hash</b></td>
```
ColdHeat, line 7: Maybe just call this "SHA1" ?

CTFd/themes/admin/assets/js/components/files/ChallengeFilesList.vue
```diff
@@ -15,6 +16,22 @@
             }}</a>
           </td>
 
+          <td class="text-center">
+            <div
+              class="d-flex flex-row align-items-center justify-content-center"
+            >
+              <span class="text-truncate d-inline-block mr-2">
+                {{ file.sha1sum || "No Hash Available" }}
```
ColdHeat, line 24: If no hash is available this should just be none, null, null, or blank. 

I wonder if perhaps this can be shown under the file name perhaps to save space?

ColdHeat (commented):
Not bad I think with some UI improvements should be good to go.

CTFd/themes/admin/assets/js/components/files/ChallengeFilesList.vue
parchinski: I'll just make this null for simplicity. I also moved the sha1sum below the filename to save horizontal space per your suggestion and it looks marginally better.

CTFd/themes/admin/assets/js/components/files/ChallengeFilesList.vue
```diff
@@ -115,6 +131,17 @@ export default {
...
+      if (!text) return;
+      navigator.clipboard
+        .writeText(text)
+        .then(() => {
+          // maybe we could add a toast notification here? I'm not sure what best practice for this would be
+        })
+        .catch((err) => {
+          console.error("Failed to copy text: ", err);
+        });
+    },
```
ColdHeat, line 144: Isn't there an existing function to handle copyToClipboard?

parchinski: I've just attempted a solution with this existing `copyToClipboard` function and received the following error.

image

I am remembering now this was my original reasoning for recreating the `copyToClipboard` function.

If you want, I can create a `copyToClipboard` utility within the admin theme assets.

parchinski:
Alternatively, because copying text from a web page is not a difficult task we can remove this button and the function. This results in the image below.

image

PR #2742, opened 2025-04-05
fix(cache): fix possible cache poisoning attack (#2741)

### LINKED ISSUE #2741: Path parameter not in cache key for `/api/v1/scoreboard/top/<int:count>`
https://github.com/CTFd/CTFd/blob/1346d4cbcf39a1bc0c3e79faacd26f02a2ba3c3f/CTFd/api/v1/scoreboard.py#L89-L98

The `count` path parameter is not used in the cache key, allowing anyone to poison the amount of teams in the scoreboard chart cache:
```js
// Wait 60 seconds before running this to ensure the cache is clean
const top1 = await fetch("/api/v1/scoreboard/top/1").then(res => res.json()).then(res => res.data)
const top10 = await fetch("/api/v1/scoreboard/top/10").then(res => res.json()).then(res => res.data)
console.log({ top1, top10 });
```
```
{
    "top1": {
        "1": {
            "id": 4,
            "account_url": "/teams/4",
            "name": "Team Bob",
            "score": 8288,
            "bracket_id": null,
            "bracket_name": null,
            "solves": []
        }
    },
    "top10": {
        "1": {
            "id": 4,
            "account_url": "/teams/4",
            "name": "Team Bob",
            "score": 8288,
            "bracket_id": null,
            "bracket_name": null,
            "solves": []
        }
    }
}
```

Image

The impact is very low, so I don't think this is a security vulnerability.

daanbreur:
Made improvements to caching mechanism inside of the `make_cache_key_with_query_string` function. Modified the `make_cache_key_with_query_string` function to accept `allowed_route_params` in addition to renaming `allowed_params` to `allowed_query_params`, and updated the logic to include route parameters in the cache key generation.

Fixes #2741

### DIFF
--- a/CTFd/api/v1/scoreboard.py
+++ b/CTFd/api/v1/scoreboard.py
@@ -93,7 +93,9 @@ class ScoreboardDetail(Resource):
     @check_score_visibility
     @cache.cached(
         timeout=60,
-        key_prefix=make_cache_key_with_query_string(allowed_query_params=["bracket_id"], allowed_route_params=["count"]),
+        key_prefix=make_cache_key_with_query_string(
+            allowed_query_params=["bracket_id"], allowed_route_params=["count"]
+        ),
     )
     def get(self, count):
         response = {}
--- a/CTFd/api/v1/scoreboard.py
+++ b/CTFd/api/v1/scoreboard.py
@@ -93,7 +93,7 @@ class ScoreboardDetail(Resource):
     @check_score_visibility
     @cache.cached(
         timeout=60,
-        key_prefix=make_cache_key_with_query_string(allowed_params=["bracket_id"]),
+        key_prefix=make_cache_key_with_query_string(allowed_query_params=["bracket_id"], allowed_route_params=["count"]),
     )
     def get(self, count):
         response = {}
--- a/CTFd/cache/__init__.py
+++ b/CTFd/cache/__init__.py
@@ -53,7 +53,9 @@ def make_cache_key(path=None, key_prefix="view/%s"):
     return cache_key
 
 
-def make_cache_key_with_query_string(allowed_query_params=None, query_string_hash=None, allowed_route_params=None):
+def make_cache_key_with_query_string(
+    allowed_query_params=None, query_string_hash=None, allowed_route_params=None
+):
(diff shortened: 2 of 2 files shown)


ColdHeat:
Thank you for making this. I think this is reasonable but might need a few tweaks. However you are failing tests so that will need to be fixed up.

daanbreur:
Hi, no problem love to help. However, All these tests seem to be in parts that are unrelated to what I changed.

ColdHeat:
The tests are failing likely b/c changing this function modified the behavior for those other endpoints.

ColdHeat:
I have a better approach for this however the issue is actually invalidation. Because we don't know what the possible values are for count.

daanbreur:
> The tests are failing likely b/c changing this function modified the behavior for those other endpoints.

I checked there is absolutely nothing using this function only that scoreboard route

daanbreur:
> I have a better approach for this however the issue is actually invalidation. Because we don't know what the possible values are for count.

What is your idea for a better approach of doing it?
And do we need to have invalidation, if we do, do we _need_ to know all possible bracket and count combinations?

Since the keys/paths are already in a pretty nice pattern (`view/api.scoreboard_scoreboard_detail/bracket_hash` and `view/api.scoreboard_scoreboard_detail/count_hash`) we could just grab everything starting with the `view/api.scoreboard_scoreboard_detail/` for doing the `clear_standings()`

ColdHeat:
I just merged the path arguments into the query arguments and let the existing hashing logic do the rest of the work. 

I think your strategy for invalidation might make the most sense. The only issue is that on the filesystem cache I don't think we can do a search and it's not optimal to do it on redis.

daanbreur:
> I just merged the path arguments into the query arguments and let the existing hashing logic do the rest of the work.
> 
> I think your strategy for invalidation might make the most sense. The only issue is that on the filesystem cache I don't think we can do a search and it's not optimal to do it on redis.

With your strategy we will still have the issue of cache invalidation and might make it even harder

ColdHeat:
Closing this in favor of #2744. There unfortunately isn't a great way to invalidate the cache here without just replicating the logic from flask-caching.

PR #2744, opened 2025-04-12
Extract scoreboard detail logic to make it easier to cache per count and per bracket values

### LINKED ISSUE #2741: Path parameter not in cache key for `/api/v1/scoreboard/top/<int:count>`

ColdHeat:
* Extract scoreboard detail logic to make it easier to cache per count and per bracket values
* Closes #2741

### DIFF
--- a/CTFd/api/v1/scoreboard.py
+++ b/CTFd/api/v1/scoreboard.py
@@ -4,15 +4,15 @@ from flask import request
 from flask_restx import Namespace, Resource
 from sqlalchemy import select
 
-from CTFd.cache import cache, make_cache_key, make_cache_key_with_query_string
-from CTFd.models import Awards, Brackets, Solves, Users, db
+from CTFd.cache import cache, make_cache_key
+from CTFd.models import Brackets, Users, db
 from CTFd.utils import get_config
-from CTFd.utils.dates import isoformat, unix_time_to_utc
 from CTFd.utils.decorators.visibility import (
     check_account_visibility,
     check_score_visibility,
 )
 from CTFd.utils.modes import TEAMS_MODE, generate_account_url, get_mode_as_word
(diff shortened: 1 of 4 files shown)


ColdHeat:
This could possibly use a test that checks to see that different count values are cached seperately.

daanbreur:
Guess im late to the party, but personally I don't really like the having a specified range of rows you can get. There might be reasons you want actually usable scoreboard data for more.

ColdHeat:
@daanbreur Yeah I understand however I think that query can cause a lot of slowdowns. Also in the original versions of CTFd the limit was 20 and here its bumped up to 50. If there is an issue someone can just open an issue and we can discuss raising it or making it configurable.

PR #2759, opened 2025-05-24
Add pyproject.toml for dependency and package management

calebstewart:
Currently, CTFd uses a basic `requirements.in` file for dependency management. This is not ideal in modern Python packages, and also requires manual human dependency and conflict management/resolution. This PR adds `pyproject.toml` with dependencies accounted for with matching versions from the latest `requirements.in`. The `pyproject.toml` in this case uses Poetry for packaging, and the PR includes a `poetry.lock`.

Having a real project definition also allows you to manage development dependencies within the same file and with the same tools, so `development.txt` can also go away (all items from `development.txt` were also added to `pyproject.toml` via `poetry add -G dev`).

The real benefit though, in my opinion, is that this makes development of out-of-tree plugins much easier. For instance, a plugin being developed in an external repository can gain typing and LSP support for CTFd-related imports by simply adding a development dependency on CTFd itself (in Poetry: `poetry add -G dev path/to/CTFd`). Further, this dependency can be version controlled, and the dependencies of the plugin and CTFd are automatically deconflicted. The plugin developer will be explicitly warned if there are version conflicts (as opposed to the current "just install the requirements file and pray" method). It gets even easier if you were to eventually push the CTFd package to PyPI, although that's not strictly required.

I know this is a bit out of the blue, so I'm happy to discuss but I thought I'd get the ball rolling with a PR. :)


requirements.txt
ColdHeat, line 1: I think we should be keeping requirements.txt regardless of whichever underlying package manager we use.

ColdHeat:
@calebstewart can you reduce this PR so that it is only introducing pyproject.toml? I do not think that introducing pyproject.toml also needs to additionally introduce a new package manager.

Issue #156, opened 2016-09-22
Email everyone form


ColdHeat:
Need a form to email all teams

ColdHeat:
I actually think this feature can be abused. I'm not sure I like it as much anymore.

ColdHeat:
The potential for abuse here is too high. I think it can exist as perhaps a function, but it shouldn't exist as something accessible from the admin panel.

nu11pointer:
People need this feature @ColdHeat 
This is super useful to host CTFs and advertise when a new challenge comes out for instance, or to send notifications, since the current notifications system does not allow to send emails.

ColdHeat:
@nu11pointer You are right. We do need this.

ColdHeat:
There's tons of abuse potential in this and it's very hard to mitigate. 

Here is what I've come up with so far:

When a user is registered, they can:
    - They can check a box to say "subscribed to announcements"
    - User will also receive an email that they can click to confirm their email address

When a user is manually added or bulk added, they are by default unsubscribed
    - After logging in they can choose to subscribe their email in their profile

Can an admin subscribe a user themselves? 
    - CTFd is not fundamentally a mailing list. If a user themselves wishes to opt out of emails then an admin should not be able to resubscribe them. 

Users can subscribe at register and unsubscribe via link or in their profile

ColdHeat:
Started working on nudging users to verify their email addresses more frequently: 
https://github.com/CTFd/CTFd/pull/2791

PR #2775, opened 2025-07-16
Admin Email All Users

parchinski:
Add "Email All". Admins can now send all Users an email from the Admin Panel. Includes protections that should keep this under control.

https://github.com/CTFd/CTFd/issues/156


CTFd/admin/__init__.py
```diff
@@ -67,6 +71,13 @@ def view():
     return redirect(url_for("auth.login"))
 
 
+@admin.route("/admin/email_all", methods=["GET", "POST"])
```
ColdHeat: This controller should be on a new page following the pattern that all the others do. This should be a new top nav link.

ColdHeat: Also this should be `/admin/emails`

CTFd/forms/email.py
```diff
@@ -8,3 +8,18 @@
...
+    body = TextAreaField("Body", validators=[DataRequired()])
+    submit = SubmitField("Email All")
+
+    def validate(self):
+        if not super(EmailAllForm, self).validate():
+            return False
+        # Enforce plain-text only (no HTML)
+        if "<" in self.body.data or ">" in self.body.data:
+            self.body.errors.append("Plain text only no HTML allowed")
+            return False
```
ColdHeat: Not sure this is needed or even if this validation is needed?

CTFd/themes/admin/assets/js/pages/email_all.js
```diff
@@ -0,0 +1,153 @@
+import "./main";
```
ColdHeat: I would rather see this in Vue

ColdHeat (changes requested):

CTFd/forms/email.py
parchinski: Definitely unnecessary, removing this now.

CTFd/themes/admin/assets/js/pages/email_all.js
parchinski: Fixed.

CTFd/forms/email.py
parchinski: Fixed.

CTFd/admin/__init__.py
parchinski: Fixed.

parchinski:
I am going to wait on building the theme until I perfect the Vue component locally. So I only need to build once in Codespaces.

ColdHeat:
```
Registering a user
    - User can check a box to say "subscribed to announcements"
    - User will receive an email that they can click to confirm their email address

When a user is manually added, they are by default unsubscribed. After logging in they can choose to subscribe their email. 

When a user is bulk added, they are by default unsubscribed. After logging in they can choose to subscribe their email. 

Can an admin subscribe a user themselves? 

CTFd is not fundamentally a mailing list. If a user themselves wishes to opt out of emails then an admin should not be able to resubscribe them. 

Users can subscribe at register and unsubscribe via link or in their profile
```

PR #2778, opened 2025-07-22
Add API endpoints and database models for challenge solutions

ColdHeat:
* Add ability to store challenge solutions along with a challenge
* Add `/api/v1/solutions` and `/api/v1/solutions/[solution_id]`
* `/api/v1/unlocks` are now also used to unlock solutions for viewing by users
* `solution_id` is now included in the Challenge object response. If there is no visible solution the challenge has a null solution_id. 
* Files can now be stored as SolutionFiles which are connected to the Solution that they are associated with


ColdHeat:
Related to https://github.com/CTFd/CTFd/pull/2396

PR #2810, opened 2025-08-18
Add python 3.13 for actions

khiemdoan:


ColdHeat:
I do think this is a good idea but the tests are quite slow overall and testing all of these versions will make it take awhile. Perhaps only add this for sqlite or MySQL?

Issue #2681, opened 2024-12-18
Cache busting logic inconsistent with production deployment


adamdoupe:
This is less a bug and more of a design issue with the [cache busting](https://github.com/CTFd/CTFd/commit/69dc0a799ace1596d5a5f079db92684907b63882) functionality.

Production deployment of CTFd involves running more than one worker (as mentioned in the documentation) using the `WORKERS` environment variable in the Dockerfile, for instance:
https://docs.ctfd.io/docs/deployment/installation

However, the cache busting functionality creates a unique `run_id` based on `datetime.datetime.utcnow()`:
https://github.com/CTFd/CTFd/blob/master/CTFd/__init__.py#L57C1-L61C56

The issue here is that each worker in a production deployment will have a different `run_id`, in which case the user will get a different cache busting URL for each request depending on what back-end worker that they visit. This significantly decreases the chances of data being cached on the client-side browser (and thus, the cache being useful). 

Not sure what exactly is the best way to fix this, reducing the precision of time used for `run_id` would increase the chances that the workers all use the same time. Another way might be to allow the container (or whoever runs the worker) to specify an environment variable (such as `RUN_ID`) that each worker will use if defined.

ColdHeat:
I believe you're referring to the code here: https://github.com/CTFd/CTFd/blob/d7632b45422c93ace31dc52cff1702dbed305204/CTFd/utils/helpers/__init__.py#L50-L53

The code you linked uses max_age which doesn't get busted. 

You're right since I believe this code was to cache bust b/c the vite development server was caching very aggressively. 

I will have to dig through this and see what an alternative solution would be but if you have a strategy feel free to share. Likely it could just be based on server start time.

PR #2839, opened 2025-09-08
feat: Allow deployment to specify RUN_ID environment variable 

adamdoupe:
Allow deployments to specify a `RUN_ID` environment variable so that all workers will use the same run_id for cache busting (see #2681).

Logic is:

- If in debug mode, use prior behavior run_id (time-based)
- Otherwise, if RUN_ID environment variable is set, use that, otherwise fall back to time-based.

Should be backward compatible but allow deployments to specify a RUN_ID.

### DIFF
--- a/CTFd/__init__.py
+++ b/CTFd/__init__.py
@@ -60,15 +60,15 @@ class CTFdFlask(Flask):
         # Store server start time
         self.start_time = datetime.datetime.utcnow()
 
-        # Create time-based unique run identifier
-        time_based_run_id = sha256(str(self.start_time))[0:8]
+        # Create time-based run identifier
+        self.time_based_run_id = sha256(
+            str(round(self.start_time.timestamp() / 60) * 60)
+        )[0:8]
 
-        if self.debug:
-            # When in debug we want a new run_id each time worker starts
-            self.run_id = time_based_run_id
-        else:
-            # use RUN_ID if exists, otherwise fall back to time_based_run_in
-            self.run_id = os.getenv("RUN_ID", time_based_run_id)
+    @property
+    def run_id(self):
+        # use RUN_ID if exists, otherwise fall back to time_based_run_in
+        return self.config.get("RUN_ID") or self.time_based_run_id
 
     def create_jinja_environment(self):
         """Overridden jinja environment constructor"""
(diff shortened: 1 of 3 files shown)


ColdHeat:
Thank you for submitting this. 

Do you still find that you need this on 3.8.0? 

The new core theme has a nonce in the filename itself which makes it a bit easier to cache bust.

adamdoupe:
Didn't realize that functionality was in there, very nice. 

Yes, for much of the core theme having the nonce in the name is great and does help.

However, there are still assets by default that don't use that (from what I can tell) and get a varying `d=` cache buster based on number of workers. 

From my local test of the index I see:

```
<link rel="shortcut icon" href="/themes/core/static/img/favicon.ico?d=76c24792" type="image/x-icon">
<img class="w-100 mx-auto d-block" style="max-width: 500px;padding: 50px;padding-top: 14vh;" src="/themes/core/static/img/logo.png?d=8edcff22" />
```

Also, any themes content won't get the nonce-based functionality.

ColdHeat:
@adamdoupe I've adjusted the PR to allow specifying the RUN_ID via config.ini in addition I've rounded the time_based_run_id to the nearest minute so deployments are more likely to get the same query parameter. What do you think about this?

adamdoupe:
Very cool! 

I think rounding to the nearest minute is a great fallback for production, but I worry that this will break people's dev workflow. 

If people rely on `python serve.py` to dev/debug, and specifically flask's auto reload feature, if they're doing quick dev on assets that are cached in a minute they won't see the asset change unless they blow away the browser's cache. 

This is why I added the `self.debug` check: in my mind these are all production optimizations (including the minute rounding), so you'd want to disable them during development/debugging. 

I would recommend fallback to the prior behavior when debugging.

ColdHeat:
@adamdoupe Good point, should be good to go now!

pandatix:
Out of curiosity, is it configurable only through the `config.ini` or does the PR still support getting it from env ?

With the use case of Kubernetes-based deployment, the instance ID (e.g. a STS ID) is easy to pass to CTFd containers with environment variable while the config.ini, even if possible, requires more integration work.

Issue #2849, opened 2025-09-16
Ability to bulk edit solution visibility and allow visibilty if solved


adb014:
With the inclusion of solutions in version 3.8.0, the solution visibility is updated in the challenge itself. To make it easier to make all challenges (or a large group of challenges) solutions visible, it would be good to have the solution visibility updated in the same manner as the challenge visibility in the challenge list.

I had this in my plugin (https://github.com/adb014/CTFd-solutions) and I'd dearly like it in 3.8.0. If you are ok with this change I could propose a pull request

ColdHeat:
I think this is okay but I am unsure of exactly how this will look alongside other unlocking behaviors.

adb014:
I created a PR for this change. I'll look at the visible if solved request afterwards

adb014:
Now that the PR is merged, closing this issue

PR #2851, opened 2025-09-17
Add bulk editing of solution visibility

adb014:
See issue #2849

### DIFF
--- a/CTFd/admin/challenges.py
+++ b/CTFd/admin/challenges.py
@@ -1,7 +1,7 @@
 from flask import abort, render_template, request, url_for
 
 from CTFd.admin import admin
-from CTFd.models import Challenges, Flags, Solves
+from CTFd.models import Challenges, Solutions, Flags, Solves
 from CTFd.plugins.challenges import CHALLENGE_CLASSES, get_chal_class
 from CTFd.schemas.tags import TagSchema
 from CTFd.utils.decorators import admins_only
@@ -24,6 +24,15 @@ def challenges_listing():
     query = Challenges.query.filter(*filters).order_by(Challenges.id.asc())
     challenges = query.all()
     total = query.count()
+
+    for challenge in challenges:
+        solution = Solutions.query.filter_by(challenge_id=challenge.id).first()
+        if solution:
+            challenge.has_solution = True
+            challenge.solution_state = solution.state
+        else:
+            challenge.has_solution = False
+            challenge.solution_state = "hidden"
 
     return render_template(
         "admin/challenges/challenges.html",
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -24,6 +24,7 @@ from CTFd.models import (
     Hints,
     HintUnlocks,
     Ratings,
+    Solutions,
     Solves,
     Submissions,
     Tags,
@@ -35,6 +36,7 @@ from CTFd.schemas.challenges import ChallengeSchema
 from CTFd.schemas.flags import FlagSchema
 from CTFd.schemas.hints import HintSchema
 from CTFd.schemas.ratings import RatingSchema
+from CTFd.schemas.solutions import SolutionSchema
 from CTFd.schemas.tags import TagSchema
 from CTFd.utils import config, get_config
 from CTFd.utils import user as current_user
@@ -552,6 +554,7 @@ class Challenge(Resource):
     )
     def patch(self, challenge_id):
         data = request.get_json()
+        solution_state = data.pop("solution", None)
 
         # Load data through schema for validation but not for insertion
         schema = ChallengeSchema()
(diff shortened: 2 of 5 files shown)


adb014:
I realised that there are several ways of implementing this and you might not be ok with the way I did it. What I did is add a "solution_state" field to the PATCH api of "/api/v1/challenges/<id>", then the logic in the serialized data from "pages/challenges.js" edit button doesn't need to be changed. You might prefer to split the serialized data in "pages/challenges.js" in to two parts and use "/api/v1/solutions/<solution_id>" to patch the solution visibility and thus keep the same REST API as currently.

Frankly, thinking about I prefer the second option as well. I'll try to implement it as an additional patch to this PR

adb014:
Ok, I've implemented it without changing the existing PATCH api of /api/v1/challenges/<id>

adb014:
As the solved patch depends on the bulk editing patch,, I just combined them. You can ignore the pull request [#2852](https://github.com/CTFd/CTFd/pull/2852) and just look at this one

ColdHeat:
This works but the UI isn't the best because we don't query to reload the challenge until the challenge board button is clicked again.

adb014:
> This works but the UI isn't the best because we don't query to reload the challenge until the challenge board button is clicked again.

Not sure I understand. If you mean the endpoint "/challenges#chall" is not updated if an administrator changes the solution visibility before the user presses the challenge button again, this is already the case and in a much wider sense if the administrator changes the description of the challenge it won't be updated till the user clicking the challenge button again either... So this PR doesn't change this existing issue (if it really is an issue)... This is why I suspect you mean something different.

Can you explain further what you mean

ColdHeat:
The issue is that if the solution visibility is `solved`, the solution tab isn't shown when the user first solves the challenge. It's only shown after the user refreshes or loads the challenge again.

adb014:
Ok I see two ways of doing what you want 

### Idea 1
- Change the "/api/vi/challenges/N" endpoint so that solution_id is always supplied but a boolean solution_visible attribute is added to flag if the solution is already visible. 
- Then in challenges.js add a getSolutionVisible method that returns `Alpine.store("challenge").data.solution_state` 
- Then in challenge.html change

```
          {% block solution %}
            <li class="nav-item" x-show="(getSolutionVisible() || response.data.status == 'correct' || response.data.status == 'already_solved'">
              <button class="nav-link challenge-solution" data-bs-target="#solution" @click="showSolution()">
                {% trans %}Solution{% endtrans %}
              </button>
            </li>
          {% endblock %}
```

### Idea 2
- Force challenges.js:renderSubmissionResponse to call /api/v1/challenges/N if the status of the attempt response is "solved"  or "already_solved" and reset the value of Alpine.store("challenge").data
- Then modify challenge.html like

```
          {% block solution %}
            <li class="nav-item" x-show="getSolutionId()">
              <button class="nav-link challenge-solution" data-bs-target="#solution" @click="showSolution()">
                {% trans %}Solution{% endtrans %}
              </button>
            </li>
          {% endblock %}
```

The second idea is less intrusive but needs an extra API call. Which solution do you prefer ?

adb014:
I've implemented the second idea, but if you prefer the first solution I can do that as well

ColdHeat:
I think this is a tough situation. I think you were right all along to include a solution_state field in the API response so I apologize for changing that. 

I think likely what should happen is that we should include `solution_state` in the API response but keep the ID behavior as is. If the solution_state is not `hidden` then we do the refresh from the API. What do you think? 

EDIT: Hmm still need to think about this one. I think an attribute should be added to the API response but it's still not clear what.

adb014:
I don’t think it will be possible to keep the exact behavior of solution_id. With what you seem to want there are 3 cases to treat

- solution_state == ‘hidden’
In this case solution_id can be null as the solution_id will never be needed to display the solution tab in the challenge modal. This behavior stays the same

- solution_state == ‘visible’
Here solution_id is always set as the solution tab of the challenge modal is always visible 

- solution_state == ‘solved’
Now this case is problematic. The logic of the visibility of the solution tab of the challenge modal can continue to be based on solution_id only if challenges.js:renderSubmissionResponse make another API like in this PR. In that case the presence or absence os solution_state in the API doesn’t come into play as it’s not used for the visibility of the solution tab

The visibility of the solution tab can’t depend on solution_id and solution_state alone, it will also need to depend on response.data.status being ‘correct’ or ‘already_solved’. In this case solution_id needs to be set


An alternative solution would be to always set solution_id and use principally solution_state to determine the solution tab visibility. 

Frankly, thinking about it I prefer the solution in this PR because the user is not informed of the existence or not of a solution via the API until they’ve solved the challenge. After solution_state could be added or not to the API

ColdHeat:
What I'm suggesting is that the `solution_id` behavior is kept the same, and `solution_state` returns the state and if the state is not hidden, we then do the API call. I believe this covers the cases that you mentioned but it does reveal the existence of the solution. 

Personally I don't like having to do the extra API call for every response but it might be unavoidable.

adb014:
We could limit the extra API to only response.data.status equaling ‘correct’. If it’s ‘already_solved the solution_id will already be non null. This come to the same as using solution_state to see if an api call is needed

adb014:
If we had solution_state in renderSubmissionResponse then the extra API would only be needed if solution_state was “solved” and response.data.status is correct. So it come down to more API calls or solution_state exposed. Ok I’ll propose something

ColdHeat:
I'm also looking to make this more resilient to changes in the future. If we want to add in more state behaviors, I don't really want to have to change the front end code too much. 

For example what if we wanted to add in a state where you see the solution if you went past max attempts.

ColdHeat:
Still thinking through this. I've so far decided that hiding that we should include solution_state in the challenge API response. It should be defaulted to hidden. 

Keeping the solution completely secret doesn't make much sense because in a situation where the solution would become public, it will eventually become public as well as the situations where it will become public so it makes sense to show solution_state if it helps us.

adb014:
You made the changes before me. Is there anything else you want changed in this PR ?

ColdHeat:
I should be able to handle it from here, thank you!

PR #2862, opened 2025-09-27
Fix brute force with max attempts

AnsonYeung:
When a user brute force flag submission on a challenge that limits max attempts, the attempt count can increase indefinitely as a new fail is added without checking for max attempts. This PR moves the brute force checking to be after checking max attempt so this cannot happen.

### DIFF
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -676,34 +676,6 @@ class ChallengeAttempt(Resource):
 
         chal_class = get_chal_class(challenge.type)
 
-        # Anti-bruteforce / submitting Flags too quickly
-        kpm = current_user.get_wrong_submissions_per_minute(user.account_id)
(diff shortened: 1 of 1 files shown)


ColdHeat:
I see what you're saying since the max attempts flow doesn't actually record the ratelimited attempts but the global block does. 

I do believe it was an intentional design choice to record the additional attempts. 

I think the better idea is to store these attempts under a ratelimited type instead of discarding them.

ColdHeat:
Also currently these ratelimited attempts are counted towards the user's overall fail count statistic. So should we not count this towards their failed attempts or should it be counted towards a ratelimited statistic?

ColdHeat:
I handled this in https://github.com/CTFd/CTFd/pull/2910. It's quite involved so just moving the code around wasn't enough but thank you for the PR.

Issue #2881, opened 2025-10-24
Improper HTML character handling in chart tooltip


tanpsi:
The chart for "Solve Percentages per Challenge" on the statistics page in admin panel is not handling HTML characters in challenge name correctly in the tooltip. In the image, the challenge "Hello, <script>!" only shows as "Hello, " in the tooltip.

<img width="1362" height="529" alt="Image" src="https://github.com/user-attachments/assets/2cc55e35-3838-4524-9899-b4638655ad87" />

ColdHeat:
Fixed with https://github.com/CTFd/CTFd/pull/2882

PR #2882, opened 2025-10-24
Encode HTML chars in tooltip formatter function of echarts

tanpsi:
It is recommended to use `echarts.format.encodeHTML` in tooltip formatters.[^1][^2] This fixes the name of challenge with HTML special characters in tooltip of the "Solve Percentages per Challenge" chart on statistics page. There might even be a XSS due to not encoding but probably not practical in this case because it needs to be in the challenge name.

[^1]: https://echarts.apache.org/en/option.html#tooltip.formatter
[^2]: https://echarts.apache.org/handbook/en/best-practices/security/#passing_raw_html_safely

### DIFF
--- a/CTFd/themes/admin/assets/js/pages/statistics.js
+++ b/CTFd/themes/admin/assets/js/pages/statistics.js
@@ -444,9 +444,9 @@ const graph_configs = {
         tooltip: {
           trigger: "item",
           formatter: function (data) {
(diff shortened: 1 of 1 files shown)


tanpsi:
Fixes #2881

ColdHeat:
Great report and fix, thank you very much!

PR #2885, opened 2025-10-25
Optimize Dockerfile layers and improve user handling in release image

eduardo010174:
This PR refactors the Dockerfile to improve layer efficiency, cache utilization, and user clarity in the final image, while maintaining functional equivalence with the previous version.

Key improvements:
Fixed redundant Docker layer caused by ownership changes
Identified with the tool dive, the command
chown -R ctfd:ctfd /var/log/CTFd /var/uploads /opt/CTFd
was creating an additional Docker layer containing duplicate /opt/CTFd data due to permission changes.

The file copy and ownership change have now been reorganized to avoid this unnecessary duplication, reducing final image size for 29MB and improving build caching.

Use of username instead of UID
Replaced direct use of USER 1001 and --chown=1001:1001 with the explicit user name ctfd.
This improves readability, maintainability, and consistency with standard Docker practices.

Restructured user and ownership creation
The useradd step now creates /opt/CTFd along with other directories before applying ownership.
Ensures all relevant directories are owned by ctfd:ctfd in a single step.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,22 +34,15 @@ RUN apt-get update \
         libffi8 \
         libssl3 \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/* \
+    && useradd --no-log-init --shell /bin/bash -u 1001 ctfd \
+    && mkdir -p /var/log/CTFd /var/uploads /opt/CTFd \
+    && chown -R ctfd:ctfd /var/log/CTFd /var/uploads /opt/CTFd
 
-COPY --chown=1001:1001 . /opt/CTFd
-
-RUN useradd \
-    --no-log-init \
-    --shell /bin/bash \
-    -u 1001 \
-    ctfd \
-    && mkdir -p /var/log/CTFd /var/uploads \
-    && chown -R 1001:1001 /var/log/CTFd /var/uploads /opt/CTFd \
-    && chmod +x /opt/CTFd/docker-entrypoint.sh
-
-COPY --chown=1001:1001 --from=build /opt/venv /opt/venv
+COPY --chown=ctfd:ctfd . /opt/CTFd
+COPY --chown=ctfd:ctfd --from=build /opt/venv /opt/venv
 ENV PATH="/opt/venv/bin:$PATH"
 
-USER 1001
+USER ctfd
 EXPOSE 8000
 ENTRYPOINT ["/opt/CTFd/docker-entrypoint.sh"]


ColdHeat:
This looks sane but I need to spend some time to test it out

Dockerfile
```diff
@@ -34,22 +34,17 @@ RUN apt-get update \
         libffi8 \
         libssl3 \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
-
-COPY --chown=1001:1001 . /opt/CTFd
+    && rm -rf /var/lib/apt/lists/* \
+    && useradd --no-log-init --shell /bin/bash -u 1001 ctfd \
+    && mkdir -p /var/log/CTFd /var/uploads /opt/CTFd \
+    && chown -R ctfd:ctfd /var/log/CTFd /var/uploads /opt/CTFd
```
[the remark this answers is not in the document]
eduardo010174: It's not true.
If apply this suggested change /opt/CTFd will get owner root. You can check this using [Dive ](https://github.com/wagoodman/dive) tool.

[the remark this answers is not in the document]
eduardo010174: It will set /opt/CTFd with root owner and all subfiles and directory's with correct permission. Same problem from old comment.

Dockerfile
```diff
@@ -34,22 +34,17 @@ RUN apt-get update \
...
-RUN useradd \
-    --no-log-init \
-    --shell /bin/bash \
-    -u 1001 \
-    ctfd \
-    && mkdir -p /var/log/CTFd /var/uploads \
-    && chown -R 1001:1001 /var/log/CTFd /var/uploads /opt/CTFd \
-    && chmod +x /opt/CTFd/docker-entrypoint.sh
+COPY --chown=ctfd:ctfd . /opt/CTFd
+RUN chmod +x /opt/CTFd/docker-entrypoint.sh
```
ColdHeat: Could this be part of the upper RUN command to save the extra layer?

eduardo010174: I will do ever better.
Change permission of file directly in git repository to execution permission.
Now the docker not need change permission in compile time.

For the future: git update-index --chmod=+x docker-entrypoint.sh

I think now it's ready for merge.

eduardo010174: Interesting, GitHub does not track changes in file permissions in web interface.

https://ionic.zendesk.com/hc/en-us/articles/360000160067-Adding-execute-permissions-using-git

However, if you check the commit in the terminal, the permission change is listed. Also, if you clone my fork, the file has the correct executable permission. This is exactly what happened with the prepare.sh script.

PR #2886, opened 2025-10-26
Add docker alpine image based

eduardo010174:
After pybluemonday is replaced for nh3, alpine is now supported.

Base image it's the same structure for the pull request #2885.


ColdHeat:
While reviewing this and checking other literature I found these updated blog posts which mirror my experience with Alpine, even in 2026:

https://pythonspeed.com/articles/base-image-python-docker-images/
https://pythonspeed.com/articles/alpine-docker-python/

I am aware that Python has improved the musl packaging situation but I also would still be very hesitant to use Alpine just b/c of how weird the bugs can be. 

I think that the changeset can left here but I dont want to support Alpine images as official yet.

PR #2887, opened 2025-10-26
Add some logic to catch infinite loop situation where verify_emails is on but we cant send emails

### LINKED ISSUE #2877: Infinite redirect loop when Verify Emails is enabled but mail settings are empty
### Summary
When Verify Emails is enabled while no email backend (SMTP) is configured, an infinite redirect loop occurs between `/challenges` and `/confirm`.
This might not be a strict bug, since enabling verification without configuring mail is a misconfiguration, but it could be worth handling or warning about.

### Steps to Reproduce

1. Deploy a fresh CTFd instance (tested on v3.8.0, commit a16c4c5).
2. Leave all email settings empty (no SMTP).
3. Enable Verify Emails under Admin → Settings → Accounts.
4. Register a new user (unverified).
5. Visit `/challenges`.

### Observed Behavior

Accessing `/challenges` redirects to `/confirm` (expected).

`/confirm` detects if not `can_send_mail():` and redirects back to `/challenges`.

Because the user is still unverified, `/challenges` again redirects to `/confirm`.
→ Infinite redirect loop.

### Expected / Discussion

It’s understandable that email verification requires a working mail backend.

However, the current loop may confuse admins testing configuration without SMTP.

It might be better to:

- show a static /confirm page explaining that mail is not configured, or

- display a warning when enabling “Verify Emails” without SMTP settings.

Relevant Code
https://github.com/CTFd/CTFd/blob/a16c4c5189856dee425c2cf5c78cfe4d1703cdfd/CTFd/auth.py#L37-L43

This unconditional redirect creates the loop when Verify Emails is ON but no mail backend is available.

### Environment

CTFd version: 3.8.0

### Conclusion
Not necessarily a bug — more of an edge case or usability concern — but it might deserve a small guard or admin warning to prevent confusion.

ColdHeat:
* Add some logic to catch infinite loop situation where verify_emails is on but we cant send emails
* Closes #2877

### DIFF
--- a/CTFd/auth.py
+++ b/CTFd/auth.py
@@ -38,9 +38,17 @@ auth = Blueprint("auth", __name__)
 @auth.route("/confirm/<data>", methods=["POST", "GET"])
 @ratelimit(method="POST", limit=10, interval=60)
 def confirm(data=None):
+    # If we can't send mails our behavior depends on verify_emails
     if not can_send_mail():
-        # If the CTF doesn't care about confirming email addresses then redierct to challenges
-        return redirect(url_for("challenges.listing"))
+        if get_config("verify_emails") is False:
+            return redirect(url_for("challenges.listing"))
+        else:
+            return render_template(
+                "confirm.html",
+                errors=[
+                    "Email verification is enabled but email sending isn't available. Please contact an admin to confirm your account"
+                ],
+            )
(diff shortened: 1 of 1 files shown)


ColdHeat:
image

PR #2895, opened 2025-11-06
Add action to build debian and alpine in parallel

eduardo010174:
Add strategy matrix to build in parallel debian and alpine image.

Dependent for pull requests #2886 and #2894 .


ColdHeat:
See https://github.com/CTFd/CTFd/pull/2886#issuecomment-3956126378 for rationale regarding Alpine base images

Basically Alpine is risky (but has gotten better) and I don't want to support it officially for the time being.

PR #2914, opened 2025-12-02
Fix: Do not check CSRF for static theme content

florianhofhammer:
Hi,

We have been running CTFd (thanks for the great project!) behind a reverse proxy for our CTF. To reduce load on the backend, we enabled caching in the reverse proxy to serve static cacheable content directly from the reverse proxy.

Unfortunately, this caused hard to debug issues where sometimes users would get logged in as different users (in a non-reproducible manner). We've narrowed it down to CTFd setting the `Set-Cookie` header even for static content. This caused the same header being served to users as long as it's in the cache and therefore sessions getting mixed up.

Flask only sets the header if the session has been modified. The only place a session for the static themes is modified is in the CSRF `before_request` hook. By returning early from the hook if the request is for static theme content, the session object is not modified and Flask doesn't add the header, therefore making static theme content cacheable without issues.

Note that the probably even cleaner fix would be to return from the CSRF check for all `GET`/`HEAD`/... requests but I didn't know whether that could cause issues further down the line.

Cheers,
Florian

### DIFF
--- a/CTFd/utils/initialization/__init__.py
+++ b/CTFd/utils/initialization/__init__.py
@@ -354,6 +354,7 @@ def init_request_processors(app):
 
     @app.before_request
     def csrf():
+        # TODO: CTFd 4.0 Consider reorganizing this function to only run on non safe methods
         # Early exit: no CSRF for functions explicitly marked as bypassing CSRF
         try:
             func = app.view_functions[request.endpoint]


florianhofhammer:
Hi,

I see that the theme verification run fails but the CI failure seems not to have anything to do with my changes. This also seems to be an issue in some of the other open PRs. Is this something that occurs sporadically or is this a blocker that would need to be resolved in my branch? Thanks for your feedback in advance!

ColdHeat:
@florianhofhammer Theme Verification is flaky for some reason that I haven't been able to figure out. I will take over on this PR and merge it in eventually.

florianhofhammer:
Amazing, thanks! :)

florianhofhammer:
Hi @ColdHeat, any news on this? We're planning to use CTFd again for another CTF and I'd prefer being able to use the upstream version over a fork with this fix. Thanks in advance!

ColdHeat:
My only concern with this code is the change to the csrf function as it technically affects security sensitive code. While I don't expect there to be an issue it still needs to be treated a bit carefully.

florianhofhammer:
Fair! But the CSRF shouldn't matter because the `views.themes` and `views.themes_beta` routes are GET-only anyways, no?

ColdHeat:
How about updating the code to only skip the csrf check if the endpoint
matches and the request method is GET?

florianhofhammer:
The method is already checked later in the function anyways: https://github.com/CTFd/CTFd/blob/b8299d7b7cd908ffc4a43a606c8095a956c08231/CTFd/utils/initialization/__init__.py#L361

I'll refactor the function so that this becomes a bit cleaner!

florianhofhammer:
I refactored the CSRF check function to not filter out based on the views anymore but generally on the safe methods (see [RFC 7231](https://www.rfc-editor.org/rfc/rfc7231#section-4.2.1). I also documented the cases where CSRF checks are bypassed a bit better and added additional error handling. Everything seems to work fine locally.

Let me know if that's useful! :)

ColdHeat:
I think the previous approach was easier to merge. Your newest change is certainly cleaner but has other side effects with unknown consequences. Most if not all `/theme` requests should end up being GET requests anyway but on the off chance something isn't I would rather continue to use existing behavior. 

I also haven't seen any other reports of this and I feel like you may be able to strip the Set-Cookie header in your cache.

florianhofhammer:
Thanks for the feedback and suggestions!

We've tried stripping the header in the cache as a first step of course, but for a reason I don't recall this did not work properly (could be a misconfiguration on our side but our config seemed sane). We disabled caching in the end to solve this issue temporarily but CTFd getting hammered for static assets that could easily be cached in the reverse proxy during the initial player inrush at CTF start seems wasteful :)

ColdHeat:
@florianhofhammer If this PR works for you I can merge it

florianhofhammer:
Thanks for the update!

Just a quick comment, maybe that's something for version 4.0: I think the check in https://github.com/CTFd/CTFd/blob/bb6c1c5a2fbeaf2c826dbab6c0852d8a1d8d6549/CTFd/utils/initialization/__init__.py#L388 is a bit risky: if the content type isn't JSON, it just tries to get the token from form content. If I understand how this works (and how Flask works internally, I suppose you're way more knowledgeable on that than me :)), that could lead to an exception and consequently an 500 error code being returned by flask if the request wasn't actually a form (which could always be the case, as a client can basically send whatever they want). Maybe actually just getting the nonce from form contents if the content type is a form and returning a 415 if the request is neither JSON nor a form would be a bit more descriptive and push the error handling to the client instead of causing issues on the server.
I don't know exactly how Flask behaves in this case, so maybe that's a non-issue from the start!

ColdHeat:
It could be more strict to have the additional specificity to define the CSRF handling for each content type but I don't think there's much of a risk to just generate a 500. If you want you can create an additional issue and I can put it for the 4.0 milestone.

PR #2917, opened 2025-12-08
Update Turkey to Turkiye

Bagcimurat:
Hi, 

For my country, Turkiye is used in international communication, not Turkey.
My next project is Turkish language support.

Thanks


ColdHeat:
@Bagcimurat I've added Turkish to Crowdin: https://crowdin.com/project/CTFd. Please send any translations there. 

As for this PR, I believe this list of timezones was pulled from some standards source. I'm not a huge fan of this JS blob so it's probably easier to simply remove this entire area and replace it with something else.

PR #2554, opened 2024-06-05
Handle IntegrityError arrising from duplicate adds to Solves table, add many BaseChallenge tests

cgoecknerwald:
The primary goal of this PR is to resolve a long-standing IntegrityError arising from failure to check for existing Solves rows before attempting to add another.

1. [Check for existing_solve](https://github.com/CTFd/CTFd/pull/2554/files#diff-bf014fce9dead43b10bb25ec04702d8245f4e4adcdd0f2973ad1ce73ec910fc6R150)
2. Gracefully handle malformed requests in [BaseChallenge.attempt()](https://github.com/CTFd/CTFd/pull/2554/files#diff-bf014fce9dead43b10bb25ec04702d8245f4e4adcdd0f2973ad1ce73ec910fc6R122), [BaseChallenge.solve()](https://github.com/CTFd/CTFd/pull/2554/files#diff-bf014fce9dead43b10bb25ec04702d8245f4e4adcdd0f2973ad1ce73ec910fc6R146), and [BaseChallenge.fail()](https://github.com/CTFd/CTFd/pull/2554/files#diff-bf014fce9dead43b10bb25ec04702d8245f4e4adcdd0f2973ad1ce73ec910fc6R186)
3. Add tests for the `/api/v1/challenges/attempt` endpoint for: [missing challenge_id](https://github.com/CTFd/CTFd/pull/2554/files#diff-7958118edbb33742e0ebf1c1b6752e0fd51276180a175815672307d5bf6b4adbR39), [missing submission](https://github.com/CTFd/CTFd/pull/2554/files#diff-7958118edbb33742e0ebf1c1b6752e0fd51276180a175815672307d5bf6b4adbR46), and [duplicate solves](https://github.com/CTFd/CTFd/pull/2554/files#diff-7958118edbb33742e0ebf1c1b6752e0fd51276180a175815672307d5bf6b4adbR54)
4. Add tests for the `BaseChallenge` class for `attempt()`, `solve()`, and `fail()` functions. These tests each cover: good user mode calls, good team mode calls, duplicate calls, incorrect flag calls, empty flag calls, submission=None calls, and empty form calls.

PR #2918, opened 2025-12-08
fix duplicate solves

vellvoid:
Marking submissions as correct sends a PATCH request to a submission endpoint which did not check whether a solve already exists, resulting in a duplicate and a 500 error. Similarly `BaseChallenge.solve()` did not have any validation, although in this case the controller would not allow this to be called correctly returning `already_solved`. 

Added a check for duplicate solves, the submissions API will respond with a 400 error, and `BaseChallenge` will just silently return.

Supersedes #2554

### DIFF
--- a/CTFd/plugins/challenges/__init__.py
+++ b/CTFd/plugins/challenges/__init__.py
@@ -146,8 +146,8 @@ class BaseChallenge(object):
 
         for attr in ("initial", "minimum", "decay"):
             if (
-                    challenge.function in DECAY_FUNCTIONS
-                    and getattr(challenge, attr) is None
+                challenge.function in DECAY_FUNCTIONS
+                and getattr(challenge, attr) is None
             ):
                 db.session.rollback()
                 raise ChallengeUpdateException(


CTFd/api/v1/submissions.py
```diff
@@ -190,6 +190,20 @@ def patch(self, submission_id):
...
         if submission_type == "correct":
+            existing_solve = Solves.query.filter_by(
+                challenge_id=submission.challenge_id,
+                user_id=submission.user_id,
+                team_id=submission.team_id if submission.team_id else None,
+            ).first()
+
+            # If a solve for this user / team pair exists, don't create a new solve
+            if existing_solve:
+                db.session.rollback()
```
ColdHeat, line 201: No need for rollback because nothing was created right?

CTFd/plugins/challenges/__init__.py
```diff
@@ -249,6 +249,17 @@ def solve(cls, user, team, challenge, request):
...
+        existing_solve = Solves.query.filter_by(
+            challenge_id=challenge.id,
+            user_id=user.id,
+            team_id=team.id if team else None,
+        ).first()
+
+        # If a solve for this user / team pair exists, don't create a new solve
+        if existing_solve:
+            return
+
```
ColdHeat, line 262: Can you put this and the related test into a separate PR? I feel like we should allow the BaseChallenge method to throw an exception.

vellvoid:
Superseded by #2919 #2920

PR #2920, opened 2025-12-09
fix duplicate solves in challenge

vellvoid:
splitting as discussed in #2918

### DIFF
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -13,6 +13,7 @@ from CTFd.cache import cache, clear_challenges, clear_ratings, clear_standings
 from CTFd.constants import RawEnum
 from CTFd.exceptions.challenges import (
     ChallengeCreateException,
+    ChallengeSolveException,
     ChallengeUpdateException,
 )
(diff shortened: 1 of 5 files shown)


vellvoid:
@ColdHeat You mentioned that BaseChallenge should rather throw an exception and I'm fine with either approach. 

~Overall unlike in #2919, these changes are for completeness' sake only, it's not something that really happens.~

I think this does actually happen, so we should check the challenge attempt validation as well.

ColdHeat:
I think silent return is the wrong choice. My current take is that I think these methods should actually be allowed to error and let the caller catch whatever exception it is. Perhaps we could create wrapper exceptions that denote the conflict. Open to suggestions.

CTFd/api/v1/challenges.py
```diff
@@ -811,9 +812,35 @@ def post(self):
...
+                        )
+
+                        return {
+                            "success": True,
+                            "data": {
+                                "status": "already_solved",
+                                "message": f"{message} but you already solved this",
+                            },
+                        }
+
```
ColdHeat, line 843: 👍

Issue #2916, opened 2025-12-06
Migration error on first run for MariaDB 12.1 and above


GramThanos:
Starting with MariaDB 12.1, the FOREIGN KEY constraint names should be unique only within the tables ([ref](https://mariadb.org/per-table-unique-foreign-key-constraint-names-new-feature-in-mariadb-12-1/)). This breaks the CTFd compatibility with MariaDB 12.1 and above due to the use of `_ibfk_1` suffix during migration when looking to drop constrains.

Affected files:
- [b37fb68807ea_add_cascading_delete_to_dynamic_.py](https://github.com/CTFd/CTFd/blob/945aa5472e3e5235a2ef608472b258f5fee750fe/CTFd/plugins/dynamic_challenges/migrations/b37fb68807ea_add_cascading_delete_to_dynamic_.py)
- [b295b033364d_add_ondelete_cascade_to_foreign_keys.py](https://github.com/CTFd/CTFd/blob/945aa5472e3e5235a2ef608472b258f5fee750fe/migrations/versions/b295b033364d_add_ondelete_cascade_to_foreign_keys.py)

Related error:
```
ctfd-1     | INFO  [alembic.runtime.migration] Running upgrade b5551cd26764 -> b295b033364d, Add ondelete cascade to foreign keys
ctfd-1     | Traceback (most recent call last):
ctfd-1     |   File "/opt/venv/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1905, in _execute_context
ctfd-1     |     self.dialect.do_execute(
ctfd-1     |   File "/opt/venv/lib/python3.11/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
ctfd-1     |     cursor.execute(statement, parameters)
ctfd-1     |   File "/opt/venv/lib/python3.11/site-packages/pymysql/cursors.py", line 148, in execute
ctfd-1     |     result = self._query(query)
ctfd-1     |              ^^^^^^^^^^^^^^^^^^
ctfd-1     |   File "/opt/venv/lib/python3.11/site-packages/pymysql/cursors.py", line 310, in _query
ctfd-1     |     conn.query(q)
ctfd-1     |   File "/opt/venv/lib/python3.11/site-packages/pymysql/connections.py", line 548, in query
ctfd-1     |     self._affected_rows = self._read_query_result(unbuffered=unbuffered)
[...]

PR #2922, opened 2025-12-14
dynamic fallback to new constraint name used by MariaDB 12.1 and above

GramThanos:
MariaDB 12.1 and above breaks some migration as the new constraint names are not matching the default ones.
More info: https://mariadb.org/per-table-unique-foreign-key-constraint-names-new-feature-in-mariadb-12-1/

This pr fixes the #2916


migrations/versions/b295b033364d_add_ondelete_cascade_to_foreign_keys.py
```diff
@@ -13,21 +13,55 @@
...
+                    op.drop_constraint(
+                        name.split("_ibfk_", 1)[1], table_name, type_=type_
+                    )
+                    return
+            # Re-raise any other unexpected operational error
+            raise
+    else:
+        # Try to run the constraint drop operation as normal
+        op.drop_constraint(name, table_name, type_=type_)
+
```
ColdHeat, line 49: This function is shared across the migrations, is it possible to centralize it in a shared file?

ColdHeat:
I'm not against this but I worry what will happen as MariaDB drifts further away from MySQL...

This concept has been applied before but mostly for MySQL & Postgres:
https://github.com/CTFd/CTFd/blob/bb6c1c5a2fbeaf2c826dbab6c0852d8a1d8d6549/migrations/versions/b295b033364d_add_ondelete_cascade_to_foreign_keys.py

Perhaps a better solution is to have a helper function that determines the appropriate foreign key name depending on the database configuration?

GramThanos:
> I'm not against this but I worry what will happen as MariaDB drifts further away from MySQL...
> 
> This concept has been applied before but mostly for MySQL & Postgres:
> https://github.com/CTFd/CTFd/blob/bb6c1c5a2fbeaf2c826dbab6c0852d8a1d8d6549/migrations/versions/b295b033364d_add_ondelete_cascade_to_foreign_keys.py
> 
> Perhaps a better solution is to have a helper function that determines the appropriate foreign key name depending on the database configuration? 

I am not sure if you can do that as you will have to detect both type (MariaDB) and version. Ideally the code shouldn't reference dynamically generated names (such as the constraint names) and instead name them during creation.

Right now I opted for this fallback as the code already guesses the name, and this seemed to best way not to interfere too much with the database due to the fact that this could be a bit critical. I don't think that there will be a similar issue with MariaDB any time soon as, based on what I have seen, this it they only part that assumes something database specific and the rest utialise external libraries for that. On top of that, if MariaDB deviate again form MySQL, it will means that it will break backwards compatibility with its previous versions.

To sum up, if you think it should be implemented in another way, my bet will be to try and dynamically recover from the database the contrain names through the functionalities of a library and not assume the names.

GramThanos:
@ColdHeat we can also investigate using `sqlalchemy.engine.reflection.Inspector.get_foreign_keys`to recover the constraint name. This may be a good way to remove code that is database technology specific.

PR #2935, opened 2026-01-15
add weight to challenges

vellvoid:
This adds weight to challenges, for sorting purposes. 
Ascending by default - e.g. lower weight first. Falls back to challenge value and ID as before. 
It also works within custom sort functions (challenge.weight).


vellvoid:
Weight should be descending and unsigned. If ascending, when an user would like to push a challenge to be first - the weight would have to be negative, as the default is 0.

ColdHeat:
I really worry that  we are overloading the create and update forms with too many inputs. Also in the case of this PR I think the changes to the create/update forms should be smaller with no whitespace changes.

PR #2938, opened 2026-01-19
Add for_each challenge logic

QuintenVdW:
Added a logic function that works for each flag submission.

It functions like the all flag required logic function but has a different working upon the partial answer,
it uses an award to give point matching the challenge points.

the logic function works for individual scoring and grants point for teams as well.
duplicate answers for teams are allowed

### DIFF
--- a/CTFd/api/v1/challenges.py
+++ b/CTFd/api/v1/challenges.py
@@ -254,6 +254,7 @@ class ChallengeList(Resource):
             ),
         },
     )
+    
     def post(self):
         data = request.form or request.get_json()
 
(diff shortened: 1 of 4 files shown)


NicoFgrx:
Hello there :wave: 
How is the logic supposed to work with "dynamic" challenges where points change over time based on the number of solves?
It seems you're using `challenge.value`, which, if I'm not mistaken, is the current value of the challenge.

Aren't the awards updated once they're given?

QuintenVdW:
> Hello there 👋 How is the logic supposed to work with "dynamic" challenges where points change over time based on the number of solves? It seems you're using `challenge.value`, which, if I'm not mistaken, is the current value of the challenge.
> 
> Aren't the awards updated once they're given?

the challenge looks if the flag was submitted if it was not yet submitted it will grant points if it is not the last flag it will use the award sytem to grant points for each individual submission, if it is the last flag it will not use the award but use the solve challenge to award points.

PR #2957, opened 2026-02-12
fix duplicate solution 500 error

vellvoid:
Creating a new solution for the same challenge causes a 500 error


ColdHeat:
I wonder if perhaps PUT was a better call for this endpoint.

PR #2962, opened 2026-02-17
add multi factor authentication

vellvoid:
Add support for TOTP code based MFA.

Few assumptions:
- token authentication is exempt from MFA
- admin can reset mfa status for any user
- implemented over the api in user settings
- qrcodes  are generated on the backend and returned as base64
- currently no setting to enforce mfa for users


vellvoid:
I know modifying the flask global `g` will probably raise some questions. The idea behind this is that we must know whether an user has been fully authenticated with a token, just checking if an authorization header exists is not enough without checking the token value.

I'm open to ideas how to better handle this, this PR does it:

1. in before_request hook, to skip requiring the full mfa flow if the access is token based
2. in the token create route, also to not require the full mfa with the api

PR #2982, opened 2026-03-09
better display error messages

vellvoid:
- fix the threatening 403 forbidden when ctf has not started / has ended
- add handling for 401s 
- ensure api returns an explicit reason for 401 / 403


CTFd/themes/core-deprecated/templates/errors/403.html
ColdHeat: Shouldn't be touching the core-deprecated theme

mohamedsolaiman:
Hey! I recently went through the CTFd codebase while working on a fix for the user creation endpoint (#3034). If you need help implementing better error messages, I can point you to the right files — the validation happens in `api/v1/users.py` and the schema is in `schemas/users.py`. Happy to help if you want to pair on this! 🙂

mohamedsolaiman:
Hey! I recently went through the CTFd codebase while working on a fix for the user creation endpoint (#3034). If you need help implementing better error messages, I can point you to the right files. The validation happens in api/v1/users.py and the schema is in schemas/users.py. Happy to help if you want to pair on this!

ColdHeat:
In the current structure this is probably the only way to control this behavior. 

I've been thinking about whether we could use Pages to pull this off but the during_ctf_time_only decorator doesn't do so well with that. 

My issue with this is that this is a single case. It may be the most common case but I would like a more general solution. For example an admin can't add an image if they wanted to. 

Another thing that I don't like is that this is currently controllable with JavaScript and this config is not significantly better than that. Arguably it's worse because you can't control text.

PR #3012, opened 2026-04-07
Add user list filtering by status

erdnaxe:
Currently it is only possible to list non verified users, or hidden users by parsing CTF admin HTML.
This patch proposes to add `hidden`, `banned` and `verified` as GET arguments to the userlist API endpoint.

The commit also includes some tests to make sure we are not leaking sensible information to non-admin users.


ColdHeat:
Hmm I am surprised that it's not possible to filter hidden/banned via the API already. Needs some review but looks generally reasonable.

PR #3014, opened 2026-04-07
Replace request content_type check with is_json

hidehiroanto:
Currently, if a user sends a request with the Content-Type `application/json;charset=utf-8`, the `request.content_type == "application/json"` check fails. This PR aims to make the check more robust by simply replacing all such checks with `request.is_json`.


ColdHeat:
This looks reasonable. Thanks!

PR #3032, opened 2026-04-28
fix rare 500 on users create

vellvoid:
Creating a user via the API can cause a 500 error in a rare scenario where the email is a duplicate of the caller's email. Catch the error properly and return a 400 instead of 500.


ColdHeat:
This seems wrong since we might throw the wrong error message. Can this be put into UserSchema into validate_email?

PR #3035, opened 2026-05-01
Add length validation for affiliation and website fields

mohamedsolaiman:
Hey! Was setting up a CTF for my university and one of the participants managed to crash the user creation by submitting a super long affiliation name. Turns out the affiliation and website fields in UserSchema have no length validation even though theyre limited to 128 chars in the database.

If you submit an affiliation or website longer than 128 chars, you get a 500 from the database instead of a proper validation error. This PR adds Length validators to match the DB column size.

Changes:
- Added validate.Length(max=128) to affiliation field
- Added validate.Length(max=128) to website field
- Added test for affiliation length validation

### DIFF
--- a/CTFd/schemas/users.py
+++ b/CTFd/schemas/users.py
@@ -47,11 +47,23 @@ class UserSchema(ma.ModelSchema):
                 schemes={"http", "https"},
             )(website)
             if website
-            else True
+            else True,
+            validate.Length(
+                max=128, error="Website must be less than 128 characters"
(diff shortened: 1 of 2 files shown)


ColdHeat:
I'm not sure this is needed. 

image

```
❯ pytest -q -s -k test_affiliation_length_validation
Results (2.04s):
       1 passed
     654 deselected
```

If you want you can slim your PR down to just the test and I can accept that.

PR #3057, opened 2026-06-01
fix 500 on removing solutions with files

vellvoid:
`files.solution_id` FK has no cascade, so deleting a `Solutions` row that still has `SolutionFiles` violates the constraint leading to a 500 and making solutions impossible to delete before removing the referenced files.


ColdHeat:
Isn't the better solution here to add cascade delete to SolutionFiles and PageFiles?

ColdHeat:
I investigated this and SolutionFiles currently mimic PageFiles in that a SolutionFile can be shared between solutions.

I think this is actually the wrong behavior and solution files should instead be scoped to the solution itself. 

So the decision here to delete the solution files is the right one however I do think that the more complete solution is to cascade delete as well.

PR #3072, opened 2026-06-29
Add modules and audiences

ColdHeat:
* Adds Modules & Audiences


CTFd/utils/modules/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+
+def get_accessible_module_ids(user):
+    """Module IDs accessible to this account, cached per-request on flask.g."""
+    if user is None:
+        return set()
+
+    cache_key = "_ctfd_accessible_module_ids"
+    cached = getattr(g, cache_key, None)
+    if cached is not None and cached.get("user_id") == user.id:
+        return cached["ids"]
```
ColdHeat: Strongly suspect that this caching isn't sufficient as this is per request.

PR #3082, opened 2026-07-17
Improve DX and upgrade tooling

vellvoid:
Replaces pip, pipdeptree, black, isort, bandit, requirements.in + pip-compile-in-docker script with just ruff and uv.

Most of the linting dependencies were already outdated, all of them would be now replaced with ruff. Reformatting is manageable, and would have occurred anyway even if we just updated black to a more recent version.

Updated prettier and made the version match between the themes, as well as switched the makefile to use project-local prettier via yarn (so that it won't pick-up system-level prettier with perhaps a different version). This was the cause for a lot of lint failures in the past - prettier versions fighting. Pinned node version, and made slight improvements to the verify-themes job in hopes that it will become more reproducible and reliable.

Also added non-default themes and plugins to gitignore - this one is debatable, but i feel useful for working on ctfd with various plugins that are managed separately.

Finally dropped pretty outdated Vagrantfile and prepare script. I believe this is not relevant with Dockerfile and docker-compose present.


.prettierignore
ColdHeat, line 1: I believe this is superceded by gitignore

ColdHeat:
Seems like translations need to be regenerated under uv now for some reason. 

I checked one of the generated files and it seems the difference is simply version

```
❯ diff diff-old.txt diff-new.txt
490c490
< 00001e90: 2042 6162 656c 2032 2e31 382e 300a 0025   Babel 2.18.0..%
---
> 00001e90: 2042 6162 656c 2032 2e31 372e 300a 0025   Babel 2.17.0..%
```

PR #3084, opened 2026-07-21
Avoid 500 when concurrent requests race on the same Tracking row

dustt112313:
## Problem

Visit tracking commits a `Tracking` row on every authenticated `POST`/`PATCH`/`PUT`/`DELETE` request (`init_request_processors` in `CTFd/utils/initialization/__init__.py`). When a single session fires **concurrent** mutating requests, both handlers read and update the **same** `(user_id, ip)` `Tracking` row, and the second commit fails with an `OperationalError`:

```
sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError)
(1020, "Record has changed since last read in table 'tracking'; try restarting transaction")
```

The surrounding `try/except` only catches `InvalidRequestError` and `IntegrityError`, so the `OperationalError` propagates and the whole request returns **500** — and the intended action is rolled back.

## Reproduction

Log in as admin and delete several challenges at once (the admin UI fires the `DELETE /api/v1/challenges/<id>` calls in parallel). One request completes; the others return 500 and their challenges are **not** deleted. The same shape happens for any bulk admin action, or any two concurrent mutating requests from one session, on MySQL/MariaDB.

Note this only affects concurrent requests **within a single session** (same `user_id` + `ip`). Separate users racing each other touch different rows and are unaffected.

## Fix

Catch `OperationalError` around the tracking commit and roll it back. Visit tracking is best-effort, so a dropped tracking write should not fail the underlying request. No behavior change for the normal (sequential) path.

## Notes

A deterministic unit test is awkward because the failure depends on a real concurrent commit against MySQL/MariaDB, so this PR is the minimal defensive change. Happy to add a test if you can point me at a preferred pattern for exercising concurrent request handling.

### DIFF
--- a/CTFd/utils/initialization/__init__.py
+++ b/CTFd/utils/initialization/__init__.py
@@ -4,7 +4,7 @@ import os
 import sys
 
 from flask import abort, redirect, render_template, request, session, url_for
-from sqlalchemy.exc import IntegrityError, InvalidRequestError
+from sqlalchemy.exc import IntegrityError, InvalidRequestError, OperationalError
 from werkzeug.middleware.dispatcher import DispatcherMiddleware
 
 from CTFd.cache import clear_user_recent_ips
@@ -278,6 +278,12 @@ def init_request_processors(app):
                     db.session.rollback()
                     db.session.close()
                     logout_user()
+                except OperationalError:
+                    # Concurrent requests within the same session can race to
+                    # update the same Tracking row (e.g. MySQL error 1020 or a
+                    # deadlock). Visit tracking is best-effort, so roll back and
+                    # continue serving the request instead of returning a 500.
+                    db.session.rollback()
                 else:
                     clear_user_recent_ips(user_id=session["id"])
 


ColdHeat:
Hello how did this affect your instance and what happens if the rollback is applied? I imagine one request succeeds and the other will just silently fail?

dustt112313:
Hi @ColdHeat, thanks for taking a look!

**How it affected our instance:** we run CTFd (MariaDB + gunicorn with multiple workers) for a live competition. During the event, admin bulk operations kept failing: deleting several challenges from the admin UI (which fires the `DELETE /api/v1/challenges/<id>` calls in parallel) deleted one challenge and returned 500 for the rest, so the admin had to retry them one by one. The same 500s showed up for any burst of concurrent authenticated mutating requests from a single session. The traceback for every failure pointed at the tracking commit in `init_request_processors`, not at the endpoint logic.

**What happens with the rollback applied:** both requests succeed — nothing user-visible fails. The key detail is that this commit runs in `before_request`, *before* the view executes. So today the losing request dies with a 500 and its real action never runs at all. With the patch, the loser just drops its tracking write (rollback + close, same as the existing handler), then proceeds into the view normally.

The only thing that is "silently" lost is the loser's update of `Tracking.date` — and the concurrent winner has already committed the same timestamp (same second) to the same row, so the stored visit data ends up identical either way. No tracking rows go missing: this race only happens on the *update* path, where the `(user_id, ip)` row already exists; the insert race is already covered by the existing `IntegrityError` handling.

So it's the same best-effort trade-off the surrounding `except (InvalidRequestError, IntegrityError)` already makes for the insert race — this PR just extends it to the update race, which surfaces as `OperationalError` on MySQL/MariaDB.

ColdHeat:
This seems like LLM generated commentary. 

Can you provide reproduction steps using the provided docker compose used in the repository? 

Or I should say, can you reproduce this issue using the provided docker compose?

dustt112313:
Not hypothetical - hit this live on my own instance. Repro: stock docker compose (WORKERS=4), log in as admin, fire ~20 concurrent authenticated DELETE/POST requests with the same session cookie (any bulk admin action does it). Some requests 500 with the traceback pointing at the `db.session.commit()` in `tracker()`, not the endpoint. The patch just extends the existing best-effort except by one error class.

Issue #1091, opened 2019-08-23
Feature Request : Customisable rate limiting


sachinkamath:
### Background

I had to host a CTF in a college which had all the PC behind a NAT and around 300-350 were taking part in the CTF. Due to this, all the participants were getting rate-limited.

I tried to adjust the kpm present in [/api/v1/challenges#L380](https://github.com/CTFd/CTFd/blob/97f52756bc579d5aa8bdf5376cfbeb0700d5f91d/CTFd/api/v1/challenges.py#L380) and restarted the platform but that didn't seem to do the trick.

### Solution

It'd be nice to have a setting in the admin to customize the lockout or disable it altogether (maybe by setting it to zero). 

@ColdHeat if you can point me to the right lines where the lockouts for challenge attempts are implemented, I can work on this.

ColdHeat:
This is implemented in the ratelimit decorator. https://github.com/CTFd/CTFd/blob/master/CTFd/utils/decorators/__init__.py#L130

pandatix:
Up on this issue. Same problem here... 😅 
When dealing with configuration I have a ton of jobs that sets up in parallel, coming from the same host... That hits the `@ratelimit` decorator thus break mid-job.

One solution would have been to use an API key but in this very specific context it is not possible (it would in turn create a ton of API keys as they try to be idempotent scripts, from a fresh or already-configured CTFd instance, hence making any key management strategy impossible or lacky). Every alternatives I though about were just workarounds to a proper solution.

Do you want me to set up a PR to solve this @ColdHeat ?
Would you prefer it live-configurable (in the Config/Settings forms) or varenv for immutability ? Both are fine to me 🤷

ColdHeat:
API rate limits should probably be defined in config and KPM can be defined in config with an override in the Admin Panel. 

A PR is okay.

PR #3090, opened 2026-07-31
Make ratelimits configurable

felix-m-c:
### Summary:
currently, it is not possible to configure or disable the ratelimit system.
This poses an issue for our events where many (30+) users may share a single IP address.
In these cases features like flag submission immediately run into rate limits for all users.
This issue was previously discussed in #1091, but did not lead to a PR at the time.

### Change:
This PR moves the configuration of rate limits from the sourcecode to the config.ini file. (excerpt see below)
It also adds a new RL_DISABLE_COMPLETELY option that disables ratelimits outright.

### Backwards Compatibility:
Plugins using the syntax `@ratelimit(method="POST", limit=10, interval=60)` still work because I only add (optional!) keyword args. If these Values are provided, the new logic uses them instead of falling back to `RL_DEFAULT`.
The decorator `@ratelimit(method="POST")` now uses the configured default ratelimit.

Note that the `RL_DISABLE_COMPLETELY` config option overrides the ratelimit decorator for exiing plugins too.

```ini
...
[ratelimits]
# RL_DISABLE_COMPLETELY
# Controls whether the ratelimit functionality is used at all.
# setting this to true overrides all RL settings below
RL_DISABLE_COMPLETELY = false

# All of the following RL configs follow the same format
# format: key = limit,interval
# users may send <limit> calls in <interval> seconds (both int!)

# Fallback ratelimit
#   - if the given key was not found in this config
#   - and if the given key has no built-in default in config.py (backwards compat)
RL_DEFAULT = 10,60
# POST to /confirm and /confirm/data
RL_CONFIRM = 10,60
# POST to /reset_password and /reset_password/<data>
RL_RESET_PASSWORD = 10,60
# POST to /register
RL_REGISTER = 10,5
# POST to /login
RL_LOGIN = 10,5
# GET /redirect
RL_OAUTH_REDIRECT = 10,60
# POST to /teams/join
RL_JOIN_TEAM = 10,5
# POST to the exports/raw endpoint
RL_EXPORT = 10,60
# POST to users/<userid>/email (sending an email to a user)
RL_EMAIL_USER = 10,60
# GET /events
RL_EVENTS = 150,60
...
```

### DIFF
--- a/CTFd/api/v1/exports.py
+++ b/CTFd/api/v1/exports.py
@@ -14,7 +14,7 @@ exports_namespace = Namespace("exports", description="Endpoint to retrieve Expor
 @exports_namespace.route("/raw")
 class ExportList(Resource):
     @admins_only
-    @ratelimit(method="POST", limit=10, interval=60)
+    @ratelimit(method="POST", rl_key="RL_EXPORT")
     def post(self):
         req = request.get_json()
         export_type = req.get("type", "_")
--- a/CTFd/api/v1/users.py
+++ b/CTFd/api/v1/users.py
@@ -514,7 +514,7 @@ class UserEmails(Resource):
         description="Endpoint to email a User object",
         responses={200: ("Success", "APISimpleSuccessResponse")},
     )
-    @ratelimit(method="POST", limit=10, interval=60)
+    @ratelimit(method="POST", rl_key="RL_EMAIL_USER")
     def post(self, user_id):
         req = request.get_json()
         text = req.get("text", "").strip()
--- a/CTFd/auth.py
+++ b/CTFd/auth.py
@@ -34,7 +34,7 @@ auth = Blueprint("auth", __name__)
 
 @auth.route("/confirm", methods=["POST", "GET"])
 @auth.route("/confirm/<data>", methods=["POST", "GET"])
-@ratelimit(method="POST", limit=10, interval=60)
+@ratelimit(method="POST", rl_key="RL_CONFIRM")
 def confirm(data=None):
     # If we can't send mails our behavior depends on verify_emails
     if not can_send_mail():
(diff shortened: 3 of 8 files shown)


jusito:
@felix-m-c and I discussed about this.

- For us ratelimit on / off is sufficient.
- Individual options seemed to be a good follow up to be accepted, but we are unsure about the use-cases.
- ip & ip range whitelisting was also discussed but it requires more code. AI produced something which seems good but we would need to review it in depth.
- Is config only fine, do we need an admin route, and / or admin ui option?

ColdHeat:
Hello, I will need to think about whether these should be configured from code config or UI config but I'm more curious about the underlying issue. I believe that the ratelimit applies to IP address and code so what was the issue caused by the shared IP? Were all of the users also sharing the same account? 

> This poses an issue for our events where many (30+) users may share a single IP address.

jusito:
Hello @ColdHeat , thanks for the quick reply, all our cases use different accounts.

1. The mentioned event was a shared workstation, different accounts.
2. We are doing E2E testing via playwright, each client registers its own account.

felix-m-c:
The Ratelimiter uses the Cache System to track the number of accesses over time.
The Cache Key is defined as `key = "{}:{}:{}".format(key_prefix, ip_address, request.endpoint)` (key_prefix is always "rl").
The access-counter is therefore identified by (ip, endpoint).

Not tying access counters to usernames/emails makes sense because many rate-limited endpoints would be called before login. (/login, /register, /reset_password, /redirect)

For some of these, we could add another identifier to the cache key, but that would be some user-provided (untrusted) string that someone trying to overwhelm the system could just randomize anyways. (e.g., the requested email adress for /reset_password).


(i would also like to mention here that the rate-limiting logic does not limit users to N accesses over M seconds, but instead it counts until N and only resets if a gap of M seconds occurred in between. E.g., 1 access every 50 seconds would trigger a (limit=10, interval=60) configuration after 500 Seconds)

ColdHeat:
I agree that this should be implemented via config.ini/config.py. 

The names of the variables should be `RATELIMIT` and not `RL`. Instead of `RL_DISABLE_COMPLETELY` it should be `RATELIMIT_DISABLE`. 

Also instead of hardcoding the values in config.ini I think it would be better to set them in config.py so that it's easier for an envvar to override.

